PR #87, opened 2023-11-03
Add lcov Script Support in CMake Files

tom-daubney-arm:
This PR makes two small modifications to CMake files to enable the TF-PSA-Crypto repository to make use of the newly modified `lcov.sh` script that was introduced in [Mbed TLS PR 8395](https://github.com/Mbed-TLS/mbedtls/pull/8395).

At the time of raising this PR, the new `lcvo.sh` script has not been merged into TF-PSA-Crypto from Mbed TLS yet, and so this might need rebasing after that happens.


ronald-cron-arm (changes requested):
I think we are missing the copy of lcov.sh by psa_crypto.py, function `copy_from_scripts()`. Otherwise, this looks good to me.

tom-daubney-arm:
> I think we are missing the copy of lcov.sh by psa_crypto.py, function `copy_from_scripts()`. Otherwise, this looks good to me.

Thanks, I think I have done this correctly now. PTAL.

ronald-cron-arm (approved):
LGTM, @gabor-mezei-arm please have a look.

gabor-mezei-arm (approved):
LGTM

PR #89, opened 2023-11-08
License change

ronald-cron-arm:
For reviewers: Inspired by #8462 in mbedtls repo.

### DIFF
--- a/README.md
+++ b/README.md
@@ -237,8 +237,4 @@ as parameters that will be passed to the test function.
 License
 -------
 
-Unless specifically indicated otherwise in a file, TF-PSA-Crypto files are
-provided under the [Apache-2.0](https://spdx.org/licenses/Apache-2.0.html)
-license. See the [LICENSE](LICENSE) file for the full text of this license.
-Contributors must accept that their contributions are made under the
-Apache-2.0 license.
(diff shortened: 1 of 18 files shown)


tom-daubney-arm (approved):
I don't think any files have been missed here. The ones that have not been changed get brought over from Mbed TLS so I think this is all good. LGTM.

README.md
```diff
@@ -237,8 +237,4 @@ as parameters that will be passed to the test function.
 License
 -------
 
-Unless specifically indicated otherwise in a file, TF-PSA-Crypto files are
-provided under the [Apache-2.0](https://spdx.org/licenses/Apache-2.0.html)
-license. See the [LICENSE](LICENSE) file for the full text of this license.
-Contributors must accept that their contributions are made under the
```
daverodgman, line 243: Shouldn't we retain this line `Contributors must accept...` (as we have done in Mbed TLS)?

daverodgman (changes requested):

README.md
tom-daubney-arm: I think you're right. I didn't notice that.

ronald-cron-arm: I removed it because we do not accept contributions currently. We could add it back when we open to contributions. That's the rationale behind that change.

daverodgman: OK - please make a note of this somewhere so we don't forget

daverodgman (approved):

PR #91, opened 2023-11-28
Add component for check_generated_files testing

### LINKED ISSUE #50: Add generated file checks
Add generated file checks
- [ ] Adapt tests/scripts/check-generated-files.sh to work for both mbedtls and TF-PSA-Crypto. Note: TF-PSA-Crypto has less generated files: no error.c, version_features.c, ssl_debug_helpers_generated.c, visualc/VS2013 ... Keep check of all files when called in the context of mbedtls for the time being.
- [ ]  Add component_check_generated_files all.sh component in `tf-psa-crypto/tests/scripts/components-basic-checks.sh`. 

Note: no make build system, CMake should be used instead

Depends on: https://github.com/Mbed-TLS/mbedtls/pull/9720.

DOD: Same support for generated file checking in TF-PSA-Crypto as in mbedtls.

tom-daubney-arm:
Fixes #50 (partially) along with [Mbed TLS PR 8523.](https://github.com/Mbed-TLS/mbedtls/pull/8523).

This PR adds an all.sh component that is designed to replicate the same level of testing that `component_check_generated_files` achieves in Mbed TLS.

Note to reviewers:- I don't think this PR is quite complete, but I would appreciate some feedback on what might need to happen to get it there. Thanks.

### DIFF
--- a/scripts/psa_crypto.py
+++ b/scripts/psa_crypto.py
@@ -153,6 +153,7 @@ def copy_from_tests(mbedtls_root_path, psa_crypto_root_path):
                            "generate_test_code.py|"\
                            "scripts_path.py|"\
                            "test_generate_test_code.py|"\
+                           "check_generated_files.sh|"\
                            "test_psa_compliance.py",
                            file_), os.listdir(scripts_source_path))
     for file_ in scripts_files:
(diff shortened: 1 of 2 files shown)


ronald-cron-arm:
Superseded by #201.

Issue #77, opened 2023-10-03
Add support for p256m driver


ronald-cron-arm:
The goal of this issue is to add the p256m driver in psa-crypto.

List of tasks to be done (may not be comprehensive):
- [ ] Add copy of directory p256m into drivers/p256m
- [ ] Add a drivers/CMakeLists.txt similar to the 3rdparty/CMakeLists.txt one (only p256m, not everest)
- [ ] Add a drivers/p256m/CMakeLists.txt (do we need to install the p256m includes as in mbedtls?)
- [ ] In core/CMakeLists.txt add two target_link_libraries( ... PRIVATE p256m) like in library/CMakeLists.txt (not `PUBLIC p256m` but `PRIVATE p256m` though I'd say).
- [ ] Add PSA_CRYPTO_P256M_DRIVER_ENABLED config option and its documentation in the section "PSA driver interface implementation configuration options" of crypto_config.h. Similar to MBEDTLS_PSA_P256M_DRIVER_ENABLED in mbedtls. Disabled by default.
- [ ] Add its translation to MBEDTLS_PSA_P256M_DRIVER_ENABLED in config_adjust_mbedtls_from_psa_crypto.h.
- [ ] Add a test_ccm_aes_sha256_secp256r1_p256m test component based on the configs/ccm-aes-sha256-secp256r1.h reference configuration with PSA_CRYPTO_P256M_DRIVER_ENABLED enabled similar to the test_tfm_config_p256m_driver_accel_ec test component in mbedtls.
- [ ] Check that the test component do what it intends to do and give evidence of this in this issue or the associated PR for the benefit of reviewers.

ronald-cron-arm:
This was done as part of the the repo split.

PR #121, opened 2024-12-17
Out of source build development

Harry-Ramsey:
## Description
Closes [#9755](https://github.com/Mbed-TLS/mbedtls/issues/9755)

## PR checklist

- [x] **changelog** not required because: testing API changes.
- [x] **development PR** provided.
- [x] **framework PR** provided [#77](https://github.com/Mbed-TLS/mbedtls-framework/pull/77)
- [x] **3.6 PR** not required because: for repo split.
- **tests**  not required.

## Notes for the submitter

Please refer to the [contributing guidelines](https://github.com/Mbed-TLS/mbedtls/blob/development/CONTRIBUTING.md), especially the

Help make review efficient:
* Multiple simple commits
* Avoid force-push

### DIFF
--- a/tests/scripts/components-build-system.sh
+++ b/tests/scripts/components-build-system.sh
@@ -11,16 +11,12 @@
 
 component_test_tf_psa_crypto_cmake_out_of_source () {
     msg "build: cmake tf-psa-crypto 'out-of-source' build"
-    TF_PSA_CRYPTO_ROOT_DIR="$PWD"
-    mkdir "$OUT_OF_SOURCE_DIR"
-    cd "$OUT_OF_SOURCE_DIR"
+    cd $OUT_OF_SOURCE_DIR
     # Note: Explicitly generate files as these are turned off in releases
     cmake -D CMAKE_BUILD_TYPE:String=Check -D GEN_FILES=ON "$TF_PSA_CRYPTO_ROOT_DIR"
     make
     msg "test: cmake tf-psa-crypto 'out-of-source' build"
     make test
-    cd "$TF_PSA_CRYPTO_ROOT_DIR"
-    rm -rf "$OUT_OF_SOURCE_DIR"
 }
 
 component_test_tf_psa_crypto_cmake_as_subdirectory () {
--- a/tests/scripts/all.sh
+++ b/tests/scripts/all.sh
@@ -17,13 +17,7 @@ if [ -d tf-psa-crypto -a -d include/mbedtls ]; then
     exit 255
 fi
 
-if [ -d ../include/mbedtls -a -d ../framework ]; then
-    # Running inside an mbedtls checkout: get the framework from mbedtls.
-    FRAMEWORK="$PWD/../framework"
-else
-    # Running standalone: use our own framework.
-    FRAMEWORK="$PWD/framework"
-fi
+FRAMEWORK="$PWD/framework"
 
 source $FRAMEWORK/scripts/all-core.sh
 


tests/scripts/components-build-system.sh
```diff
@@ -11,21 +11,16 @@
...
     msg "test: cmake tf-psa-crypto 'out-of-source' build"
     make test
-    cd "$TF_PSA_CRYPTO_ROOT_DIR"
-    rm -rf "$OUT_OF_SOURCE_DIR"
 }
 
 component_test_tf_psa_crypto_cmake_as_subdirectory () {
     msg "build: cmake 'as-subdirectory' build"
-    cd programs/test/cmake_subproject
+    cd  "$TF_PSA_CRYPTO_ROOT_DIR/programs/test/cmake_subproject"
```
ronald-cron-arm, line 23: The CI is not happy with $TF_PSA_CRYPTO_ROOT_DIR. I am not sure why we need it in the first place. In TF-PSA-Crypto only context we do not need it. In mbedtls CI context I think `tests/scripts/all.sh` goes `cd tf-psa-crypto; ./tests/scripts/all.sh` thus it should be needed either. But I may miss something.

Harry-Ramsey: The problem that I have ran into is how we identify tests that require out of source builds and tests that don't. The current solution at [line 410](https://github.com/Mbed-TLS/mbedtls-framework/blob/1797ef20b0c8aff96146b682b782fbc31fabb855/scripts/all-core.sh#L410) in the framework pull request just sets up an out of build directory regardless of the test.

ronald-cron-arm: Ah yes I see, sorry for the noise. We need that for this very peculiar component but will not need it for "standard" ones.

tests/scripts/components-build-system.sh
```diff
@@ -11,21 +11,17 @@
...
```
ronald-cron-arm, line 24:
```suggestion
    cd  programs/test/cmake_subproject
```

tests/scripts/components-build-system.sh
```diff
@@ -34,7 +30,7 @@ component_test_tf_psa_crypto_cmake_as_subdirectory () {
 
 component_test_tf_psa_crypto_cmake_as_package_install () {
     msg "build: cmake 'as-installed-package' build"
-    cd programs/test/cmake_package_install
+    cd $TF_PSA_CRYPTO_ROOT_DIR/programs/test/cmake_package_install
```
ronald-cron-arm, line 33:
```suggestion
    cd programs/test/cmake_package_install
```

ronald-cron-arm (changes requested):

ronald-cron-arm:
In the CI, the three TF-PSA-Crypto fail. I think this is because of:
```
if [ -d ../include/mbedtls -a -d ../framework ]; then                           
    # Running inside an mbedtls checkout: get the framework from mbedtls.       
    FRAMEWORK="$PWD/../framework"                                               
else                                                                            
    # Running standalone: use our own framework.                                
    FRAMEWORK="$PWD/framework"                                                  
fi 
```
in TF-PSA-Crypto:tests/scripts/all.sh
It means I think that while running the CI, the framework used by tf-psa-crypto/tests/scripts/all.sh is the mbedtls one that does not set  TF_PSA_CRYPTO_ROOT_DIR or create automatically the directory for out-of-tree build. @mpg could probably help to resolve that properly.

mpg:
> @mpg could probably help to resolve that properly.

I don't think this is complicated to resolve: this block of code was written based on the assumption that we wanted tf-psa-crypto to use the framework from Mbed TLS - but that's not the direction we're taking now. (Also, something had to be done because the framework subdir of tf-psa-crypto did not exist when that code was written.)

So, I'd suggest simply replacing the quoted `if else` block with `FRAMEWORK="$PWD/framework"` and I think we should be OK.

mpg (changes requested):
Please do the change requested here: https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/121#issuecomment-2559235090

Harry-Ramsey:
> Please do the change requested here: [#121 (comment)](https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/121#issuecomment-2559235090)

Sorry, I missed that entirely. Fixed now

mpg (commented):
LGTM. I'm just waiting for the CI results before I formally approve.

ronald-cron-arm:
The same on my side, LGTM, waiting for the CI.

ronald-cron-arm (approved):
LGTM, thanks.

ronald-cron-arm (approved):
LGTM

mpg (approved):
LGTM, thanks!

PR #123, opened 2024-12-17
Add GitHub pull request template

### LINKED ISSUE #98: Add CONTRIBUTING.md
A file called `CONTRIBUTING.md` is a de facto standard to indicate how to contribute to a project. This repository should have a file, even if the current content is just to tell people to contribute to the `mbedtls` repository for the time being.

ronald-cron-arm:
Add GitHub pull request template
Fixes #98

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -10,8 +10,8 @@ Please remove the segment/s on either side of the | symbol as appropriate, and a
 If the provided content is part of the present PR remove the # symbol.
 
 - [ ] **changelog** provided | not required because: 
-- [ ] **development PR** provided # | not required because: 
 - [ ] **framework PR** provided Mbed-TLS/mbedtls-framework# | not required
+- [ ] **mbedtls PR** provided Mbed-TLS/mbedtls# | not required
 - **tests**  provided | not required because: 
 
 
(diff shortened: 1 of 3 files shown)


.github/pull_request_template.md
```diff
@@ -0,0 +1,29 @@
...
+
+
+
+## PR checklist
+
+Please remove the segment/s on either side of the | symbol as appropriate, and add any relevant link/s to the end of the line.
+If the provided content is part of the present PR remove the # symbol.
+
+- [ ] **changelog** provided | not required because: 
+- [ ] **development PR** provided # | not required because:
```
Harry-Ramsey, line 13: Should we not have a development/3.6 PR field for pull requests that must target MbedTLS and current TF-PSA-Crypto?

ronald-cron-arm: Good point, in the middle/long term, it should be rare that a TF-PSA-Crypto PR breaks mbedtls and there is a need for an mbedtls companion PR, and thus there is not really a need for an mbedtls development PR line here. But in the short term it is more likely that changes in TF-PSA-Crypto break mbedtls:development. Thus adding the line. Otherwise there is no relationship between TF-PSA-Crypto and mbedtls-3.6 but through the framework. But in that case we will rather do the link through the framework PR I'd say.

Harry-Ramsey (approved):
LGTM

mpg (approved):
LGTM, thanks!

PR #124, opened 2024-12-17
Remove USE_PSA from standalone documentation

yanesca:
This is the crypto part of https://github.com/Mbed-TLS/mbedtls/pull/9818. The last commit is the only new commit, the rest is a result of rebasing the original PR.

I checked for broken references like this:
```
grep -r -ho '[a-zA-Z\_-]*\.md' docs | sort | uniq > filelist.txt
cat filelist.txt | xargs -I filename find . -name filename
<manually check that every file is found>
```

### DIFF
--- a/docs/driver-only-builds.md
+++ b/docs/driver-only-builds.md
@@ -200,7 +200,8 @@ consequence these are not supported in builds without `MBEDTLS_ECDSA_C`.
 
 Similarly, there is no PSA support for interruptible ECDH operations so these
 are not supported without `ECDH_C`. See also limitations regarding
-restartable operations in [the documentation about using PSA Crypto](use-psa-crypto.md).
+restartable operations in [the documentation about PSA Crypto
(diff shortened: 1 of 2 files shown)


gilles-peskine-arm (approved):
LGTM based on my [previous review](https://github.com/Mbed-TLS/mbedtls/pull/9818#pullrequestreview-2502614277) of `pr_9818-2` (a874922cfc424a6555807e672be1f193c2843984) and checking the differences in the files that have moved to TF-PSA-Crypto:

```
git diff upstream-crypto/development upstream-crypto/pr/124 -- $(comm -23 <(git diff --name-only upstream-public/development...pr_9818-2 | sort) <(git ls-tree -r --name-only upstream-public/development | sort))
```

The diff is empty, and I did not expect further changes in this file.

waleed-elmelegy-arm (approved):
LGTM

PR #132, opened 2024-12-23
Make legacy declarations private: macro guard

gilles-peskine-arm:
Make legacy functions private by guarding their declaration with a macro.

As an example, do `cipher.h` (functions only, I didn't check if some macros and types should become private).

This is up for design review. There is no update to documentation yet. Some open questions:

* Is this approach ok? It's a quick win for not having legacy functions declared .
* Whether we want to update documentation at the same time we make things private, or later.
* Whether we want to add changelog entries, and whether to do it incrementally or all at once at the end.

Resolves https://github.com/Mbed-TLS/TF-PSA-Crypto/issues/219, ~https://github.com/Mbed-TLS/TF-PSA-Crypto/issues/220~ (scope restricted to 219).

~Prerequisite: https://github.com/Mbed-TLS/mbedtls/pull/9868 to avoid breaking mbedtls.~ (Was only a prerequisite for the part addressing 220.)

## PR checklist

- [x] **changelog** not required because: no user-visible change in this PR
- [x] **mbedtls PR** not required, but see https://github.com/Mbed-TLS/mbedtls/pull/10089 though
- **tests**  provided

### DIFF
--- a/docs/4.0-migration-guide/private-decls.md
+++ b/docs/4.0-migration-guide/private-decls.md
@@ -1,0 +1,34 @@
+## Low-level crypto functions are no longer part of the public API
+
+Low-level crypto functions, that is, all non-PSA crypto functions except a few
+that don't have a proper PSA replacement yet, have been removed from the public
+API.
+
+If your application was using those functions, please see
+`docs/psa-transition.md` (currently in the Mbed TLS repo) for ways to migrate to
+the PSA API instead.
+
+Some of the associated types (for example, `mbedtls_aes_context`) still need to
+be visible to the compiler, so the headers declaring them (for example, `aes.h`)
+are still on the default include path, but we recommend you no longer include
+them directly.
+
+Sample programs have not been fully updated yet and some of them might still
+use APIs that are no longer public. You can recognize them by the fact that they
+define the macro `MBEDTLS_DECLARE_PRIVATE_IDENTIFIERS` (or
+`MBEDTLS_ALLOW_PRIVATE_ACCESS`) at the very top (before including headers). When
+you see one of these two macros in a sample program, be aware it has not been
+updating and parts of it do not demonstrate current practice.
+
+We strongly recommend against defining `MBEDTLS_DECLARE_PRIVATE_IDENTIFIERS` or
+`MBEDTLS_ALLOW_PRIVATE_ACCESS` in your own application. If you do so, you code
+may not compile or work with future minor releases. If there's something you
+want to do that you feel can only be achieved by using one of these two macros,
+please reach out on github or the mailing list.
+
+The following modules had functions removed from the public API:
+- see private-decls/ subdirectory for now - one file per header, to avoid
+  conflicts caused by all PRs editing the same place in this file, to be merged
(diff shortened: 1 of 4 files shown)


drivers/builtin/include/mbedtls/cipher.h
```diff
@@ -1164,6 +1167,9 @@ int mbedtls_cipher_auth_decrypt_ext(mbedtls_cipher_context_t *ctx,
                                     unsigned char *output, size_t output_len,
                                     size_t *olen, size_t tag_len);
 #endif /* MBEDTLS_CIPHER_MODE_AEAD || MBEDTLS_NIST_KW_C */
+
+#endif /* MBEDTLS_DECLARE_PRIVATE_IDENTIFIERS */
```
gilles-peskine-arm, line 1171: It would be convenient to additionally annotate all function declarations between `#if defined(MBEDTLS_DECLARE_PRIVATE_IDENTIFIERS)` and the corresponding `#endif` with `MBEDTLS_DECLARE_PRIVATE_FUNCTION` which can be defined to expand to something like `__attribute__((__deprecated__))`. This would allow detecting the uses of private functions through compiler warnings, and gradually eliminating them by eliminating warnings.

We can add these annotations mechanically, once we've manually added the preprocessor guards.

mpg (approved):
The design looks good to me (and the execution as well).

gilles-peskine-arm:
I'm adding `needs-work` because we should update the documentation.

gilles-peskine-arm:
According to the revised plan in https://github.com/Mbed-TLS/TF-PSA-Crypto/issues/219, https://github.com/Mbed-TLS/mbedtls/pull/9868 needs to go in first.

mpg:
Adopting this PR. I'll need to force-push because I need to rebase on recent development to get the directory for the migration guide. Head before that is e3abfab8edc364b4ecc70145b0af0dab4772bc13

I'll also drop the latest commit that was addressing #220 so that this PR now only addresses #219 (another PR will be raised later to address #220).

I'm adding something to the migration guide, but no changelog entry for this PR as it's not making any user-visible change - that will be done by subsequent PRs (like the one addressing #220) and changelog entries will be added there.

felixc-arm (approved):
LGTM other than nit :+1:

felixc-arm (dismissed):
LGTM :+1:

mpg:
(Btw sorry for the extra force-push: this is my first PR on this repo (!) and I hadn't set up automation for the DCO yet, so the checker was quite correctly complaining that it was missing from the last two commits.)

gabor-mezei-arm (dismissed):
LGTM

mpg:
The CI failures in the merge queue are while checking out the merge commit. Seems spurious. I'll just re-enqueue.

mpg:
One CI failure was real: code-style. The others were just a consequence: when the code-style check failed, github removed this PR from the queue, so the virtual merge commit was removed, so later components could no longer check it out.

In order to fix the code style issue it was convenient for me to rebase over latest development so that I could run the code style script locally.

felixc-arm (approved):
LGTM !

Issue #117, opened 2024-12-11
Add some all.sh components based on baremetal configs 


ronald-cron-arm:
- [ ] Add baremetal and baremetal_size config type in config.py
- [ ] Add build_baremetal component in components-configuration-crypto.sh. Remove component_build_crypto_baremetal in Mbed TLS.
- [ ] Add build_arm_none_eabi_gcc component in components-platform.sh
- [ ] Add build_arm_linux_gnueabi_gcc_arm5vte in components-platform.sh
- [ ] Add build_arm_none_eabi_gcc_arm5vte in components-platform.sh
- [ ] Add build_arm_none_eabi_gcc_m0plus in components-platform.sh
- [ ] Add build_arm_none_eabi_gcc_no_udbl_division in components-platform.sh
- [ ] Add build_arm_none_eabi_gcc_no_64bit_multiplication in components-platform.sh
- [ ] Add build_arm_clang_thumb in components-platform.sh

PR #143, opened 2025-01-09
Enable check names component

### LINKED ISSUE #52: Add name checks
Add name checks

- [ ] Adapt tests/scripts/check_names.py to work for both mbedtls and TF-PSA-Crypto. Keep check of all files when called in the context of mbedtls for the time being.
- [ ] Add component_tf_psa_crypto_check_names in `tf-psa-crypto/tests/scripts/components-basic-checks.sh`.

DOD: Same support for name checking in TF-PSA-Crypto as in mbedtls

Harry-Ramsey:
Enable check names component for TF-PSA-Crypto. Closes #52.

## PR checklist

- [x] **changelog** provided | not required because: testing enhancement.
- [x] **framework PR** provided Mbed-TLS/mbedtls-framework: [#117](https://github.com/Mbed-TLS/mbedtls-framework/pull/117)
- [x] **mbedtls PR** provided Mbed-TLS/mbedtls: [#9915](https://github.com/Mbed-TLS/mbedtls/pull/9915)

## Notes for the submitter

Help make review efficient:
* Multiple simple commits
* Avoid force-push


Harry-Ramsey:
The failures reported here are to do with the test successfully running but have been propogated to TF-PSA-Crypto.

Harry-Ramsey:
> > The failures reported here are to do with the test successfully running but have been propogated to TF-PSA-Crypto.
> 
> I am not sure to understand. What is needed for the new component `tf_psa_crypto_check_names` to run successfully? We cannot merge this PR in that state.

Sorry, there was an extreme number of errors much higher than expected. I forgot to remove code from `check_names.py` which overwrote selected paths resulting in these errors.

ronald-cron-arm:
Regarding 
```
#if defined(MBEDTLS_SSL_PROTO_TLS1_2) && defined(MBEDTLS_USE_PSA_CRYPTO) && \   
    !(defined(PSA_WANT_ALG_SHA_1) || defined(PSA_WANT_ALG_SHA_256) || defined(PSA_WANT_ALG_SHA_512))
#error "MBEDTLS_SSL_PROTO_TLS1_2 defined, but not all prerequisites"            
#endif 
```
in ./drivers/builtin/src/check_crypto_config.h, this should rather be in `mbedtls:include/mbedtls/check_config.h`. No need for `defined(MBEDTLS_USE_PSA_CRYPTO)` as MBEDTLS_USE_PSA_CRYPTO is always on now. Thus just:
```
#if defined(MBEDTLS_SSL_PROTO_TLS1_2) && \   
    !(defined(PSA_WANT_ALG_SHA_1) || defined(PSA_WANT_ALG_SHA_256) || defined(PSA_WANT_ALG_SHA_512))
#error "MBEDTLS_SSL_PROTO_TLS1_2 defined, but not all prerequisites"            
#endif
```

ronald-cron-arm:
Regarding MBEDTLS_ERR_SSL_CRYPTO_IN_PROGRESS in `crypto_config.h` and `ecp.h` I think we can just remove the references to MBEDTLS_ERR_SSL_CRYPTO_IN_PROGRESS. Looking in ssl.h at the documentation of the TLS functions that return MBEDTLS_ERR_SSL_CRYPTO_IN_PROGRESS I do not feel the need to change their documentation because of the removals of the comments in TF-PSA-Crypto.

ronald-cron-arm:
Regarding psa_util_internal.h:
```
#if defined(MBEDTLS_USE_PSA_CRYPTO) || defined(MBEDTLS_SSL_PROTO_TLS1_3)
extern const mbedtls_error_pair_t psa_to_ssl_errors[7];
#endif
```
can be moved to `library/ssl_misc.h`. It is defined in `library/ssl_tls.c`.

ronald-cron-arm:
In `entropy_poll.c`
```
#if defined(MBEDTLS_TIMING_C)
#include "mbedtls/timing.h"
#endif
```
can just be removed it seems.

ronald-cron-arm:
In crypto_config.h the note:
```
 * \note If MBEDTLS_TIMING_C is set - to enable the semi-portable timing
 *       interface - timing.c will include time.h on suitable platforms
 *       regardless of the setting of MBEDTLS_HAVE_TIME, unless
 *       MBEDTLS_TIMING_ALT is used. See timing.c for more information.
```
can be just removed. We already have the same note in MBEDTLS_TIMING_C documentation.

Harry-Ramsey:
Just need to create an Mbed TLS pull request where I add the necessary checks and this can be reviewed.

ronald-cron-arm:
Regarding the MBEDTLS_SSL_MAX_... symbols in cipher.h, after some discussions with @gilles-peskine-arm,I propose to remove the comments in cipher.h and add some unit tests in test_suite_ssl that assert that `MBEDTLS_SSL_MAX... >= MBEDTLS_MAX_...`. A `static_checks()` test function similar to the one in test_suite_psa_crypto.function should do the job.

ronald-cron-arm:
Do you mean: `MBEDTLS_SSL_MAX_BLOCK_LENGTH >= PSA_BLOCK_CIPHER_BLOCK_MAX_SIZE` (greater or equal than)?

gilles-peskine-arm:
Oops, yes, I had the comparisons the wrong way round. I edited my comment.

mpg:
I don't think those tests would be logically correct. It may very well be that in the future we'll add a cipher with variable IV length and TLS only uses a certain IV size that's smaller than the maximum, and then `SSL_MAX_IV_LENGTH` will be smaller than `PSA_CIPHER_IV_MAX_SIZE` and that would be perfectly OK. Similar thought about key length, and block size (we could have a cipher supported in PSA that's not used by any TLS ciphersuite).

So I'm actually opposed to adding those assertions.

ronald-cron-arm:
> I don't think those tests would be logically correct. It may very well be that in the future we'll add a cipher with variable IV length and TLS only uses a certain IV size that's smaller than the maximum, and then `SSL_MAX_IV_LENGTH` will be smaller than `PSA_CIPHER_IV_MAX_SIZE` and that would be perfectly OK. Similar thought about key length, and block size (we could have a cipher supported in PSA that's not used by any TLS ciphersuite).
> 
> So I'm actually opposed to adding those assertions.

Thanks, I think I understand better the ins and outs of the MBEDTLS_SSL_MAX_ macros now. If I understand correctly, the comments in cipher.h stating that MBEDTLS_MAX_ macros should be kept in sync with the MBEDTLS_SSL_MAX_ ones in ssl_misc.h are actually at least misleading thus we are just going to remove them.

include/psa/crypto_config.h
```diff
@@ -1675,7 +1665,7 @@
  *
  * Uncomment this macro to enable restartable ECC computations.
  */
-//#define MBEDTLS_ECP_RESTARTABLE
+// #define MBEDTLS_ECP_RESTARTABLE
```
valeriosetti, line 1668:
```suggestion
//#define MBEDTLS_ECP_RESTARTABLE
```

tests/scripts/components-basic-checks.sh
```diff
@@ -117,3 +117,10 @@ component_tf_psa_crypto_check_test_helpers () {
     msg "unit test: translate_ciphers.py"
     python3 -m unittest framework/scripts/translate_ciphers.py 2>&1
 }
+
+component_tf_psa_crypto_check_names () {
+    msg "Check: declared and exported names (builds the library)" # < 3s
+    # Make generated files required for check
+    cmake -DGEN_FILES=ON .
```
Harry-Ramsey, line 124: Currently an error with this line, I believe this locks the Makefile resulting in a faliure down the line.

ronald-cron-arm: Yes I understand this issue. Thinking about it ...

include/psa/crypto_config.h
valeriosetti: Also the commit message that introduces this change seems unrelated with the change itself

tests/scripts/components-basic-checks.sh
ronald-cron-arm: To unblock things, I propose for the time being to do just:
```suggestion
    ./scripts/generate_driver_wrappers.py
```
This seems to work and we can revisit that when we thing more generally about the generated files in the context of TF-PSA-Crypto.

include/psa/crypto_config.h
```diff
@@ -1665,15 +1665,15 @@
...
 
 /**
  * Uncomment to enable using new bignum code in the ECC modules.
  *
  * \warning This is currently experimental, incomplete and therefore should not
  * be used in production.
  */
-//#define MBEDTLS_ECP_WITH_MPI_UINT
+// #define MBEDTLS_ECP_WITH_MPI_UINT
```
ronald-cron-arm, line 1676:
```suggestion
//#define MBEDTLS_ECP_WITH_MPI_UINT
```

ronald-cron-arm (changes requested):
Only one small thing, one comment from @valeriosetti to address as well I think.

include/psa/crypto_config.h
Harry-Ramsey: Thanks, I dropped the commit containing those changes as that is all it appeared to have.

include/psa/crypto_config.h
Harry-Ramsey: Thanks, I dropped the commit containing those changes as that is all it appeared to have.

ronald-cron-arm (approved):
LGTM

drivers/builtin/include/mbedtls/oid.h
```diff
@@ -628,7 +628,7 @@ int mbedtls_oid_get_md_hmac(const mbedtls_asn1_buf *oid, mbedtls_md_type_t *md_h
  */
 int mbedtls_oid_get_md_alg(const mbedtls_asn1_buf *oid, mbedtls_md_type_t *md_alg);
 
-#if !defined(MBEDTLS_X509_REMOVE_INFO)
+#if 1 /* OID_INFO_STRINGS */
```
valeriosetti, line 631: Are we going to keep this `#if 1`? Can't we just remove it?
If so, then there are other occurrences also in `oid.c`.

drivers/builtin/include/mbedtls/config_adjust_legacy_crypto.h
```diff
@@ -92,10 +92,7 @@
     defined(MBEDTLS_ENTROPY_C) || \
     defined(MBEDTLS_PK_C) || \
     defined(MBEDTLS_PKCS12_C) || \
-    defined(MBEDTLS_RSA_C) || \
-    defined(MBEDTLS_SSL_TLS_C) || \
-    defined(MBEDTLS_X509_USE_C) || \
-    defined(MBEDTLS_X509_CREATE_C)
```
valeriosetti, line 98: The commit says
`This commit removes MBEDTLS_SSL_TLS_C, MBEDTLS_X509_USE_C and
MBEDTLS_X509_CREATE_C from TF-PSA-Crypto to Mbed TLS where they more
appropriately belong.`

but I cannot see them being added somewhere in https://github.com/Mbed-TLS/mbedtls/pull/9915. Am I missing something?

valeriosetti (changes requested):

drivers/builtin/include/mbedtls/oid.h
Harry-Ramsey: Since oid.c will be reworked at some point, the decision was to replace this macro with 1 instead to preserve information but enable `check_names` to work without reporting errors.

drivers/builtin/include/mbedtls/config_adjust_legacy_crypto.h
ronald-cron-arm: See https://github.com/Mbed-TLS/mbedtls/pull/9915#issuecomment-2630361279 explaining why we can actually just remove `defined(MBEDTLS_X509_USE_C) || defined(MBEDTLS_X509_CREATE_C)` here. Regarding MBEDTLS_SSL_TLS_C, it depends on MBEDTLS_MD_C as stated in mbedtls_config.h thus it can be just removed as well.

drivers/builtin/include/mbedtls/oid.h
ronald-cron-arm: We prefer to keep the information in the code of where the "!defined(MBEDTLS_X509_REMOVE_INFO)" were to ease the rework of the oid.c APIs (see https://github.com/Mbed-TLS/mbedtls/issues/9380).

valeriosetti (approved):

drivers/builtin/include/mbedtls/config_adjust_legacy_crypto.h
Harry-Ramsey: Those commits were dropped from MbedTLS as I believe x509 doesn't have a dependency on them afterall, this commit message needs to be changed to reflect the real reason now :)

ronald-cron-arm: We have discussed the various issues in several places, I completely understand it is not easy to follow.

valeriosetti (approved):

ronald-cron-arm (approved):
LGTM

drivers/builtin/include/mbedtls/oid.h
gilles-peskine-arm: We're actually not redesigning the OID API in TF-PSA-Crypto 1.0: it's becoming private. So I've created a separate task for handling the split of the effect of `MBEDTLS_X509_REMOVE_INFO`: https://github.com/Mbed-TLS/mbedtls/issues/10124. Fortunately it seems that none of the OIDs used in crypto need a description any more.

PR #145, opened 2025-01-09
Plans for TF-PSA-Crypto legacy headers

gilles-peskine-arm:
Resolves https://github.com/Mbed-TLS/mbedtls/issues/9882 (not completely, but separate issues have been created for the remaining points).

Status: the architectural overview is complete. Some parts need further refinement and many parts need task breakdown. In many places, the task breakdown needs input from prioritization, which needs input from the architectural overview, and thus is beyond the scope of this issue. Thus this document fulfills #9882.

Resolves https://github.com/Mbed-TLS/mbedtls/issues/8663.

## PR checklist

- [x] **changelog** not required because: doc only
- [x] **mbedtls PR** not required


docs/architecture/0e-plans.md
```diff
@@ -0,0 +1,474 @@
...
+Prototype: https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/132
+
+Thus, for the 0ε milestone, in each affected header, we just need to place `#ifdef` guards around the function declarations.
+
+Rationale:
+
+* This is a small amount of work.
+* This requires few changes to the code, and they are very localized, so it will not disrupt other work happening in parallel.
+
+TODO: this leaves Doxygen comments around
```
mpg, line 267: I think Doxygen obeys `#ifdef` (otherwise we wouldn't bother with `config.py realfull` in `apidoc_full.sh`), so can we find a way to ensure `MBEDTLS_DECLARE_PRIVATE_IDENTIFIERS` is not defined when we run doxygen? Like, in `common.h`:
```c
#if !defined(__DOXYGEN__)
#define MBEDTLS_DECLARE_PRIVATE_IDENTIFIERS
#endif
```
(Note: `__DOXYGEN__` is defined by `doxygen/mbedtls.doxyfile`.)

Wdyt?

docs/architecture/0e-plans.md
```diff
@@ -0,0 +1,474 @@
...
+
+#### Separating private interfaces from exposed interfaces
+
+If a private interface of TF-PSA-Crypto is declared in an exposed header, in the medium term, we should move it to a private header. Note that we can only do that if the interface is not called from Mbed TLS code.
+
+When none the functions declared in an exposed header file are either public or called by Mbed TLS, we can move all the declarations gated by `MBEDTLS_DECLARE_PRIVATE_IDENTIFIERS` to a private header file and adjust `#include` directives accordingly. This can be done by a script.
```
mpg, line 273: Wording: I found "when none [...] are either [...] or [...]" hard to parse. (I you're curious why, I think I interpreted "either A or B" as "A or B but not both" then putting "none" in front resulted in a double negation and that was too much for me. On 2nd though, I think that's not that either means here, but rather "when all of the functions [...] are neither public nor used by Mbed TLS"?)

Still, I'm not sure I get the first part of the sentence: since the rest of the sentence is about moving functions gated by `MBEDTLS_DECLARE_PRIVATE_IDENTIFIERS`, which by definition are not public, why mention public in the first place?

All this to say, I'm a bit confused, can you clarify?

docs/architecture/0e-plans.md
```diff
@@ -0,0 +1,474 @@
...
+pem.h
+pk.h
+platform.h
+platform_time.h
+platform_util.h
+psa_util.h
+threading.h
+```
+
+#### Headers that remain public for private macros
```
mpg, line 298: s/public/exposed/?

docs/architecture/0e-plans.md
```diff
@@ -0,0 +1,474 @@
...
+
+The following headers solely define exposed macros, and must remain exposed. They can be
```
mpg, line 300: Unfinished sentence: "They can be"

docs/architecture/0e-plans.md
```diff
@@ -0,0 +1,474 @@
...
+```
+build_info.h
+config_adjust_legacy_from_psa.h
+config_adjust_psa_superset_legacy.h
+config_adjust_test_accelerators.h
+config_psa.h
+private_access.h
+```
+
+#### Privatization of built-in cryptographic headers with context types
```
mpg, line 311: Minor: does not follow the same pattern "Headers that" as the sections before and after.

docs/architecture/0e-plans.md
```diff
@@ -0,0 +1,474 @@
...
+
+Main loss of functionality:
+
+* CTR\_DRBG in itself (i.e. other than for the PSA Crypto RNG instance). We intend to restore this functionality through a PSA API, but the API isn't designed yet, so this will happen after 1.0 and not with the existing API.
+* Direct access to entropy sources. We've decided that this is acceptable.
+* The ability to configure entropy sources on a platform. This is not an acceptable loss. In the long term (likely after 1.0), this will be resolved by the PSA crypto random driver API. In the short term, we will expose a modified `mbedtls_hardware_poll()` (https://github.com/Mbed-TLS/mbedtls/issues/9618).
+
+#### Privatization of Everest headers
+
+Everest headers (`drivers/everest/include/everest/include/**/*.h`) contain some exposed types: they are exposed via `mbedtls_ecdh_context` from `mbedtls/ecdh.h` which is exposed via `mbedtls_psa_key_agreement_interruptible_operation_t` indirectly from `psa/crypto.h`. The rest of their content is private (to be consumed only by `ecdh.c`) or internal (to be consumed only by `everest/**/*.c`) definitions.
```
mpg, line 393: Off-topic: in the long run, the built-in driver for ECDH (interruptible or not) should really call `ecp_mul` directly, just like the built-in driver for FFDH calls `exp_mod` directly, and then we just remove `ecdh.c` and `ecdh.h`.

mpg (commented):
First draft is looking pretty good so far, thanks!

docs/architecture/0e-plans.md
gilles-peskine-arm: I meant, but forgot to write: Doxygen comments in the rest of the file, e.g. the `\file` comment and comments on the exposed types.

docs/architecture/0e-plans.md
```diff
@@ -0,0 +1,474 @@
...
+
+* `lms.h`: The PSA API does not support stateful signatures yet. This is planned, but the API design is still under discussion. This is a critical feature in TF-A, hence considered necessary in the TF-PSA-Crypto MVP.
+* `nist_kw.h`: The PSA API does not have an encoding for KW. It is under discussion, but it is mostly used to wrap key material or blobs containing key material, rather than to directly manipulate text, which complicates the API design. This is a request from many silicon vendors, hence considered necessary in the TF-PSA-Crypto MVP.
+
+#### Cryptography-adjacent headers
+
+The following header files define cryptography-adjacent interfaces which we have no plans to replace.
+
+* `asn1.`, `asn1write.h`: ASN.1, needed for key parsing/writing as well as for X.509.
+* `base64.h`, `pem.h`: Encoding help, needed for key parsing/writing as well as for X.509. Arguably Base64 could be made private, for the use of PEM only, but it is currently used for non-PEM purposes in Mbed TLS. Thus I propose to keep it officially public in TF-PSA-Crypto 1.x.
```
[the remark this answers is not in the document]
gilles-peskine-arm: Good point, I checked base64 but went by our previous assessment (which lands on the safe but perhaps suboptimal side) for PEM. I'll revise.

docs/architecture/0e-plans.md
```diff
@@ -0,0 +1,474 @@
...
+| Header | Function prefix | Fate | Notes |
+| ------ | --------------- | ---- | ----- |
+| `aes.h` | `mbedtls_aes_` | Expose | [context types](#privatization-of-built-in-cryptographic-headers-with-context-types) |
+| `aria.h` | `mbedtls_aria_` | Expose | [context types](#privatization-of-built-in-cryptographic-headers-with-context-types) |
+| `asn1.h` | `mbedtls_asn1_` | Public | [cryptography-adjacent](#cryptography-adjacent-headers) |
+| `asn1write.h` | `mbedtls_asn1_write_` | Public | [cryptography-adjacent](#cryptography-adjacent-headers) |
+| `base64.h` | `mbedtls_base64_` | Public | [cryptography-adjacent](#cryptography-adjacent-headers) |
+| `bignum.h` | `mbedtls_mpi_` | Expose | [context types](#privatization-of-built-in-cryptographic-headers-with-context-types) |
+| `block_ciper.h` | `mbedtls_block_cipher_` | Expose | [context types](#privatization-of-built-in-cryptographic-headers-with-context-types) |
+| `build_info.h` | `MBEDTLS_` | Exposed | [Only for private macros ](#headers-that-remain-public-for-private-macros) |
```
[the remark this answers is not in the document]
gilles-peskine-arm: Hmm, now that you mention it, the job of this `build_info.h` isn't exactly to define macros, even indirectly. That's the job of `<tf-psa-crypto/build_info.h>` now. I seem to recall that #9862 was the plan, although I couldn't find an issue about it.

docs/architecture/0e-plans.md
```diff
@@ -0,0 +1,474 @@
...
+| `cipher.h` | `mbedtls_cipher_` | Expose | [context types](#privatization-of-built-in-cryptographic-headers-with-context-types) |
+| `cmac.h` | `mbedtls_cipher_cmac_` | Expose | [context types](#privatization-of-built-in-cryptographic-headers-with-context-types) |
+| `config_adjust_*.h` | N/A | Exposed | [Only for private macros ](#headers-that-remain-public-for-private-macros) |
+| `config_psa.h` | N/A | Exposed | [Only for private macros ](#headers-that-remain-public-for-private-macros) |
+| `constant_time.h` | `mbedtls_ct_` | Public | [cryptography-adjacent](#cryptography-adjacent-headers) |
+| `ctr_drbg.h` | `mbedtls_ctr_drbg_` | Private | [RNG header privatization](#rng-header-privatization) |
+| `des.h` | `mbedtls_des_` | Expose | [context types](#privatization-of-built-in-cryptographic-headers-with-context-types) |
+| `dhm.h` | `mbedtls_dhm_` | Private | [can be made fully private](#headers-that-can-be-made-fully-private) |
+| `ecdh.h` | `mbedtls_ecdh_` | Expose | [context types](#privatization-of-built-in-cryptographic-headers-with-context-types) |
```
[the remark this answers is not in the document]
gilles-peskine-arm: All the legacy crypto headers can be fully removed after they're made private. I don't see how `ecdh.h` is different from others: it's used for key agreement in PSA, including the interruptible version which uses `mbedtls_ecdh_context` in an operation structure.

docs/architecture/0e-plans.md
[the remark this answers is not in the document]
gilles-peskine-arm: Actually, the reason for our previous assessment still stands: eventually we would like Mbed TLS to only rely on the public APIs of TF-PSA-Crypto. This has to include PEM. So the question isn't whether TF-PSA-Crypto should have a PEM API, but whether we keep the current PEM API public or hide it and make a possibly different PEM API public later.

docs/architecture/0e-plans.md
gilles-peskine-arm: I don't think we'd need a `!__DOXYGEN__` guard: we wouldn't be feeding `#define MBEDTLS_DECLARE_PRIVATE_IDENTIFIERS` to Doxygen since that would only appear when building library files.

gilles-peskine-arm:
Thanks for the review. I've addressed your comments, but not yet added the missing sections that I'm still working on.

docs/architecture/0e-plans.md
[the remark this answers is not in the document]
mpg: I think it is different, but that's off-topic so let's leave it there.

docs/architecture/0e-plans.md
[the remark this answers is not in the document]
mpg: Indeed. I'm not sure what the current status is, but I remember being deeply unsatisfied with our base64 API last time I looked at it (to the point of starting trying to fix it locally, which I evidently never got time to finish). The main problem being we're pretty inconsistent in where the reported sizes include the terminating nul byte or not, and things are not clearly documented. I don't remember if PEM suffer from similar issues these days.

Taking a step back, it could be argued that neither PEM nor base64 should rely on nul-termination, but rather only on size arguments.

I'm not sure if that's enough of a reason to not include them in 1.0 and re-include a better version later. (Obviously in an ideal world we'd have time to improve their API before the 1.0 and things would be easier, but we don't live in that world.)

mpg (commented):
Thanks for addressing my comments!

docs/architecture/0e-plans.md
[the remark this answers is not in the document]
gilles-peskine-arm: Point noted about being dissatisfied with the current API (I share your dissatisfaction). That drives us towards making the APIs private in 1.0 and making an improved version public later. The question is, is there a hope that we will have time to make an _improved_ version public? If we think that we'll again have no time for anything but taking the existing interface unchanged, then we might as well give up now.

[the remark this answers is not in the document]
gilles-peskine-arm: I've updated the document to include the consideration of applications that want PEM (unusual but I found a couple) or Base64 (less rare, but also less justifiable in the sense that PEM is pretty much always related to crypto, whereas Base64 isn't).

Our position on removing interfaces that are currently used by applications can be “wait until they complain”. But if we know to expect complaints, we might as well not wait and just save energy for everyone by keeping the interface public.

docs/architecture/0e-plans.md
```diff
@@ -0,0 +1,536 @@
...
+
+The ASN.1 interfaces use `mbedtls_mpi` for INTEGER parsing/writing. This must change to a byte array. This is filed as https://github.com/Mbed-TLS/mbedtls/issues/9373 and https://github.com/Mbed-TLS/mbedtls/issues/9372 .
+
+### Private types in `nist_kw.h`
+
+`nist_kw.h` must switch from a legacy cipher ID to a PSA key type: https://github.com/Mbed-TLS/mbedtls/issues/9382.
+
+### Private types in `psa_util.h`
+
+The functions `mbedtls_ecc_group_to_psa()` and mbedtls_ecc_group_from_psa()` are no longer relevant for public use since the legacy side of the conversion is no longer a public interface. They are not used in Mbed TLS. They should be moved to an internal header.
```
davidhorstmann-arm, line 478: Nit: missing backtick
```suggestion
The functions `mbedtls_ecc_group_to_psa()` and `mbedtls_ecc_group_from_psa()` are no longer relevant for public use since the legacy side of the conversion is no longer a public interface. They are not used in Mbed TLS. They should be moved to an internal header.
```

docs/architecture/0e-plans.md
```diff
@@ -0,0 +1,536 @@
...
+| `ecdsa.h` | `mbedtls_ecdsa_` | Expose | [context types](#headers-with-context-types) |
+| `ecjpake.h` | `mbedtls_ecjpake_` | Expose | [context types](#headers-with-context-types) |
+| `ecp.h` | `mbedtls_ecp_` | Expose | [context types](#headers-with-context-types) |
+| `entropy.h` | `mbedtls_entropy_` | Private | [RNG header privatization](#rng-header-privatization) |
+| `error_common.h` | `mbedtls_*err*` | Private | TODO |
+| `gcm.h` | `mbedtls_gcm_` | Expose | [context types](#headers-with-context-types) |
+| `hkdf.h` | `mbedtls_hkdf_` | Delete | https://github.com/Mbed-TLS/mbedtls/issues/9150 |
+| `hmac_drbg.h` | `mbedtls_hmac_drbg_` | Private | [can be made fully private](#headers-that-can-be-made-fully-private) with a little work for [RNG header privatization](#rng-header-privatization) |
+| `lms.h` | `mbedtls_lms_` | Public | [no PSA equivalent](#cryptographic-mechanisms-with-no-PSA-equivalent) |
+| `md.h` | `mbedtls_md_` | Expose | [context types](#headers-with-context-types), but likely [Public hash-only `md.h`](#public-hash-only-md.h) |
```
davidhorstmann-arm, line 174: Nit: The link doesn't work, maybe:
```suggestion
| `md.h` | `mbedtls_md_` | Expose | [context types](#headers-with-context-types), but likely [Public hash-only `md.h`](#changes-to-public-crypto-headers) |
```

docs/architecture/0e-plans.md
```diff
@@ -0,0 +1,536 @@
...
+* In the Mbed TLS sample program `ssl_mail_client`, for a tiny bit of SMTP that requires Base64.
+* In application code, sometimes, for miscellaneous things, often not directly related to cryptography. On the one hand, many of these uses are out of scope. On the other hand, since TF-PSA-Crypto has a Base64 implementation anyway, users who like TF-PSA-Crypto for its small code size would be justifiably disappointed not to have a Base64 interface.
+
+TODO: decide.
+
+#### OID interface
+
+`oid.h` and `MBEDTLS_OID_C` are in charge of defining all the OIDs used internally or by application code. This is a known problem which can result in applications wasting code size on many OIDs that they don't care about. We are planning a redesign: https://github.com/Mbed-TLS/mbedtls/issues/9380 .
+
+At this point, we do not consider this critical for TF-PSA-Crypto 1.0, and we are likely to work on it during the lifetime of TF-PSA-Crypto 1.x. This may change if we reevaluate the priority based on concrete use cases. In the meantime, `oid.h` will become private, but will remain used in Mbed TLS.
```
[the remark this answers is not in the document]
gilles-peskine-arm: I've tried to consistently use “private” for unofficial interfaces that may still be exposed, and “internal” for interfaces where there is some technical measure limiting outside use (e.g. function only declared in a header in `core` or `**/src`). So there's no contradiction here: `oid.h` is not an official interface, but it's exposed to Mbed TLS. Did I get the terminology wrong somewhere above?

docs/architecture/0e-plans.md
```diff
@@ -206,10 +206,34 @@ The header files listed in this section define cryptographic mechanisms which do
...
+
+PEM is used:
+
+* Inside the crypto library, to parse and write keys.
+* Inside the Mbed TLS library, to parse and write X.509 objects.
+* In application code, very occasionally. (Examples: [Fire-evacuation-guidance-system-IoT](https://github.com/2nd-Chance/Fire-evacuation-guidance-system-IoT/blob/33031a8255fe1ae516ddd58f1baa808801cd3abf/iotivity/resource/csdk/security/src/credresource.c#L3185) (dead project), [SiLabs Bluetooth attestation server](https://github.com/SiliconLabs/bluetooth_applications/blob/3eb0f3c9e234ada1f10714fb9376fcbc8e95807f/bluetooth_secure_attestation/bt_secure_attestation_server/src/ecdh_util3.c#L375))
+
+The use of PEM inside the Mbed TLS is intrinsic. It doesn't leak through the API of Mbed TLS, but Mbed TLS cannot be implemented without PEM. This is different from other private modules that Mbed TLS currently calls internally, but will no longer need to call once Mbed TLS has fully migrated to PSA. Thus, in the long term, TF-PSA-Crypto needs to expose its PEM API to Mbed TLS. (We reject the hypotheses of independent PEM implementations, or of making PEM its own library, as too much maintenance work.)
```
mpg, line 229: Nit: "the Mbed TLS"

docs/architecture/0e-plans.md
```diff
@@ -206,10 +206,34 @@ The header files listed in this section define cryptographic mechanisms which do
...
+
+The current PEM interface is unsatisfactory. We would like to improve it (https://github.com/Mbed-TLS/mbedtls/issues/9374) but it is unlikely that we will have enough bandwidth to do so before the 1.0 release. We need to make the PEM interface public to reach the milestone where Mbed TLS stops relying on private interfaces of TF-PSA-Crypto. We can choose to make it public now, or wait until later. Waiting is only advantageous if we believe that we will have enough bandwidth to actually clean up the PEM interface.
+
+Base64 is used:
+
+* Inside the crypto library, to implement PEM.
+* In the `pem2der` sample program.
+* In Mbed TLS SSL test programs, for context serialization. (It's not clear to me why we encode in Base64 rather than binary.)
```
mpg, line 237: Not clear to me either, seems completely arbitrary to me. (And if we want something printable, we already have utility functions for hex in the test programs and could use that instead.)

docs/architecture/0e-plans.md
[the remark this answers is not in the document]
mpg: Thanks for the update, especially searching for uses in applications. I agree that users who pick us for our small code size don't want to have to use an extra base64 implementation while we already have one.

Considering this, I'm now inclined to keep the interface public. I don't think removing now to make an improved version public later would be a good strategy, this is almost sure to annoy users and we're likely to end up adding back a non-improved version due to a combination of time pressure and other priorities.

I guess our dissatisfaction with the current API should go through the normal process of either adding new functions with a better API and deprecating the old ones (any time in 1.x) or reworking the API in 2.0.

docs/architecture/0e-plans.md
```diff
@@ -0,0 +1,543 @@
...
+| `poly1305.h` | `mbedtls_poly1305_` | Expose | [context types](#headers-with-context-types) |
+| `private_access.h` | N/A | Exposed | [Only for exposed macros ](#headers-that-remain-public-for-exposed-macros) |
+| `psa_util.h` | N/A | Public | [remains public](#headers-that-remain-public) but see [Private types in `psa_util.h`](#private-types-in-psa_utilh) |
+| `ripemd160.h` | `mbedtls_ripemd160_` | Expose | [context types](#headers-with-context-types) |
+| `rsa.h` | `mbedtls_rsa_` | Private | [can be made fully private](#headers-that-can-be-made-fully-private) with a little work (TODO) |
+| `sha1.h` | `mbedtls_sha1_` | Expose | [context types](#headers-with-context-types) |
+| `sha256.h` | `mbedtls_sha256_` | Expose | [context types](#headers-with-context-types) |
+| `sha3.h` | `mbedtls_sha3_` | Expose | [context types](#headers-with-context-types) |
+| `sha512.h` | `mbedtls_sha512_` | Expose | [context types](#headers-with-context-types) |
+| `threading.h` | `mbedtls_threading_` | Public | [Platform headers](#platform-headers) |
```
davidhorstmann-arm, line 195: This list is missing `timing.h` I believe

mpg: Is it, though? I'm not seeing any `timing.h` in `drivers/builtin/include/mbedtls/` in today's development, and apparently it was already not there in e35785c3c15924d1a7024a4b8695ce32ed186bf9 (when the repo was created).

gilles-peskine-arm: It was briefly in crypto, but already moved back to mbedtls before the repo split.

docs/architecture/0e-plans.md
```diff
@@ -590,3 +590,24 @@ https://github.com/Mbed-TLS/mbedtls/issues/8452
...
+| `mbedtls_psa_get_random` | crypto internal, `ssl_test_lib.c` | Deprecate |
+| `MBEDTLS_PSA_RANDOM_STATE` | same as `mbedtls_psa_get_random` | Deprecate |
+| `mbedtls_ecc_group_to_psa` | crypto internal | Internalize |
+| `mbedtls_ecc_group_from_psa` | crypto internal | Internalize |
+| `mbedtls_md_psa_alg_from_type` | crypto internal, SSL library | Privatize, deprecate or remove? |
+| `mbedtls_md_type_from_psa_alg` | crypto internal, SSL library | Privatize, deprecate or remove? |
+| `mbedtls_ecdsa_raw_to_der` | crypto internal | Keep public |
+| `mbedtls_ecdsa_der_to_raw` | crypto internal | Keep public |
+
+The RNG wrapper function `mbedtls_psa_get_random` (and the associated constant `MBEDTLS_PSA_RANDOM_STATE`) has a negligible cost for us and can help users transition their code. Keep it, but mark it as deprecated, possibly inlined, and change it (code and documentation) to just return the return value of `psa_get_random`.
```
mpg, line 606: > can help users transition their code

What do you have in mind here? My thinking is this function's primary purpose was it could be used as the `f_rng` argument of functions that have such a parameter. But we'll be removing all such parameters, so once that's done how will this function still be helpful to users?

docs/architecture/0e-plans.md
```diff
@@ -590,3 +590,24 @@ https://github.com/Mbed-TLS/mbedtls/issues/8452
...
+
+The ECC group conversion functions are now purely private since the type `mbedtls_ecp_group_id` is private. They aren't used by Mbed TLS, so move their declaration to an internal header.
+
+The digest algorithm conversion functions are heavily used in TLS code (mostly if not exclusively TLS 1.3). They perform a very simple conversion, so it should be easy to get rid of them. However, as long as `mbedtls_md_type_t` still exists (see [Public hash-only `md.h`](public-hash-only-mdh)), it could make sense to have them as public functions in `md.h`.
```
mpg: Yes, I think it makes sense to keep them in `md.h` as long as it exists (that is, until 5.0).

docs/architecture/0e-plans.md
gilles-peskine-arm: Anything that happens in X.509 or TLS as a result of API changes in crypto, is also something that's likely to happen in code that implements other protocols. Maintainers of third-party code bases that implement other protocols may want to start using TF-PSA-Crypto 1.0 before they have fully converted their code. They may even want to support both Mbed TLS 3.6 and TF-PSA-Crypto in the same code branch, if they deploy the same application to devices which are not all upgraded to the latest crypto at the same time. So they may want to pass `f_rng` arguments around in their own code, even if they end up being unused when building against TF-PSA-Crypto.

docs/architecture/0e-plans.md
```diff
@@ -568,6 +568,513 @@ Note that if we want to be able to remove all error code translations (to save c
...
+* A library build (the most common case) defines `MBEDTLS_PSA_CRYPTO_C` (which implies `MBEDTLS_PSA_CRYPTO_CLIENT`).
+* A client-only build defines `MBEDTLS_PSA_CRYPTO_CLIENT` and little to no else.
+* A server-only build defines `MBEDTLS_PSA_CRYPTO_C` and `MBEDTLS_PSA_CRYPTO_SPM`.
+
+#### Revise key store selection
+
+For historical reasons, we have three key store architectures:
+
+* Purely static, when `MBEDTLS_PSA_STATIC_KEY_SLOTS` is enabled. Key slots are allocated statically, and the core does not use `malloc()`. This only allows a small number of keys determined at compile time, and keys have a size limit that is determined at compile time. This achieves the smallest code size, but has major limitations.
+* Hybrid, when neither `MBEDTLS_PSA_KEY_STORE_DYNAMIC` nor `MBEDTLS_PSA_STATIC_KEY_SLOTS` is enabled. This is the original design, with a static array of key slots (for simplicity) and dynamic allocation of key data. This only allows a small number of keys determined at compile time, but achieves a small code size (almost on par with the static key store if `malloc()` is present anyway).
```
davidhorstmann-arm, line 1073: Does the hybrid architecture have any advantages over purely static? For example, maybe it has smaller runtime memory usage while retaining a smaller code size.

If it doesn't then perhaps it would make sense to remove this mode, as it's a bit of a historical accident.

davidhorstmann-arm (changes requested):
A few questions and small things, but on the whole I'm happy with what is there. I note that there are various TODOs and gaps currently but I expect they'll be filled as the design continues.

docs/architecture/0e-plans.md
gilles-peskine-arm: Expanded. All three have plausible scenarios where they're better than the other two.

davidhorstmann-arm (approved):
LGTM for what is there (obviously there are some TODOs)

docs/architecture/0e-plans.md
```diff
@@ -568,6 +568,513 @@ Note that if we want to be able to remove all error code translations (to save c
...
+
+Currently, `MBEDTLS_ASN1_PARSE_C` and `MBEDTLS_ASN1_WRITE_C` are automatically enabled:
+
+* If `MBEDTLS_RSA_C` is enabled. This is needed for RSA key import and export. We don't try to optimize code size when import or export of a key type is disabled, so this can stay. As part of internal refactoring, it will change to `PSA_WANT_KEY_TYPE_RSA_PUBLIC_KEY` eventually.
+* If `PSA_WANT_ALG_ECDSA` or `PSA_WANT_ALG_DETERMINISTIC_ECDSA` is built in. This used to be required to convert between ASN.1 and raw signature formats internally back when PSA was built on top of the legacy API, but today I think it isn't needed. However, it is needed for `MBEDTLS_ECDSA_C`. We do expose the signature conversion functions `mbedtls_ecdsa_raw_to_der` and `mbedtls_ecdsa_der_to_raw` when an ECDSA signature is enabled through PSA (built-in or not), but we automatically enable the necessary ASN.1 functions regardless of the ASN.1 configuration.
+
+So we have an optimization here to not auto-enable all of ASN.1 for built-in ECDSA. This will require removing or changing the guards on `mbedtls_ecdsa_read_signature()` and `mbedtls_ecdsa_write_signature()` and friends, which are now private functions.
```
mpg, line 1054: Just to be sure: to me this reads like no action required for $0\epsilon$ - just a possible improvement for 1.0 or even 1.x. Right?

mpg (commented):
Looking good to me, only minor notes.

We'll need to come back to the TODOs of course.

docs/architecture/0e-plans.md
```diff
@@ -568,6 +568,513 @@ Note that if we want to be able to remove all error code translations (to save c
...
+| `MBEDTLS_CTR_DRBG_C` | TBD | [Changes to RNG options](#changes-to-rng-options) |
+| `MBEDTLS_CTR_DRBG_USE_128_BIT_KEY` | TBD | [Changes to RNG options](#changes-to-rng-options) |
+| `MBEDTLS_DEPRECATED_REMOVED` | Keep | Generic feature support |
+| `MBEDTLS_DEPRECATED_WARNING` | Keep | Generic feature support |
+| `MBEDTLS_DHM_C` | Remove | Dead code: [removal of low-level DHM](#removal-of-low-level-dhm) |
+| `MBEDTLS_ECDH_VARIANT_EVEREST_ENABLED` | Keep | Tune built-in crypto |
+| `MBEDTLS_ECP_DP_SECP224K1_ENABLED` | Remove | [Removal of small elliptic curves](#removal-of-small-elliptic-curves) |
+| `MBEDTLS_ECP_NIST_OPTIM` | Keep | Tune built-in crypto |
+| `MBEDTLS_ECP_RESTARTABLE` | Keep | No corresponding PSA option yet |
+| `MBEDTLS_ECP_WITH_MPI_UINT` | Keep | Tune built-in crypto |
```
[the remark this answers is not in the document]
gilles-peskine-arm: We don't have a way to make a configuration option internal without extra tooling. The option is currently documented as experimental, so I think changing its handling would be unrelated to the major version change. Since we're already pressed for time, I hadn't considered taking action here.

docs/architecture/0e-plans.md
gilles-peskine-arm: Yes, that would be a code size optimization and cleanup, which can be done in 1.x.

docs/architecture/0e-plans.md
```diff
@@ -568,6 +568,513 @@ Note that if we want to be able to remove all error code translations (to save c
...
+| `MBEDTLS_CIPHER_NULL_CIPHER` | Remove | No longer used by TLS |
+| `MBEDTLS_CIPHER_PADDING_ONE_AND_ZEROS` | Remove | [Obsolete cipher padding modes](#obsolete-cipher-padding-modes) |
+| `MBEDTLS_CIPHER_PADDING_ZEROS` | Remove | [Obsolete cipher padding modes](#obsolete-cipher-padding-modes) |
+| `MBEDTLS_CIPHER_PADDING_ZEROS_AND_LEN` | Remove | [Obsolete cipher padding modes](#obsolete-cipher-padding-modes) |
+| `MBEDTLS_CTR_DRBG_C` | TBD | [Changes to RNG options](#changes-to-rng-options) |
+| `MBEDTLS_CTR_DRBG_USE_128_BIT_KEY` | TBD | [Changes to RNG options](#changes-to-rng-options) |
+| `MBEDTLS_DEPRECATED_REMOVED` | Keep | Generic feature support |
+| `MBEDTLS_DEPRECATED_WARNING` | Keep | Generic feature support |
+| `MBEDTLS_DHM_C` | Remove | Dead code: [removal of low-level DHM](#removal-of-low-level-dhm) |
+| `MBEDTLS_ECDH_VARIANT_EVEREST_ENABLED` | Keep | Tune built-in crypto |
```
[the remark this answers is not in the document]
gilles-peskine-arm: I rather see this as similar to `MBEDTLS_AESNI_C` and `MBEDTLS_AESCE_C`. Ideally these would be configured through the driver configuration mechanism, but that isn't ready yet and won't be ready in time for 1.0. Hence I propose to keep these compile-time options. In 1.x, we can make them auto-enable certain drivers.

We should update the documentation here as part of privatizing `ecdh.h` — the ECDH context is no longer a public interface. But other than that, I don't see anything in the option that commits us to having an ECDH module in any particular form.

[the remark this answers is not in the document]
mpg: >  I don't see anything in the option that commits us to having an ECDH module in any particular form.

Right, just because the name has ECDH in it does not mean we have to keep `ecdh.c` internally. We can keep the option and later give it a meaning similar to that of `MBEDTLS_PSA_P256M_DRIVER_ENABLED`.

> I rather see this as similar to `MBEDTLS_AESNI_C` and `MBEDTLS_AESCE_C`.

From an implementation perspective, I'm more eager to move Everest to a driver because it will remove a more complexity and useless code - but you're right, from an API perspective (the one that matters now) they're the same.

docs/architecture/0e-plans.md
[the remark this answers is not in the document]
gilles-peskine-arm: There are a few differences between `CFLAGS=-D...`  and a `*config.h` option. Only the latter is tracked in the outcome file. I'm not sure offhand if we'd need to do something about `check_names`. None of this is hard to deal with, but it's not completely trivial and transparent.

docs/architecture/0e-plans.md
```diff
@@ -262,30 +264,44 @@ In an exposed header, the minimum work for 0ε is:
...
-The following table summarizes the characteristics of each category.
+The following table summarizes the characteristics of each category and their location in TF-PSA-Crypto 0ε.
 
-| Category | Location | In include path? |
+| Category | Location | Installed? | Doxygen? | Visible to Mbed TLS? |
 | -------- | -------- | ---------------- |
-| Public   | `include` | yes |
-| Exposed  | `include` | yes except for Doxygen |
-| Private  | `drivers/*/include` | only for crypto |
+| Public   | `include/mbedtls` | yes | yes | yes |
```
mpg, line 282: As noted elsewhere, we might want to make that `include/tf-psa-crypto`.

docs/architecture/0e-plans.md
```diff
@@ -262,30 +264,44 @@ In an exposed header, the minimum work for 0ε is:
...
+| Exposed  | `drivers/*/include/mbedtls/private` | yes | no | yes |
```
mpg, line 283: Note: see "Move private headers" later in this document for the rationale.

docs/architecture/0e-plans.md
```diff
@@ -262,30 +264,44 @@ In an exposed header, the minimum work for 0ε is:
...
+| Private  | `drivers/*/include/mbedtls/private` | yes for now | no | yes |
+| Internal | anything except `**/include/mbedtls/*.h` | no | no | no |
```
mpg, line 285: Just to make this "anything except" more concrete, this will typically be `drivers/builtin/src/*.h` but can also be in other locations for other drivers.

docs/architecture/0e-plans.md
```diff
@@ -424,18 +454,32 @@ Main loss of functionality:
...
 
+We will make these headers internal after 0ε. For 0ε, they can remain private.
+
+#### Move private headers
+
+Move private and exposed headers so that they are always under a subdirectory called `.../private`. This is a simple way of conveying that their content is not part of the public API both to humans and to programs.
+
+Note that public headers can include private headers, since some private headers define exposed types and macros.
+
+ACTION (https://github.com/Mbed-TLS/mbedtls/issues/10087): Move TF-PSA-Crypto private headers from `drivers/builtin/include/mbedtls` to `drivers/builtin/include/mbedtls/private`. Move mbedtls private headers from `include/mbedtls` to `include/mbedtls/private`. This requires coordinated action between the three repositories.
```
mpg, line 468: This is new in this version of the document and IMO not a trivial change to bring in at the last minute. I can see why this is desirable (first paragraph) but the costs are not clear to me, notably the impact on the files that live in the framework - so at this point it's not clear to me if the benefits outweigh the cost, nor whether it's really a MUST for $0\epsilon$ or more of a nice to have.

docs/architecture/0e-plans.md
gilles-peskine-arm: `tf-psa-crypto/*.h` is not an interface I've considered in this document. It exists for new interfaces that were created by the product split. It isn't relevant for any of the interfaces that we're keeping or lightly adapting, and this document does not propose to create any new interface that would involve creating a new header.

docs/architecture/0e-plans.md
gilles-peskine-arm: This isn't new, but it took me a while to come up with a concrete action. Earlier the document didn't make a clear distinction between the header itself as an interface (what you put after `#include`), and the elements that are defined by headers (types, functions, etc).

Moving headers with only private content to a separate directory simplifies the work we have to do to describe which interfaces are public and which ones are private. As far as I remember (seeing this after a 2-week pause), if we don't do it, then I think we have to do more work inside the headers themselves, to ensure that:

* Private functions don't appear in the Doxygen documentation.
* Private functions don't appear to be public functions when application developers read the header files directly.
* Application code that attempts to use private functions won't compile without going through hoops.

gilles-peskine-arm: Hmmm, actually https://github.com/Mbed-TLS/TF-PSA-Crypto/issues/223 should be enough for documentation purposes — if it's not in the top-level `/include` then it's not a public interface. So it's not critical for that part.

I do think `#include <mbedtls/cipher.h>` should be an error in TF-PSA-Crypto 1.x. But yes, I guess technically we can get away with it compiling but not being documented or useful. Let me think about this further though.

docs/architecture/0e-plans.md
mpg: I'm not sure there's general agreement about that. @ronald-cron-arm [commented elsewhere](https://github.com/Mbed-TLS/TF-PSA-Crypto/issues/223#issuecomment-2747270457) that he thinks `include/tf-psa-crypto` would be a  more logical place for things like `asn1.h`. (Sorry, should have added a link in my first comment "noted elsewhere" is not helpful on its own out of context.)

I must admit that before Ronald's comment / this table here I hadn't really thought about what location we wanted for the headers that remain public. And not that the question becomes apparent to me, it's not clear to me that `include/mbedtls` should exist at all on the crypto side.

> `tf-psa-crypto/*.h` [...] exists for new interfaces that were created by the product split. It isn't relevant for any of the interfaces that we're keeping or lightly adapting

Is that a rule that was reviewed already or has it just been on your mind for a long time?

docs/architecture/0e-plans.md
mpg: > This isn't new

Well I think it's new that it's explicit - I agree it can be seen logical consequence of what was there before, but since that wasn't on my mind last time I reviewed the document, as least this is not covered by my previous approvals.

davidhorstmann-arm (commented):
LGTM for what's there, apart from the concern that Manuel raised

docs/architecture/0e-plans.md
mpg: I agree that it's desirable for `#include "mbedtls/cipher.h"` to be an error, but we seem to agree it's not strictly needed though. So, I'd be happy with the following resolution if you're happy with it as well:
- demote 10087 from MUST to SHOULD (and mark it as "implementation needed" again)
- keep 223 as a MUST of course
- no change required in this document.

Wdyt?

mpg (approved):
LGTM. At this point, I'd like us to:
- create DI tasks to reflect the remaining unresolved points (I think that's mostly compile options now)
- merge this PR mostly as it is, 
- then future PRs can update the document as part of the follow-up DI tasks.

I think this would accurately reflect the fact that most of the design is done and approved, and make it more visible which points remain to be decided.

davidhorstmann-arm (dismissed):
Noting my approval with an actual approval

gilles-peskine-arm:
I've now filed issues for all points that were marked as TODO. I hope we can merge this quickly now, and update the document via the remaining DI tasks.

docs/architecture/0e-plans.md
```diff
@@ -612,7 +612,50 @@ Note that if we want to be able to remove all error code translations (to save c
...
+* The option is still present in the code base, but now implied by other options. This is notably the case for legacy crypto options that are now always controlled via `PSA_WANT_xxx` (`MBEDTLS_PSA_CRYPTO_CONFIG` always on). Setting it may cause the configuration to be inconsistent, so we should complain if the option is set through a configuration file (`mbedtls/mbedtls_config.h`, `psa/crypto_config.h` or similar) rather than internally (in `*/config_adjust*.h`).
+* The option is still present in the code base and not set automatically, but it is not part of the public interface, and will be removed in 1.x/4.x. This happens when we remove a feature that tests still rely on (for example, directly removing some small elliptic curves ([#8136](https://github.com/Mbed-TLS/mbedtls/issues/8136)) would remove some test coverage). We should complain if the option is set, unless it is set from our own test scripts.
+
+#### How to remove a legacy configuration option
+
+Goal: complain if a user tries to set an option in the configuration file (`mbedtls/mbedtls_config.h`, `psa/crypto_config.h` or similar), for options that existed in 3.x but have become internal in 1.0/4.0.
```
mpg, line 629: What if a user unsets an option? For example, they use the default config file plus a few `#undef`s in `MBEDTLS_USER_CONFIG_FILE`. I think we want to complain there too.

mpg (approved):
LGTM. (The two comments/questions I have can be handled as part of https://github.com/Mbed-TLS/mbedtls/issues/10147.)

docs/architecture/0e-plans.md
```diff
@@ -612,7 +612,50 @@ Note that if we want to be able to remove all error code translations (to save c
...
+
+Note that users may be setting crypto options in the Mbed TLS configuration, which is included after the config adjustments in TF-PSA-Crypto have set some legacy symbols. We need a complaint if TF-PSA-Crypto's `build_info.h` does not set some legacy option and the user's `mbedtls_config.h` then sets it. How can we do this?
+
+* Remember the state of all implied options before including the Mbed TLS configuration, then check if it has changed. This requires a lot of boilerplate. This doesn't cause a complaint if the Mbed TLS configuration sets an implied option redundantly, only if it sets an implied option that should no longer be set.
+* Set implied options to an expansion that is a nonzero integer (e.g. 0xdeadc0f1) when doing config adjustments. After reading the Mbed TLS configuration, reject implied options that are defined but false (which is the case for `#if FOO` when `FOO` is defined with an empty expansion). This causes a complaint if the Mbed TLS configuration sets an option redundantly, but the reason for the complaint is not clear from the error message (macro redefinition warning).
```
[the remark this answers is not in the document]
gilles-peskine-arm: Looks good. #10147 has flexible enough design that it's easy to try variants like this.

docs/architecture/0e-plans.md
gilles-peskine-arm: Yes, that can be detected too. It's rarer and I'd consider it a nice-to-have.

davidhorstmann-arm (approved):
LGTM, thanks!

gilles-peskine-arm:
The CI is failing because this PR is too old compared to mbedtls. This PR only touches a document, so it doesn't affect the CI apart from `check_files.py`. So I'm going ahead and merging.

PR #159, opened 2025-01-15
Neon impl of ChaCha20 (better size & perf)

daverodgman:
## Description

Neon implementation of ChaCha20 which, depending on how it's configured, is ~500b smaller than the existing impl, or up to 4.6x faster. In its default setting it has slightly better code size than the existing impl, and is 2-2.7x faster.

Tested on all combinations of Armv7 arm/thumb2; Armv8 arm/thumb2/aarch64 on clang 14, gcc 11, and some more recent versions, plus Armv8 aarch64_be on gcc 7.5.

## PR checklist

- [ ] **changelog** provided
- [ ] **development PR** provided
- [ ] **3.6 PR** not required because: non-functional change
- [ ] **2.28 PR** not required because: non-functional change
- [ ] **tests**  to follow via separate PR in mbedtls repo - see https://github.com/Mbed-TLS/mbedtls/pull/9701/commits/a8e9a884bd5582f9d5a9adf22472380b45864d2b

| | multiblock setting | size delta c.f. development (aarch64) | perf delta c.f. development (aarch64) |
| --- | --- | --- | --- |
| clang | 0 | -84 | 0% |
| | 1 | -548 | 14% |
|  | 2 | -344 | 64% |
|  | 3 | -144 | 98% |
|  | 4 | 76 | 172% |
|  | 5 | 352 | 161% |
|  | 6 | 576 | 224% |
|  |  |  |  |
| gcc | 0 | -71 | 0% |
| | 1 | -519 | 57% |
|  | 2 | -275 | 129% |
|  | 3 | 1 | 177% |
|  | 4 | 309 | 280% |
|  | 5 | 665 | 273% |
|  | 6 | 881 | 355% |


ronald-cron-arm:
Thanks for moving your PR to the new repo.

davidhorstmann-arm (changes requested):
This looks good to me generally. Could you add a testcase in `tests/suites/test_suite_chacha20.data` that depends on `MBEDTLS_HAVE_NEON_INTRINSICS`? It can be a duplicate of one of the existing cases.

That way the CI should complain if we don't test this implementation at all, since any testcase that is never executed will generate an error in `analyze_outcomes.py`.

davidhorstmann-arm:
It also looks like this could do with a rebase on top of latest for the CI's benefit.

daverodgman:
Rebased (I just ran `git rebase` which was clean, haven't deeply inspected the result). Added requested test.

If you can advise on how/where to pull in [‎tests/scripts/components-configuration-crypto.sh](https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/159#diff-1ed4ee9db4199382a2d6256c12dd611e60ceb3ba4d91b0c1c6437a0680238186) I'll add that too.

davidhorstmann-arm (commented):
Apologies for the delay, I think I have worked out why the CMake tests are not working (although I don't currently have an appropriate machine set up to test).

Would you be able to try out this change? I should then be able to approve.

davidhorstmann-arm (dismissed):
LGTM, thanks!

davidhorstmann-arm (dismissed):
LGTM, thanks!

davidhorstmann-arm (dismissed):
LGTM, thanks!

CI is on hold for now for release testing but I'll kick it once we are in the clear there.

davidhorstmann-arm (dismissed):
LGTM, thanks!

davidhorstmann-arm (dismissed):
LGTM, thanks!

drivers/builtin/src/chacha20.c
```diff
@@ -158,20 +157,22 @@ int mbedtls_chacha20_setkey(mbedtls_chacha20_context *ctx,
...
     /* Set key */
-    ctx->state[4]  = MBEDTLS_GET_UINT32_LE(key, 0);
-    ctx->state[5]  = MBEDTLS_GET_UINT32_LE(key, 4);
-    ctx->state[6]  = MBEDTLS_GET_UINT32_LE(key, 8);
-    ctx->state[7]  = MBEDTLS_GET_UINT32_LE(key, 12);
-    ctx->state[8]  = MBEDTLS_GET_UINT32_LE(key, 16);
-    ctx->state[9]  = MBEDTLS_GET_UINT32_LE(key, 20);
-    ctx->state[10] = MBEDTLS_GET_UINT32_LE(key, 24);
-    ctx->state[11] = MBEDTLS_GET_UINT32_LE(key, 28);
+    if (MBEDTLS_IS_BIG_ENDIAN) {
```
yanesca, line 176: Is there a reason for doing this check in runtime?

yanesca, line 176: This change is likely to make things slower if compiled with an older GCC on a platform without efficient unaligned access or if compiled with IAR on some platforms. We could prevent this if we checked for `UINT_UNALIGNED` and `UINT_UNALIGNED_STRUCT` as well.

drivers/builtin/src/chacha20.c
```diff
@@ -184,9 +185,13 @@ int mbedtls_chacha20_starts(mbedtls_chacha20_context *ctx,
     ctx->state[12] = counter;
 
     /* Nonce */
-    ctx->state[13] = MBEDTLS_GET_UINT32_LE(nonce, 0);
-    ctx->state[14] = MBEDTLS_GET_UINT32_LE(nonce, 4);
-    ctx->state[15] = MBEDTLS_GET_UINT32_LE(nonce, 8);
+    if (MBEDTLS_IS_BIG_ENDIAN) {
```
yanesca, line 200: Same as above.

drivers/builtin/src/chacha20.c
```diff
@@ -204,7 +209,7 @@ int mbedtls_chacha20_update(mbedtls_chacha20_context *ctx,
     size_t offset = 0U;
 
     /* Use leftover keystream bytes, if available */
-    while (size > 0U && ctx->keystream_bytes_used < CHACHA20_BLOCK_SIZE_BYTES) {
+    while (ctx->keystream_bytes_used < CHACHA20_BLOCK_SIZE_BYTES && size) {
```
yanesca: nit (for readability):
```suggestion
    while (ctx->keystream_bytes_used < CHACHA20_BLOCK_SIZE_BYTES && size > 0) {
```

drivers/builtin/src/chacha20.c
```diff
@@ -158,20 +157,22 @@ int mbedtls_chacha20_setkey(mbedtls_chacha20_context *ctx,
                             const unsigned char key[32])
 {
     /* ChaCha20 constants - the string "expand 32-byte k" */
-    ctx->state[0] = 0x61707865;
-    ctx->state[1] = 0x3320646e;
-    ctx->state[2] = 0x79622d32;
-    ctx->state[3] = 0x6b206574;
+    static const char EXPAND_32_BYTE_K[16] = "expand 32-byte k";
+    memcpy(ctx->state, EXPAND_32_BYTE_K, 16);
```
yanesca: I would think that on big endian machines this changes the values of the constants. What am I missing?

drivers/builtin/src/chacha20.c
```diff
@@ -22,13 +22,250 @@
...
+
+static inline uint32x4_t chacha20_neon_vrotlq_8_u32(uint32x4_t v)
+{
+    uint32x4_t result;
+#if defined(MBEDTLS_ARCH_IS_ARM64)
+    // This implementation is slightly faster, but only supported on 64-bit Arm
+    // Table look-up which results in an 8-bit rotate-left within each 32-bit element
+    const uint8_t    tbl_rotl8[16] = { 3, 0, 1, 2, 7, 4, 5, 6, 11, 8, 9, 10, 15, 12, 13, 14 };
+    const uint8x16_t vrotl8_tbl = vld1q_u8(tbl_rotl8);
+    result = vreinterpretq_u32_u8(vqtbl1q_u8(vreinterpretq_u8_u32(v), vrotl8_tbl));
```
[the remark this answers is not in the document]
yanesca: I think this raises two questions:
1. do the performance improvements persist when the DIT bit is set (I mean is this implementation faster with DIT on than the else branch without DIT set)?
2. we haven't thought about how the library would relate to the DIT feature. Do we recommend users to set it? Are we going to add it to platform abstraction and control it ourselves? Something else? What security requirements we have for DIT code and non-DIT code? What security guarantees can we achieve and how do we communicate them?

 And I am more concerned about the second point, I don't feel we are ready to integrate code for which the security relies on DIT being set.

drivers/builtin/src/chacha20.c
```diff
@@ -22,13 +22,121 @@
...
+static inline uint32x4_t chacha20_neon_vrotlq_12_u32(uint32x4_t v)
+{
+    uint32x4_t x = vshlq_n_u32(v, 12);
+    return vsriq_n_u32(x, v, 20);
+}
+
+static inline uint32x4_t chacha20_neon_vrotlq_8_u32(uint32x4_t v)
+{
+    uint32x4_t result;
+#if defined(MBEDTLS_ARCH_IS_ARM64)
```
yanesca: I am a bit concerned on the [DIT bit](https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/159/commits/bd5c34ccf89b6c2033d9b31d5d1b52b51013a35d#r2303621644) requirement for this to be constant time. Do you think we could go for the `else` branch for now and add this part back when we managed to sort out our thoughts around DIT?

tests/scripts/components-configuration.sh
```diff
@@ -146,3 +146,29 @@ component_tf_psa_crypto_test_memory_buffer_allocator () {
...
+    case $(uname -m) in
+        aarch64) true;;
+        *) false;;
+    esac
+}
+
+component_test_chacha20_neon_variations () {
+    msg "ChaCha20 Neon scalar and multiblock variations"
+
+    scripts/config.py set PSA_WANT_KEY_TYPE_AES
```
yanesca: I am sorry, am not very familiar with the new config, why do we need `PSA_WANT_KEY_TYPE_AES` and `PSA_WANT_ALG_SHA_256` for ChaCha?

tests/scripts/components-configuration.sh
```diff
@@ -146,3 +146,29 @@ component_tf_psa_crypto_test_memory_buffer_allocator () {
     msg "test: MBEDTLS_MEMORY_BUFFER_ALLOC_C"
     make test
 }
+
+support_test_chacha20_variations () {
```
yanesca: ~~I am not sure how this function is used by our test scripts, could you please help me understand where this is called from?~~

Never mind, fixed in later commit.

yanesca (commented):
Hi Dave, thank you for the PR, I had a look and I have a few questions.

drivers/builtin/src/chacha20.c
daverodgman: Because `MBEDTLS_IS_BIG_ENDIAN` is not guaranteed to be compile-time const. On just about every reasonable compiler, it is constant, and the branch is optimised away.

daverodgman: I would expect `memcpy` (the bit I've added) to be faster than repeated use of `MBEDTLS_GET_UINT32_LE`.

In any case, note that this is part of `mbedtls_chacha20_setkey` which is not perf sensitive. It is a decent size saving though.

drivers/builtin/src/chacha20.c
daverodgman: Note that the look-up table that describes how to re-order the data is a compile-time const - only the register it is shuffling is unknown. So I don't think this could result in data-dependent timing.

tests/scripts/components-configuration.sh
daverodgman: This was needed previously, but seems not necessary now.

drivers/builtin/src/chacha20.c
daverodgman: Hmm. In `https://cr.yp.to/streamciphers/timings/estreambench/submissions/salsa20/chacha8/regs/chacha.c` this is written using:

```
static const char sigma[16] = "expand 32-byte k";
constants = sigma;
x->input[0] = U8TO32_LITTLE(constants + 0);
x->input[1] = U8TO32_LITTLE(constants + 4);
```

ie treat the string as a bunch of little-endian 32-bit ints. So it will write "e" into the lowest address, then "x", and so on, regardless of endianness. So unless I've missed something, this is equivalent behaviour to `memcpy`.

It does look like previously, the constant would have been written differently on big-endian, which looks like a bug.

daverodgman: Sorry, no, you were right. I've fixed, and tested on big-endian aarch64. On little-endian (ie the common case) the `memcpy` path saves 20 bytes so worth retaining this.

drivers/builtin/src/chacha20.c
yanesca: Oh, I think I misread the code, so `tbl_rotl8` actually contains the indices and `v` is the table from which we are performing the lookup. In that case we should indeed be safe.

drivers/builtin/src/chacha20.c
```diff
@@ -22,13 +22,253 @@
...
+    // This implementation is slightly faster, but only supported on 64-bit Arm
+    // Table look-up which results in an 8-bit rotate-left within each 32-bit element
+    const uint8_t    tbl_rotl8[16] = { 3, 0, 1, 2, 7, 4, 5, 6, 11, 8, 9, 10, 15, 12, 13, 14 };
```
yanesca: Can we please rename this variable to `idx_rotl8` or something similar, the naming really threw me off when reading the code.

drivers/builtin/src/chacha20.c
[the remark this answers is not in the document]
yanesca: Never mind, I misread the code, it should be safe.

drivers/builtin/src/chacha20.c
```diff
@@ -213,6 +468,37 @@ int mbedtls_chacha20_update(mbedtls_chacha20_context *ctx,
         size--;
     }
 
+#if MBEDTLS_CHACHA20_NEON_MULTIBLOCK != 0
```
yanesca: Do you think we could move the neon code into its own source file? It is easier to read/maintain like that and that is how we have been doing it with AES as well.

(The two implementations seem to be nicely separated, I would hope that this is not a very big change.)

yanesca (changes requested):
Hi Dave, thank you very much for reworking the PR, I had a look at it and I have two small last asks.

davidhorstmann-arm (dismissed):
LGTM, thanks!

Happy with the small improvements, reviewed the refactor with `--color-moved`.

davidhorstmann-arm (approved):
LGTM, thanks!

yanesca (approved):
LGTM, thanks

drivers/builtin/src/chacha20_neon.c
```diff
@@ -0,0 +1,263 @@
...
+{
+    uint32x4_t x = vshlq_n_u32(v, 7);
+    return vsriq_n_u32(x, v, 25);
+}
+
+// Increment the 32-bit element within v that corresponds to the ChaCha20 counter
+static inline uint32x4_t chacha20_neon_inc_counter(uint32x4_t v)
+{
+    if (MBEDTLS_IS_BIG_ENDIAN) {
+        uint32x4_t counter_increment = { 0, 0, 0, 1 };
```
gilles-peskine-arm, line 69: [This initializer is not portable](https://github.com/Mbed-TLS/TF-PSA-Crypto/issues/551). It works with GCC, Clang and IAR but not MSVC. @daverodgman would you mind reviewing my proposed fix? https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/552

PR #164, opened 2025-01-20
Move configs to tf psa crypto

Harry-Ramsey:
Move TF-PSA-Crypto configs to TF-PSA-Crypto.

## PR checklist

- [ ] **mbedtls PR** provided Mbed-TLS/mbedtls: [#9908](https://github.com/Mbed-TLS/mbedtls/pull/9908)

## Notes for the submitter

Help make review efficient:
* Multiple simple commits
* Avoid force-push


ronald-cron-arm (changes requested):
For the repo split we've tried to keep the whole history of the files moved from Mbed TLS, no rewrite of the history with only the files that are moved. Thus for the three
configuration files crypto-config-ccm-aes-sha256.h, crypto-config-symmetric-only.h and crypto_config_profile_medium.h it would be better to continue on that path. https://github.com/ronald-cron-arm/TF-PSA-Crypto/tree/restore-some-crypto-configs does that but for one commit modifying crypto_config_profile_medium.h. Thus I would prefer to go that way for those files.

Harry-Ramsey:
> For the repo split we've tried to keep the whole history of the files moved from Mbed TLS, no rewrite of the history with only the files that are moved. Thus for the three configuration files crypto-config-ccm-aes-sha256.h, crypto-config-symmetric-only.h and crypto_config_profile_medium.h it would be better to continue on that path. https://github.com/ronald-cron-arm/TF-PSA-Crypto/tree/restore-some-crypto-configs does that but for one commit modifying crypto_config_profile_medium.h. Thus I would prefer to go that way for those files.

After playing around with the method above, what is the best way to keep changes after the reposplit? We can retrieve commits inbetween using `git log` and `git am` but it seems you used a different method?

ronald-cron-arm (changes requested):
Looks good to me but we need to remove the references to the TLS ciphersuites in  crypto_config_profile_medium.h as we have done in crypto_config.h.

ronald-cron-arm:
> > For the repo split we've tried to keep the whole history of the files moved from Mbed TLS, no rewrite of the history with only the files that are moved. Thus for the three configuration files crypto-config-ccm-aes-sha256.h, crypto-config-symmetric-only.h and crypto_config_profile_medium.h it would be better to continue on that path. https://github.com/ronald-cron-arm/TF-PSA-Crypto/tree/restore-some-crypto-configs does that but for one commit modifying crypto_config_profile_medium.h. Thus I would prefer to go that way for those files.
> 
> After playing around with the method above, what is the best way to keep changes after the reposplit? We can retrieve commits inbetween using `git log` and `git am` but it seems you used a different method?

I am fine with the way you have done it, thanks. I've just asked for a reference to the original Mbed TLS commit.

ronald-cron-arm (approved):
LGTM, thanks.

valeriosetti (approved):

PR #165, opened 2025-01-23
[TF-PSA-Crypto] Remove DHE-PSK key exchange

valeriosetti:
## Description

This is part of https://github.com/Mbed-TLS/mbedtls/pull/9910

The commit is cherry-picked from #143 and I just removed 2 extra spaces from there.

## PR checklist

- [ ] **changelog** provided | not required because: 
- [ ] **framework PR** provided Mbed-TLS/mbedtls-framework# | not required
- [ ] **mbedtls PR** provided Mbed-TLS/mbedtls#9910
- **tests**  provided | not required because:

PR #167, opened 2025-01-24
Do not add error codes

gilles-peskine-arm:
Never add a low-level error code to a high-level error code. Just use the low-level error code. Resolves https://github.com/Mbed-TLS/mbedtls/issues/9619, a step towards [having unified PSA and mbedtls error codes](https://github.com/Mbed-TLS/mbedtls/issues/8501).

In this pull request, we keep calls to `MBEDTLS_ERROR_ADD`. The idea is to keep them around for a while, but not to use addition in new code. Thus

* High-level error codes may still appear as false positives in searches even where they're not used, in calls of the form `MBEDTLS_ERROR_ADD(MBEDTLS_ERR_HIGH, xxx)` where `xxx` is never 0.
* We still have the opportunity to change some places where we want the high-level error to win.
* There are few lines of code to change, and unchanged code can still have bug fixes that will be easy to backport in 3.6.

Follow-up to https://github.com/Mbed-TLS/mbedtls/pull/9566, which added enforcement that we do use `MBEDTLS_ERROR_ADD` when constructing high+low error codes, and not just directly the plus operator.

## PR checklist

- [x] **changelog** provided
- [x] **mbedtls PR** provided https://github.com/Mbed-TLS/mbedtls/pull/9926
- **tests**  provided


ChangeLog.d/error-unification.txt
```diff
@@ -0,0 +1,6 @@
+API changes
+   * The PSA and Mbed TLS error space are now unified. This means that
+     psa_xxx() functions can now potentially return MBEDTLS_ERR_xxx
```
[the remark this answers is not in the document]
gilles-peskine-arm: PSA functions returning legacy errors is not happening right now. I'm anticipating that we'll get sloppy with conversion, in fact ideally I'd like to remove all instances of conversion, after which PSA crypto functions may end up returning proprietary status values that are named `MBEDTLS_ERR_xxx`. Ideally that would only happen for `MBEDTLS_ERR_xxx` constants that are aliases of a `PSA_ERROR_xxx` values though.

I wasn't sure about this. Maybe we should commit to being stricter? In any case we can revise this entry later.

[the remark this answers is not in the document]
gilles-peskine-arm: Updated to only allow `mbedtls_xxx()` to return PSA errors.

drivers/builtin/include/mbedtls/error_common.h
```diff
@@ -11,97 +11,22 @@
...
 #ifdef __cplusplus
 extern "C" {
 #endif
 
-/** Generic error */
-#define MBEDTLS_ERR_ERROR_GENERIC_ERROR       -0x0001
-/** This is a bug in the library */
-#define MBEDTLS_ERR_ERROR_CORRUPTION_DETECTED -0x006E
+/* Generic error */
+#define MBEDTLS_ERR_ERROR_GENERIC_ERROR       PSA_ERROR_GENERIC_ERROR
```
gilles-peskine-arm, line 22: Should we have a changelog entry to call out:

* when a specific `MBEDTLS_ERR_xxx` constant becomes private?
* when a specific `MBEDTLS_ERR_xxx` constant remains public, but becomes an alias of another value?

minosgalanakis: I think we shoud in both cases. This is a user facing error, and people will look at the changelog first when they see something not matching what they expect.

mpg: Agreed.

On the practical side, this may make the ChangeLog pretty verbose, so perhaps in the end we'll just want to have a generic entry like "Some MBEDTLS_ERR_ constants have become private, and others have become aliases to other values. For details, see <relevant section> in the migration guide." But in the meantime it's probably better to create a ChangeLog entry for every change we make, so we have a comprehensive list (while avoiding conflicts) and we can consolidate/edit in the late stages before the release.

mpg (dismissed):
LGTM, thanks!

drivers/builtin/include/mbedtls/error_common.h
gilles-peskine-arm: I propose to start the migration document right away. Maybe multiple files that we'll merge at the end, similar to changelog.d, to avoid conflicts during the work?

gilles-peskine-arm: I've gone ahead and created a directory for migration document fragments, and written a fragment about error codes.

minosgalanakis (approved):
Looks good, the CI failure seems unrelated to this change?

gilles-peskine-arm:
The CI failures in `check_names` are because this branch is older than https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/165 but it got tested with a commit of Mbed TLS that has the corresponding mbedtls pull request merged. I'll rebase to get a clean CI.

mpg (approved):
LGTM, thanks!

PR #135, opened 2024-12-30
Add components-sanitizers.sh

Harry-Ramsey:
This commit adds multiple sanitizer components to TF-PSA-Crypto similar to those found in Mbed TLS. Closes #130 

## PR checklist

Please remove the segment/s on either side of the | symbol as appropriate, and add any relevant link/s to the end of the line.
If the provided content is part of the present PR remove the # symbol.

- [x] **changelog** not required because: testing enhancement.
- [x] **framework PR** not required.
- [x] **mbedtls PR** not required.
- **tests**  provided.

## Notes for the submitter

Please refer to the [contributing guidelines](https://github.com/Mbed-TLS/TF-PSA-Crypto/blob/development/CONTRIBUTING.md), especially the
checklist for PR contributors.

Help make review efficient:
* Multiple simple commits
  - please structure your PR into a series of small commits, each of which does one thing
* Avoid force-push
  - please do not force-push to update your PR - just add new commit(s)
* See our [Guidelines for Contributors](https://mbed-tls.readthedocs.io/en/latest/reviews/review-for-contributors/) for more details about the review process.

PR #170, opened 2025-01-29
Fix false positives in constant time tests using MSan with Clang 16

gilles-peskine-arm:
Fix false positives of constant-flow testing when using `MBEDTLS_TEST_CONSTANT_FLOW_MEMSAN` on Clang 16 and above. Fixes #9921.

This isn't needed for the CI (in the short to medium term) because still we do this testing on an old platform with an old Clang, but is needed on typical developer machines nowadays.

## PR checklist

- [x] **development PR** https://github.com/Mbed-TLS/mbedtls/pull/9942
- [x] **crypto PR** here
- [x] **3.6 PR** https://github.com/Mbed-TLS/mbedtls/pull/9922
- [ ] **2.28 PR** TODO if 2.28 is still around when we merge this
- **tests**  provided for non-regression; requires local testing with Clang ≥16 to validate that #9921 is fixed.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -175,7 +175,7 @@ find_package(Threads)
 # If this is the root project add longer list of available CMAKE_BUILD_TYPE values
 if(NOT TF_PSA_CRYPTO_AS_SUBPROJECT)
     set(CMAKE_BUILD_TYPE ${CMAKE_BUILD_TYPE}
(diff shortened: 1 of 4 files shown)


waleed-elmelegy-arm (dismissed):
LGTM

gilles-peskine-arm:
Now that https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/135 is merged, this is incomplete: there is now a test component to update in crypto as well as mbedtls. So I've rebased on top of `development` and added a commit to update the component (similar to the one in mbedtls).

Issue #168, opened 2025-01-24
Remove RNG parameters from LMS


gilles-peskine-arm:
In TF-PSA-Crypto 1.0, all RNG calls will go to the PSA RNG. So public functions must no longer take an RNG callback (it would not be honored).

Legacy RNG callbacks have the following form:
```
int (*f_rng)(void *, unsigned char *, size_t), void *p_rng
```

The goal of this task is to update the functions in `include/mbedtls/lms.h` and `src/{lms,lmots}.*` that take an RNG argument:

* Remove the `f_rng` and `p_rng` arguments from the function prototypes.
* Change calls to `f_rng` to instead call `psa_generate_random`.
* Change calling code accordingly.
* Optional, may be done in a follow-up: where the calling code is setting up an entropy context and a DRBG context, and those are no longer needed, remove those contexts.

gilles-peskine-arm:
**Task on hold during rethink**

A possible wrinkle: in the long term, we'd like to support explicit PSA contexts in non-PSA APIs, as requested in https://github.com/ARM-software/psa-api/issues/77. Under that design, functions that take an RNG argument and no key argument should take a PSA context argument, and should make RNG calls through that context. Functions that have a key argument would use that key to determine the PSA context.

In many places, we already track a “context” in the form of RNG arguments that are passed down from API functions to legacy crypto calls. Rather than completely remove those RNG arguments, it could make sense to rewrite `f_rng, p_rng` to `psa_context`, and pass the `psa_context` down. When we reach a PSA API, we'd still only support the default context for now, in which the RNG is `psa_get_random`.

By the time of the 1.0 release, it's unlikely that the PSA crypto context API extension will be finalized, and in any case we do not intend to implement that context API until well after 1.0. So in 1.0, we would have a context parameter where the only supported value is NULL, and other values would be silently ignored. That doesn't seem very nice. But on the other hand, if we don't do it, then later we'll have to reintroduce a context parameter that's passed down, and more annoyingly, we'd have to add a `with_psa_context` variant of all the functions involved with an extra parameter. Is it worth having this temporarily in TF-PSA-Crypto 1.x and Mbed TLS 4.x, in order to avoid having to extend mbedtls APIs when we implement the context feature?

gilles-peskine-arm:
[...]

PR #182, opened 2025-02-25
Remove RNG parameters from LMS

bjwtaylor:
## Description

Remove RNG parameters from LMS resolves https://github.com/Mbed-TLS/TF-PSA-Crypto/issues/168

## PR checklist 

- [x] **changelog** not required as we have agreed to add it it as another task once all rng removal has been completed.
- [x] **mbedtls PR** not required because: API not used in mbedtls
- [x] **mbedtls 3.6 PR** not required because: API change
- [x] **mbedtls 2.28 PR** not required because: API change
- **tests**  not required because: no changes to test

### DIFF
--- a/ChangeLog.d/168.txt
+++ b/ChangeLog.d/168.txt
@@ -1,3 +1,0 @@
-API changes
-   * Remove the rng parameters from lms.* and lmots.*
-
--- a/ChangeLog.d/168.txt
+++ b/ChangeLog.d/168.txt
@@ -1,0 +1,3 @@
+API changes
+   * Remove the rng parameters from lms.* and lmots.*
+
--- a/drivers/builtin/include/mbedtls/lms.h
+++ b/drivers/builtin/include/mbedtls/lms.h
@@ -352,8 +352,6 @@ void mbedtls_lms_private_free(mbedtls_lms_private_t *ctx);
  *                           into.
  * \param type               The LMS parameter set identifier.
  * \param otstype            The LMOTS parameter set identifier.
- * \param f_rng              The RNG function to be used to generate the key ID.
- * \param p_rng              The RNG context to be passed to f_rng
  * \param seed               The seed used to deterministically generate the
  *                           key.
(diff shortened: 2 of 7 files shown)


ChangeLog.d/168.txt
```diff
@@ -0,0 +1,3 @@
+API changes
+   * Remove the rng parameters from lms.* and lmots.*
```
gilles-peskine-arm, line 2: The changelog is about the API, not about the internals, so only `lms.h` is relevant (LMOTS isn't public).

ChangeLog.d/168.txt
```diff
@@ -0,0 +1,3 @@
+API changes
```
gilles-peskine-arm, line 1: I think it would make sense to have a single entry for all the RNG parameter removals, rather than separate entries for each module. We might as well write it now, something like 

> All API functions now use the PSA random generator (psa_get_random()) internally. As a consequence, functions no longer take RNG parameters.

Also, we should give the file a more meaningful name. (The file name doesn't matter in the end, but in the meantime it's easier to find a pending changelog entry in `remove-rng-parameters-from-lms.txt` or `remove-rng-parameters.txt` than in `168.txt`.)

gilles-peskine-arm (changes requested):
LGTM except for the changelog entry

ChangeLog.d/168.txt
bjwtaylor: As agreed I've removed the ChangeLog, as the plan is to produce a consolidated version once all RNG removal is complete.

gilles-peskine-arm (approved):
LGTM

ChangeLog.d/168.txt
bjwtaylor: As agreed I've removed the ChangeLog, as the plan is to produce a consolidated version once all RNG removal is complete.

mpg (approved):
LGTM, thanks!

PR #190, opened 2025-03-07
core: Update common.h to use GCC _Static_assert

Vge0rge:
GCC supports _Static_assert from version 4.6 as a builtin feature even when you compile with std=99.

Use this _Static_assert for the implementation of MBEDTLS_STATIC_ASSERT if there is no defined static_assert.

his is mostly taken from the picolibc methodology just without using the GNUC_PREREQ macro because to improve prortability.

The mentioned picolibc solution:
[github.com/picolibc/picolibc/blob/db4d0fe5952d5ecd714781e3212d4086d970735a/ newlib/libc/include/sys/cdefs.h
](https://github.com/picolibc/picolibc/blob/db4d0fe5952d5ecd714781e3212d4086d970735a/newlib/libc/include/sys/cdefs.h#L387)
## Description

Please write a few sentences describing the overall goals of the pull request's commits.

## PR checklist

- [ ] **mbedtls 2.28 PR** provided Mbed-TLS/mbedtls# | not required because: 

## Notes for the submitter

Help make review efficient:
* Multiple simple commits
* Avoid force-push

### DIFF
--- a/core/common.h
+++ b/core/common.h
@@ -359,6 +359,14 @@ static inline void mbedtls_xor_no_simd(unsigned char *r,
  */
 #if defined(static_assert) && !defined(__FreeBSD__)
 #define MBEDTLS_STATIC_ASSERT(expr, msg)    static_assert(expr, msg)
+/* The GCC compiler supports _Static_assert since version 4.6 even with C99 so checking the
+ * compiler version should suffice.
+ * For non GCC compilers we check that C>=11 is used since C11 introduced _Static_assert.
+ */
+#elif !defined(__cplusplus) && \
+    ((defined(__GNUC__) && ((__GNUC__ == 4 && __GNUC_MINOR >= 6) || __GNUC__ > 4)) || \
+    (defined(__STDC_VERSION__) && __STDC_VERSION__ >= 201112L))
+#define MBEDTLS_STATIC_ASSERT(expr, msg)    _Static_assert(expr, msg)
 #else
 /* Make sure `MBEDTLS_STATIC_ASSERT(expr, msg);` is valid both inside and
  * outside a function. We choose a struct declaration, which can be repeated


core/common.h
```diff
@@ -359,6 +359,8 @@ static inline void mbedtls_xor_no_simd(unsigned char *r,
  */
 #if defined(static_assert) && !defined(__FreeBSD__)
 #define MBEDTLS_STATIC_ASSERT(expr, msg)    static_assert(expr, msg)
+#elif defined(__GNUC__) && ((__GNU__ == 4 && __GNUC_MINOR >= 6) || __GNU__ > 4) && !defined(__cplusplus)
```
ronald-cron-arm, line 362: shouldn't it rather be
```suggestion
#elif defined(__GNUC__) && ((__GNUC__ == 4 && __GNUC_MINOR__ >= 6) || __GNUC__ > 4) && !defined(__cplusplus)
```

Vge0rge: You are right, it seems that something went wrong when I was copying the macro. I already realized it and updated the PR as well. Please have a look again :)

core/common.h
```diff
@@ -359,6 +359,8 @@ static inline void mbedtls_xor_no_simd(unsigned char *r,
  */
 #if defined(static_assert) && !defined(__FreeBSD__)
 #define MBEDTLS_STATIC_ASSERT(expr, msg)    static_assert(expr, msg)
+#elif defined(__GNUC__) && ((__GNUC__ == 4 && __GNUC_MINOR >= 6) || __GNUC__ > 4) && !defined(__cplusplus)
+#define MBEDTLS_STATIC_ASSERT(expr, msg)    _Static_assert(expr, msg)
```
ronald-cron-arm, line 363: Sorry for the delay. After a few searches, it seems the following would support more compilers: 
```
#elif defined(__STDC_VERSION__) && __STDC_VERSION__ >= 201112L && !defined(__cplusplus)
```

Vge0rge: This doesn't work for me, the issue is that gcc supports the static assertion even when you build with std=99 (from version >=4.6) Your version requires a newer std which is not always used in our code. Are you OK to add this statement in addition to the GNUC one? I think that this is the best case because then we cover both the gnu case and the std=11 case. I am happy to add that.

gilles-peskine-arm: `defined(static_assert)` just above should already cover C11.

ronald-cron-arm: > This doesn't work for me, the issue is that gcc supports the static assertion even when you build with std=99 (from version >=4.6) Your version requires a newer std which is not always used in our code. Are you OK to add this statement in addition to the GNUC one? I think that this is the best case because then we cover both the gnu case and the std=11 case. I am happy to add that.

ah okay that makes sense. Yes I would say add both then with maybe your explanation here as a comment to explain why we need 
```#elif defined(__GNUC__) && ((__GNUC__ == 4 && __GNUC_MINOR >= 6) || __GNUC__ > 4) && !defined(__cplusplus)``` in addition to ```#elif defined(__STDC_VERSION__) && __STDC_VERSION__ >= 201112L && !defined(__cplusplus)```.

Comment (Vge0rge) [reply core/common.h]: Ok cool, I checked for both and I added a comment explaining the check.

File: core/common.h [hunk 3]
```diff
@@ -359,6 +359,14 @@ static inline void mbedtls_xor_no_simd(unsigned char *r,
...
 #if defined(static_assert) && !defined(__FreeBSD__)
 #define MBEDTLS_STATIC_ASSERT(expr, msg)    static_assert(expr, msg)
+/* The GCC compiler supports _Static_assert since version 4.6 even with C99 so checking the
+ * compiler version should suffice.
+ * For non GCC compilers we check that C>=11 is used since C11 introduced _Static_assert.
+ */
+#elif !defined(__cplusplus) && \
+      (defined(__GNUC__) && ((__GNUC__ == 4 && __GNUC_MINOR >= 6) || __GNUC__ > 4)) || \
+      (defined(__STDC_VERSION__) && __STDC_VERSION__ >= 201112L)
+#define MBEDTLS_STATIC_ASSERT(expr, msg)    _Static_assert(expr, msg)
```
ronald-cron-arm, line 369:
```suggestion
#elif !defined(__cplusplus) && \
      ((defined(__GNUC__) && ((__GNUC__ == 4 && __GNUC_MINOR >= 6) || __GNUC__ > 4)) || \
       (defined(__STDC_VERSION__) && __STDC_VERSION__ >= 201112L))
#define MBEDTLS_STATIC_ASSERT(expr, msg)    _Static_assert(expr, msg)
```

ronald-cron-arm (changes requested):

ronald-cron-arm (dismissed):
LGTM, thanks.

davidhorstmann-arm (approved):
LGTM, thanks!

ronald-cron-arm (approved):
LGTM, thanks.

gilles-peskine-arm:
The [merge queue](https://mbedtls.trustedfirmware.org/job/mbed-tls-tf-psa-crypto-multibranch/job/gh-readonly-queue%252Fdevelopment%252Fpr-190-a8d53a70c3af27c24da84d882999d535ef02e8b2/1/display/redirect) failed due to a merge in mbedtls midway, this is a [known race condition](https://github.com/Mbed-TLS/mbedtls-test/issues/197). Requeuing.

PR #192, opened 2025-03-07
Remove Everest VS2010 compatibility headers

bensze01:
## Description

These headers were necessary for compatibility with Visual Studio 2010, and interfere with the system headers on Visual Studio 2013+, eg. when building Mbed TLS using the .sln file shipped with the project.

Move the still-required definition of "inline" to callconv.h, where the definition for GCC also lives.

## PR checklist
- [x] **changelog** provided
- [x] **mbedtls development PR** provided Mbed-TLS/mbedtls#10020
- [x] **mbedtls 3.6 PR** Mbed-TLS/mbedtls#10043
- [x] **mbedtls 2.28 PR** Mbed-TLS/mbedtls#10044 
- [x] **tests**  not required: the existing tests cover building Everest on Windows


minosgalanakis (approved):
LGTM

ChangeLog.d/remove-everest-vs2010-compat-headers.txt
```diff
@@ -0,0 +1,4 @@
+Bugfix
+   * Remove Everest Visual Studio 2010 compatibility headers, which could
+     interfere with standard CRT headers in certain situations, eg. when
+     building Mbed TLS with the .sln file shipped with the project.
```
gilles-peskine-arm, line 4: I think we're planning not so ship sln files anymore?

gilles-peskine-arm (changes requested):
LGTM except the changelog entry

ChangeLog.d/remove-everest-vs2010-compat-headers.txt
bensze01: Hmm. Should I just remove everything after the comma?
```
Bugfix
   * Remove Everest Visual Studio 2010 compatibility headers, which could
     interfere with standard CRT headers in certain situations.
```

gilles-peskine-arm: Yes, that seems fine. It's not ideal to just say “certain situations”, because it's hard for users to figure out whether the changelog entry applies to them. But if the situations are “exotic” and hard to explain, but easy to detect (my application crashes when I activate SSL debugging), that's ok.

Hmmm, now that I write this, maybe explain “interfere” a bit more? So that users can know if they're affected, or perhaps more importantly if they aren't affected.

bensze01: Ok, how about:
Bugfix
   * Remove Everest Visual Studio 2010 compatibility headers, which could
     shadow standard CRT headers inttypes.h and stdbool.h with incomplete
     implementatios if placed on the include path.

gilles-peskine-arm: I don't know exactly what broke and led you to make this change. OTOH, suppose someone makes a fork of mbedtls, starting before https://github.com/Mbed-TLS/mbedtls/pull/10020, and consumes TF-PSA-Crypto. They would presumably need this change if they adopt 10020 in their fork. So presumably this is user-visible in some fashion, from the perspective of TF-PSA-Crypto.

Anyway, it's probably better to announce it if in doubt. And we've probably already spent more time than it's worth discussing the details here (sorry!).

gilles-peskine-arm (approved):
LGTM

minosgalanakis (approved):
LGTM

PR #199, opened 2025-03-12
Add an internal function that returns the AES implementation

felixc-arm:
Fixes https://github.com/Mbed-TLS/mbedtls/issues/10012

Adds a function in `aes.c` that returns the AES implementation, namely one of `unknown`, `software`, `AESCE`, `AESNI assembly`, or `AESNI intrinsics`.
This function is then used in a new program `which_aes` to then print out the implementation.

Tested manually and gives the correct result(s) on x86 & aarch64 Linux.

## PR checklist

- [x] **changelog** provided (although not sure if needed): 
- [x] **mbedtls development PR** not required because: functionality contained in TF-PSA-Crypto
- [x] **mbedtls 3.6 PR** not required because: part of 4.0 work https://github.com/Mbed-TLS/mbedtls/issues/9904
- [x] **mbedtls 2.28 PR** not required because: part of 4.0 work https://github.com/Mbed-TLS/mbedtls/issues/9904
- **tests**  not required because: internal feature to be used in further testing (not too sure on this)


drivers/builtin/src/aes.c
```diff
@@ -480,6 +480,36 @@ void mbedtls_aes_free(mbedtls_aes_context *ctx)
     mbedtls_platform_zeroize(ctx, sizeof(mbedtls_aes_context));
 }
 
+void mbedtls_aes_get_implementation(mbedtls_aes_implementation **aes_imp_out)
```
davidhorstmann-arm, line 483: Since this is `void` anyway, it would be neater to have this return the `mbedtls_aes_implementation` rather than having an out parameter. This would also enable you to replace the `goto`s with `return`s.

drivers/builtin/src/aes.c
```diff
@@ -480,6 +480,36 @@ void mbedtls_aes_free(mbedtls_aes_context *ctx)
...
+    }
+#endif
+
+#if !defined(MBEDTLS_AES_USE_HARDWARE_ONLY)
+    aes_imp = MBEDTLS_AES_IMP_SOFTWARE;
+    goto exit;
+#endif
+
+exit:
+    *aes_imp_out = &aes_imp;
```
davidhorstmann-arm, line 510: This sets the out parameter to point to `aes_imp`, which is a stack-allocated variable that will disappear when the function returns, so this will cause undefined behaviour when the caller tries to access the out parameter.

This would also be fixed by returning the value `aes_imp` directly.

davidhorstmann-arm (changes requested):
A couple of issues with the function API but the general approach looks fine to me!

davidhorstmann-arm (approved):
LGTM, thanks!

ronald-cron-arm (approved):
LGTM, thanks. However, before to merge it I would prefer to have the mbedtls PR modifying all.sh components to use `which_aes` instead of the `selftest` program. That way we will exercise `which_aes` and be more confident that it behaves as expected and provides everything we need as a replacement to `selftest`.

ronald-cron-arm:
I've set the `needs-preceding-pr` label which is not completely fit for purpose but should do the job.

mpg:
I'm not convinced `programs/psa` is a great location for the new program, considering it relies on a non-PSA API. I think `programs/test` would be more suitable.

ronald-cron-arm:
> I'm not convinced `programs/psa` is a great location for the new program, considering it relies on a non-PSA API. I think `programs/test` would be more suitable.

Yes good point, I missed that, thanks. @felixc-arm please move `which_aes` to `programs/test`.

ronald-cron-arm:
Instead of merging development HEAD, please rather rebase on top of it. That's our preferred way but in case the rebase is very painful.

felixc-arm:
Will do in future, unfortunately I misremembered how to do it and so did this merge instead... In any case the rebase was to that this works with newer mbedtls.

felixc-arm:
> LGTM, thanks. However, before to merge it I would prefer to have the mbedtls PR modifying all.sh components to use `which_aes` instead of the `selftest` program. That way we will exercise `which_aes` and be more confident that it behaves as expected and provides everything we need as a replacement to `selftest`.

PR to do this is now green: https://github.com/Mbed-TLS/mbedtls/pull/10114

ronald-cron-arm (changes requested):
Please find my comment below. Otherwise you need to rebase as `programs/test/CMakeLists.txt` has been re-organized meanwhile.

programs/test/CMakeLists.txt
```diff
@@ -14,6 +14,10 @@ tfpsacrypto_build_program_common(benchmark)
 target_sources(benchmark PRIVATE $<TARGET_OBJECTS:tf_psa_crypto_test>)
 target_link_libraries(benchmark PRIVATE ${tfpsacrypto_target} ${CMAKE_THREAD_LIBS_INIT})
 
+# benchmark
```
ronald-cron-arm, line 17:
```suggestion
# which_aes
```

ronald-cron-arm (changes requested):
This looks almost good to me but a typo and the fact that unfortunately the commit "Make aes_self_test use the new get_aes_implementation function" breaks the AES tests that search into the selftest executable. To get a green CI for this PR, I propose to do that change in a follow-up PR when this one and #10114 are merged.

ronald-cron-arm (approved):
LGTM, thanks.

ronald-cron-arm:
@davidhorstmann-arm please have a look to the last version, thanks.

davidhorstmann-arm (approved):
LGTM, thanks!

gilles-peskine-arm:
The merge queue failure is due to a [known race condition](https://github.com/Mbed-TLS/mbedtls-test/issues/197). Requeuing.

PR #204, opened 2025-03-14
Make PK partially private

gilles-peskine-arm:
Implement part of https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/203. This started as a prototype branch, and I kept stuff that is on the critical path and that's in a good enough shape to be reviewed and merged.

* New type `mbedtls_pk_spki_alg_t` — DONE
* New type `mbedtls_pk_sigalg_t` — DONE
* Make some pk functions private — mostly done but needs another pass on the documentation to finish removing references to the removed API elements.
* Always enable `MBEDTLS_PK_USE_PSA_EC_DATA` (cheaper than not doing it). Resolves https://github.com/Mbed-TLS/mbedtls/issues/9676.
* Store the algorithm for a PK context in the context, and honor it from operations. — partly done, but needs work on the unit tests
* Rename `mbedtls_pk_wrap_opaque()` to `mbedtls_pk_wrap_psa()`. — DONE

## PR checklist

- [x] **changelog** provided | not required because: task filed for later TODO
- [x] **mbedtls development PR** not required because: new stuff
- [x] **mbedtls 3.6 PR** 
- [ ] **mbedtls 2.28 PR** not required because: new stuff
- **tests**  provided plus TODO

### DIFF
--- a/drivers/builtin/include/mbedtls/pk.h
+++ b/drivers/builtin/include/mbedtls/pk.h
@@ -70,6 +70,9 @@ extern "C" {
 /**
  * \brief          Public key types
  */
+/* Note that the values must match the corresponding enum constants in
+ * mbedtls_pk_spki_alg_t and mbedtls_pk_sigalg_t until we stop using
+ * mbedtls_pk_type_t in mbedtls_pk_context and mbedtls_pk_can_do(). */
 typedef enum {
     MBEDTLS_PK_NONE=0,
     MBEDTLS_PK_RSA,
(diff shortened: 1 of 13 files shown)


drivers/builtin/include/mbedtls/pk.h
```diff
@@ -129,26 +129,8 @@ typedef enum {
...
- *
- * The main reason for the "intermediate" (USE_PSA + ECP_C) above is that as long
- * as ECP_C is defined mbedtls_pk_ec() gives the user a read/write access to the
- * ecp_keypair structure inside the pk_context so they can modify it using
- * ECP functions which are not under PK module's control.
- */
-#if defined(MBEDTLS_USE_PSA_CRYPTO) && defined(PSA_WANT_KEY_TYPE_ECC_PUBLIC_KEY) && \
-    !defined(MBEDTLS_ECP_C)
+/* Legacy auxiliary symbol, now always enabled. */
+#if defined(PSA_WANT_KEY_TYPE_ECC_PUBLIC_KEY)
```
mpg, line 134: Fully agree with doing this, but I expect this to break `mbedtls_pk_ec()` - unless I'm mistaken it will now return a pointer to an unpopulated `ecp_keypair`. This function is still declared as long as `ECP_C` is enabled.

Previously that function was either not defined, or defined and working. Now we have a set of configurations (all those that have `ECP_C` enabled where the function is defined but not working.

I suggest fully removing `mbedtls_pk_ec()` even from the private header. With a bit of luck, we had already removed all internal uses of this function and we'll be OK. Otherwise, we'll know which places need to be fixed.

(Also, I think sooner rather than later we should remove the dead code that was guarded by `!USE_PSA_EC_DATA`. The PK module is quite complex and having dead code all over the place is not helping anyone make sense of it. Happy to consider this out of scope here and defer to https://github.com/Mbed-TLS/mbedtls/issues/9676 of course.)

drivers/builtin/include/mbedtls/pk.h
```diff
@@ -485,6 +341,38 @@ int mbedtls_pk_can_do_ext(const mbedtls_pk_context *ctx, psa_algorithm_t alg,
...
+ *                      algorithm being used on a key with an incompatible
+ *                      type.
+ *
+ * \note                The PK module does not enforce key policies.
+ *                      For example, after parsing an ECC key that is
+ *                      designated as ECDH-only through the OID
+ *                      id-ecDH, you can call this function to allow the
+ *                      key to be used for ECDSA (but you probably should not).
+ */
+void mbedtls_pk_set_algorithm(mbedtls_pk_context *ctx,
```
gilles-peskine-arm, line 373: This won't work with a future workflow where parsing already creates a PSA key.

I was thinking in the future this might export the key and import it with another policy. But this can fail due to running out of resources. So the function needs to return an error code.

gilles-peskine-arm: On second thoughts, maybe `mbedtls_pk_set_algorithm` shouldn't exist at all. I need to think this further,

gilles-peskine-arm: [Having slept on it again, I don't think `mbedtls_pk_set_algorithm` should exist](https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/203/files#r2009291751).

Action: either here or in a follow-up, implement the copy-based workflow.

gilles-peskine-arm:
This prototype has been abandoned. It no longer corresponds to the current design in https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/203. However, parts of it may still be relevant for inspiration or even cherry-picking.

PR #212, opened 2025-03-18
[tf-psa-crypto] MBEDTLS_PLATFORM_GET_ENTROPY_ALT in 4.0

valeriosetti:
## Description

This helps resolving https://github.com/Mbed-TLS/mbedtls/issues/9618

## CI note

Due to dependencies with the [main repo PR](https://github.com/Mbed-TLS/mbedtls/pull/10090) please evaluate the CI of 10090 as validation of this PR.

## PR checklist

- [x] **changelog** done
- [x] **framework PR**
    - [ ] https://github.com/Mbed-TLS/mbedtls-framework/pull/151
    - [ ] https://github.com/Mbed-TLS/mbedtls-framework/pull/161
    - [ ] https://github.com/Mbed-TLS/mbedtls-framework/pull/165
- [x] **mbedtls development PR** https://github.com/Mbed-TLS/mbedtls/pull/10090
- [x] **mbedtls 3.6 PR** not required: new feature
- [ ] **tests**  TODO


drivers/builtin/include/mbedtls/platform.h
```diff
@@ -446,6 +446,32 @@ mbedtls_platform_context;
...
+ * \param[in] output_size       Size of the \p output buffer in bytes.
+ * \param[out] output_len       Amount of bytes effectively written in the
+ *                              \p output buffer.
+ * \param[out] entropy_content  Measure of the entropy content (in bits) of the
+ *                              data written in the \p output buffer. The pointer
+ *                              can be \c NULL in case the caller is not interested
+ *                              in such information, so the function must handle
+ *                              this.
+ */
+int mbedtls_entropy_hardware_poll(unsigned char *output, size_t output_size,
```
gilles-peskine-arm, line 471: Non-blocker: since the function is now declared in `platform.h`, should we call it `mbedtls_platform_xxx`? Same question about the configuration option. Maybe `MBEDTLS_PLATFORM_GET_ENTROPY_ALT` and `mbedtls_platform_get_entropy()`?

drivers/builtin/include/mbedtls/platform.h
```diff
@@ -446,6 +446,32 @@ mbedtls_platform_context;
...
+ *
+ * \warning         This is not provided by Mbed TLS.
+ *
+ * \param[out] output           Output buffer where the entropy data will be
+ *                              stored.
+ * \param[in] output_size       Size of the \p output buffer in bytes.
+ * \param[out] output_len       Amount of bytes effectively written in the
+ *                              \p output buffer.
+ * \param[out] entropy_content  Measure of the entropy content (in bits) of the
+ *                              data written in the \p output buffer. The pointer
```
gilles-peskine-arm, line 466: Please add the constraint that for the time being, the function must set `entropy_content = 8 * *output_len` on success. The library will be extended later to support partial entropy.

drivers/builtin/include/mbedtls/platform.h
```diff
@@ -446,6 +446,32 @@ mbedtls_platform_context;
...
+ *                              can be \c NULL in case the caller is not interested
+ *                              in such information, so the function must handle
+ *                              this.
```
gilles-peskine-arm, line 469: We don't need to support callers that don't want the value. `entropy_content` is guaranteed to be non-null.

drivers/builtin/include/mbedtls/platform.h
```diff
@@ -446,6 +446,32 @@ mbedtls_platform_context;
...
+#if !defined(MBEDTLS_ENTROPY_HARDWARE_POLL_MIN)
+#define MBEDTLS_ENTROPY_HARDWARE_POLL_MIN     32     /**< Minimum for the hardware source */
+#endif
+
+#if defined(MBEDTLS_ENTROPY_HARDWARE_POLL)
+/**
+ * \brief           User defined callback function that is used from the entropy
+ *                  module to gather entropy data from some hardware device.
+ *
+ * \warning         This is not provided by Mbed TLS.
```
gilles-peskine-arm, line 458: I don't think “not provided by Mbed TLS” is a useful warning. Also, it's TF-PSA-Crypto now.

On the other hand, the function's documentation should state that this function is only used when `MBEDTLS_ENTROPY_HARDWARE_POLL` is defined. This is otherwise not visible when you read the Doxygen-rendered documentation.

ChangeLog.d/9618.txt
```diff
@@ -0,0 +1,12 @@
...
+     removed.
+
+Features
+   * The new symbol `MBEDTLS_ENTROPY_HARDWARE_POLL` is added to allow the
+     defintion of a custom callback function which allows to gather entropy
+     data from some hardware device.
+     `MBEDTLS_ENTROPY_HARDWARE_POLL_MIN` can optionally optinally define the
+     minimum number of bytes that should be polled using that function before
+     declaring entropy gathering completed. If not defined, then a default
+     value of 32 bytes is used.
```
gilles-peskine-arm, line 12: `MBEDTLS_ENTROPY_HARDWARE_POLL_MIN` should not be a configuration parameter, at least not in this form. The RNG needs a certain amount of entropy. This minimum is up to the system configuration (and in particular the required cryptographic strength for the system), not up to the hardware integration. The person writing the TRNG device driver doesn't get to choose this value.

I think we should just use the existing `MBEDTLS_ENTROPY_MIN_PLATFORM` here. The distinction between “platform source” and “hardware source” doesn't really make sense as it's currently described in `entropy_poll.h`. Actually the “platform source” is the operating system integrations that the library supports out of the box, and the “hardware source” is the one for other systems where the integrator needs to provide a driver.

ChangeLog.d/9618.txt
```diff
@@ -0,0 +1,12 @@
+Removals
+   * Previously deprecated build symbol `MBEDTLS_ENTROPY_HARDWARE_ALT` is
+     removed.
+
+Features
+   * The new symbol `MBEDTLS_ENTROPY_HARDWARE_POLL` is added to allow the
```
gilles-peskine-arm, line 6: The changelog entries should probably mention the function names as well as the option names, for easier searching.

drivers/builtin/src/entropy_poll.h
```diff
@@ -57,6 +41,21 @@ int mbedtls_nv_seed_poll(void *data,
...
+ * mbedtls_nv_seed_poll). */
+#if defined(MBEDTLS_ENTROPY_HARDWARE_POLL)
+
+#include <mbedtls/platform.h>
+
+static inline int mbedtls_entropy_hardware_poll_wrapper(
+        void *data, unsigned char *output, size_t len, size_t *olen)
+{
+    (void) data;
+    return mbedtls_entropy_hardware_poll(output, len, olen, NULL);
```
gilles-peskine-arm, line 55: The wrapper should check the entropy content and fail if it's less than full entropy, i.e. if `entropy_output < 8 * *olen`. Please add a unit test for that in `test_suite_entropy`.

Also the entropy collection should fail if `*olen` is too small and as a result there isn't enough entropy, collected from all sources. I believe this is covered by the existing tests in `test_suite_entropy`, but please double-check.

gilles-peskine-arm (changes requested):
I just did a design review. I won't be able to do a full code review.

We just realized this morning that the issue description wasn't very clear, and several things were in Janos's mind and mine but we hadn't written them down. Sorry about this. I added a note to the issue and some comments here, hopefully that should clarify things.

drivers/builtin/src/entropy_poll.h
valeriosetti: >The wrapper should check the entropy content and fail if it's less than full entropy, i.e. if entropy_output < 8 * *olen. Please add a unit test for that in test_suite_entropy.

While adding the check on the wrapper and testing it with the new component `tf_psa_crypto_platform_get_entropy_alt` I found that tests are already failing if the returned entropy content is not correct, so there seems to be no need for a new test step. As reference implementation I'm using the one proposed in the companion PR https://github.com/Mbed-TLS/mbedtls-framework/pull/151.

bjwtaylor (approved):

valeriosetti:
I think that CI failures in `test_psasim` and `test_suite_with_psasim` can be ignored in this PR as the same tests [in the main PR](https://github.com/Mbed-TLS/mbedtls/pull/10090) are OK

gabor-mezei-arm (approved):

bjwtaylor (dismissed):

valeriosetti:
@bjwtaylor @gabor-mezei-arm I had to update this PR to solve a conflict in the `framework` reference. Nothing has been changed on the code.

gilles-peskine-arm:
At 0b4ca69b6387c85993a93db1764a8ad88d64f57a there is a weird [failure in `test_suite_with_psasim`](https://mbedtls.trustedfirmware.org/job/mbed-tls-tf-psa-crypto-multibranch/job/PR-212-head/8/execution/node/9281/log/):

```
[2025-04-07T08:56:10.964Z] ../tests/psa-client-server/psasim//test/start_server.sh: line 16: ../tests/psa-client-server/psasim//test/psa_server: No such file or directory
[2025-04-07T12:54:35.506Z] Sending interrupt signal to process
```

At first glance it looks like there are two problems:

1. `psa_server` isn't getting built when it should, or where it should.
2. The script hung after that problem (and eventually the CI timed out), rather than erroring out.

Since you're familiar with psasim, can you please investigate?

valeriosetti:
> * `psa_server` isn't getting built when it should, or where it should.

Yes, the problem in this case is that `psasim`'s server build fails because of the problem explained [above](https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/212#issuecomment-2778608101). 
 
> * The script hung after that problem (and eventually the CI timed out), rather than erroring out.

I just realized that when I worked on the `psasim` implementation I used to test it as:

`./tests/scripts/all.sh --force test_psasim`

in which case the test component stops immediately in case of build failures, so this would not explain why the CI went on testing. It turns out that I basically never add `--keep-going` on the command line, as the CI does instead, and that's the difference.
I need to investigate a bit and find the proper solution. Thanks for noticing.

gilles-peskine-arm:
The way `--keep-going` works uses somewhat arcane bash features and tries to do the right thing by default: stop the component if a configure or build step fails, but keep going if a test step fails. But you sometimes need to help it notice a failure, or decide whether the failure is fatal. The code to set this up is `pre_setup_keep_going` (now in `framework/scripts/all-core.sh`).

Having a quick look at `component_test_psasim`, I notice that there are build steps that look like `make … test/…`. The heuristics in `pre_setup_keep_going` recognize this as a build step. They treat `tests` as a directory name that can appear in a build step and otherwise `test` as an indication of a test step, but that doesn't work for psasim which has a directory called `test`.

But beyond that, if the build works but server can't start, the test should stop, not sit forever until it times out.

valeriosetti:
@gilles-peskine-arm thanks for the detailed explanation. I added a [fix](https://github.com/Mbed-TLS/mbedtls/pull/10090/commits/ccd05e3ef57f589080d08dab8789c3df500816cf) in the [main repo PR](https://github.com/Mbed-TLS/mbedtls/pull/10090). It won't fix the CI of this PR, of course, but it should be OK for the future.

valeriosetti:
I removed the `need-ci` label from this PR because it [cannot](https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/212#issuecomment-2778608101) have CI fully green. It has to be merged manually and then [mbed#10090](https://github.com/Mbed-TLS/mbedtls/pull/10090) must be set on its merge queue immediately after this.

gabor-mezei-arm (dismissed):
LGTM

bjwtaylor (dismissed):

drivers/builtin/src/entropy_poll.h
```diff
@@ -57,6 +41,35 @@ int mbedtls_nv_seed_poll(void *data,
...
+/* Wrapper to allow mbedtls_platform_get_entropy_alt() to be used from entropy
+ * module as other entropy polling functions (i.e. mbedtls_platform_entropy_poll
+ * and mbedtls_nv_seed_poll). */
+#if defined(MBEDTLS_PLATFORM_GET_ENTROPY_ALT)
+
+#include <mbedtls/platform.h>
+
+#define MBEDTLS_ENTROPY_MIN_HARDWARE            32
+
+static inline int mbedtls_hardware_poll(void *data, unsigned char *output,
```
gilles-peskine-arm, line 53: Why is this an inline function in a header file? It's only called inside `entropy.c`, so why not define it there and make it `MBEDTLS_STATIC_TESTABLE`?

ChangeLog.d/9618.txt
```diff
@@ -0,0 +1,9 @@
+Removals
+   * Previously deprecated build symbol `MBEDTLS_ENTROPY_HARDWARE_ALT` is
+     removed.
+
+Features
+   * The new symbol `MBEDTLS_PLATFORM_GET_ENTROPY_ALT` is added to allow the
```
gilles-peskine-arm, line 6: `XXX_ALT` normally allows users to override a built-in feature. But here the function comes in addition to the built-in feature: to replace the built-in feature, you also need to enable `MBEDTLS_NO_PLATFORM_ENTROPY`.

The original design called for leaving `MBEDTLS_NO_PLATFORM_ENTROPY` alone. But come to think of it, we only really care about two use cases at this point: either the default platform entropy works, or it gets replaced. While there might be rare scenarios where both are useful, that can wait until we implement the full RNG driver interface.

There's also the case of having neither the default platform entropy source nor the alternative one, but that is controlled by `MBEDTLS_NO_DEFAULT_ENTROPY_SOURCES` (which will need to change since users no longer have access to entropy objects, but that's a separate matter, to be handled in https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/145).

So I propose to remove `MBEDTLS_NO_PLATFORM_ENTROPY` from `crypto_config.h`, i.e. not make it a configurable symbol anymore, and instead make it an equivalent for `!defined(MBEDTLS_PLATFORM_GET_ENTROPY_ALT)` in `config_adjust_legacy_crypto.h`. And of course adjust the documentation accordingly. We'll simplify the implementation later.

If we don't do that then we should rename the new function and option. But I do think we should take this opportunity to remove one more little piece of legacy.

ChangeLog.d/9618.txt
```diff
@@ -0,0 +1,9 @@
+Removals
+   * Previously deprecated build symbol `MBEDTLS_ENTROPY_HARDWARE_ALT` is
+     removed.
```
gilles-peskine-arm, line 3: Changelog entries for removals should give the name of what to use instead.

drivers/builtin/include/mbedtls/platform.h
```diff
@@ -446,6 +446,32 @@ mbedtls_platform_context;
...
+ *              `MBEDTLS_PLATFORM_GET_ENTROPY_ALT` is defined.
+ *
+ * \param[out] output           Output buffer where the entropy data will be
+ *                              stored.
+ * \param[in] output_size       Size of the \p output buffer in bytes.
+ * \param[out] output_len       Amount of bytes effectively written in the
+ *                              \p output buffer.
+ * \param[out] entropy_content  Measure of the entropy content (in bits) of the
+ *                              data written in the \p output buffer.
+ *
```
gilles-peskine-arm, line 463: Please document the possible return values: 0 for success, `MBEDTLS_ERR_ENTROPY_SOURCE_FAILED` or some other negative value on error.

drivers/builtin/include/mbedtls/platform.h
```diff
@@ -446,6 +446,32 @@ mbedtls_platform_context;
...
+ * \warning     For the time being TF-PSA-Crypto assumes that the function provides
```
gilles-peskine-arm, line 464: I don't find the warning clear as to what happens if `entropy_content` is too small. “TF-PSA-Crypto assumes” — so what happens if the assumption is wrong?

drivers/builtin/src/entropy.c
```diff
@@ -654,7 +654,7 @@ int mbedtls_entropy_self_test(int verbose)
         }
     }
 
-#if defined(MBEDTLS_ENTROPY_HARDWARE_ALT)
+#if defined(MBEDTLS_PLATFORM_GET_ENTROPY_ALT)
     if ((ret = mbedtls_entropy_source_self_test(0)) != 0) {
         goto cleanup;
     }
```
gilles-peskine-arm, line 660: Missing: tests. I'd expect `test_suite_entropy` to have at least a test for when `mbedtls_platform_get_entropy_alt` returns an error code, and one for when it returns less entropy than expected.

ChangeLog.d/9618.txt
```diff
@@ -0,0 +1,9 @@
...
+
+Features
+   * The new symbol `MBEDTLS_PLATFORM_GET_ENTROPY_ALT` is added to allow the
+     defintion of a custom callback function named
+     `mbedtls_platform_get_entropy_alt()` which allows to gather entropy
```
gilles-peskine-arm, line 8: Minor: changelog entries are plain text, not markdown, so no backticks.

ChangeLog.d/9618.txt
```diff
@@ -0,0 +1,9 @@
...
+     data from some hardware device.
```
gilles-peskine-arm, line 9: The hardware device is commonly called a “true hardware random generator” (TRNG). (”True“ is not a good name, but it is commonly used.) This is the term that users will look for, so the changelog entry should mention it.

Suggested wording:

> To supply a platform-specific entropy source, define the compilation option MBEDTLS_PLATFORM_GET_ENTROPY_ALT and provide the callback function mbedtls_platform_get_entropy_alt(). This function should typically access a TRNG ("true hardware random generator") device on bare-metal platforms, or call an operating system function to obtain cryptographic-quality random data.

drivers/builtin/src/entropy_poll.h
valeriosetti: Actually this function is also called from `test_suite_psa_crypto_init.function` in `custom_entropy_init()` function.  If we want to remove the `inline` part, I can move the implementation to `entropy.c` and keep only the definition here. How does this sound?

ChangeLog.d/9618.txt
valeriosetti: IIUC we would like to have `mbedtls_platform_entropy_poll()` and `mbedtls_hardware_poll()` mutually exclusive, right?
If so, do you think it would make sense to rename `mbedtls_platform_entropy_poll()` to `mbedtls_hardware_poll()` so that we can have a single function prototype with different implementations depending on `MBEDTLS_PLATFORM_GET_ENTROPY_ALT`?

gilles-peskine-arm: Yes, merging the two functions would probably make sense, though with a name that matches the compilation option.

drivers/builtin/src/entropy_poll.h
gilles-peskine-arm: `MBEDTLS_STATIC_TESTABLE` lets you call a function from the tests as long as the tests are guarded by `MBEDTLS_TEST_HOOKS`.

valeriosetti:
> * Sorry for the late notice but I think we should make a slight design change and have the new function replace, rather than add to, the default platform entropy. (And if we don't then the new name doesn't fit.)
> 
>     * The documentation is unclear in places.
> 
>     * This doesn't resolve [MBEDTLS_ENTROPY_HARDWARE_ALT in 4.0 mbedtls#9618](https://github.com/Mbed-TLS/mbedtls/issues/9618): it's missing tests and part of the documentation (migration guide). Those can be done in follow-ups, but we need to keep track of the work that's left.

@gilles-peskine-arm I think I've addressed all the requirements so far (including [this](https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/212#discussion_r2042389919)). The only missing thing is the migration document, but before starting it I would like to have a stable design so that I don't have to update documentation too many times.
Can you please take a look at the current state and let me know if it's OK?

drivers/builtin/src/entropy_poll.h
```diff
@@ -18,19 +18,22 @@
...
-int mbedtls_platform_entropy_poll(void *data, unsigned char *output, size_t len, size_t *olen);
-
-#endif /* MBEDTLS_PLATFORM_GET_ENTROPY_ALT */
+/**
+ * This function get called from the entropy module when it's gathering entropy
+ * data. Backends are:
+ * - on Windows: no need to define MBEDTLS_PLATFORM_GET_ENTROPY_ALT and
+ *              CryptoAPI will be used.
+ * - on Linux: no need to define MBEDTLS_PLATFORM_GET_ENTROPY_ALT and
+ *              /dev/urandom will be used.
```
gilles-peskine-arm, line 29: As in the public documentation: not just `/dev/urandom`, but also `getrandom` and variants are preferred. And this is not just on Linux but also some Linux-like systems and most BSD variants.

gilles-peskine-arm (changes requested):
Mostly looks good to me now, apart from some of the documentation.

drivers/builtin/include/mbedtls/platform_util.h
```diff
@@ -190,6 +190,14 @@ struct tm *mbedtls_platform_gmtime_r(const mbedtls_time_t *tt,
                                      struct tm *tm_buf);
 #endif /* MBEDTLS_HAVE_TIME_DATE */
 
+#if !defined(MBEDTLS_PLATFORM_GET_ENTROPY_ALT)
+/* This is the private defintion of mbedtls_platform_get_entropy(). The full
+ * documentation is at drivers/builtin/include/mbedtls/platform.h. */
```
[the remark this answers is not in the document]
valeriosetti: Yeah, the idea was to look for an header file which was only accessible to the library and not to the user in order to hide the prototype of `mbedtls_platform_get_entropy()` when `!MBEDTLS_PLATFORM_GET_ENTROPY_ALT`. Of course after looking around a bit, I fooled myself and I erroneously put that in another public header file. Sorry.
Thinking about it having a single definition in `platform.h` is not that bad. The worst that can happen is that the user has `!MBEDTLS_PLATFORM_GET_ENTROPY_ALT` and they still provide the implementation for `mbedtls_platform_get_entropy()`, but I guess that in this case the compiler will complain about the fact that the same function is implemented twice. So as long as there's a build failure somehow I'm fine :)

valeriosetti:
I removed the `need-ci` label because the PR on the main repo ([10090](https://github.com/Mbed-TLS/mbedtls/pull/10090)) now has CI green.

valeriosetti:
I removed also the `need-work` label because I've just added the migration guide document so I think I've addressed all the comments (@gilles-peskine-arm). I think the PR is ready for reviews again.

drivers/builtin/src/platform.c
```diff
@@ -410,34 +410,37 @@ void mbedtls_platform_teardown(mbedtls_platform_context *ctx)
...
-    *olen = 0;
+    *output_len = 0;
 
     /*
      * BCryptGenRandom takes ULONG for size, which is smaller than size_t on
      * 64-bit Windows platforms. Extract entropy in chunks of len (dependent
      * on ULONG_MAX) size.
      */
-    while (len != 0) {
+    while (output_size != 0) {
```
gilles-peskine-arm, line 431: Non-blocker: since this is now an internal function, `output_size` always comes from our library code, and we can assume that it has a “reasonable” size — tens of bytes, definitely less than `ULONG_MAX` which is 2^32. So the loop is no longer needed (just error out if `output_size` is out of sensible range).

docs/4.0-migration-guide/platform-entropy-gathering.md
```diff
@@ -0,0 +1,63 @@
...
+  - Only useful when running on high level OSes, not in baremetal projects.
+
+- Hardware
+  - Disabled by default, it could be enabled through `MBEDTLS_ENTROPY_HARDWARE_ALT`.
+  - It allowed the user to define a custom function named `mbedtls_hardware_poll()`
+    to gather entropy data from some hardware source.
+  - Very useful for baremetal projects.
+  - Unfortunately the prototype for the function was in a private header file.
+
+- NV seed
```
gilles-peskine-arm, line 25: The NV seed part hasn't changed. It would be useful to mention it in passing, but the document should not suggest that it has changed or that it has been removed.

docs/4.0-migration-guide/platform-entropy-gathering.md
```diff
@@ -0,0 +1,63 @@
+## Platform-specific Entropy Gathering
+
+### Disclaimer
+
+All the following sections refer to the case where entropy module uses default
+entropy sources, i.e. `MBEDTLS_NO_DEFAULT_ENTROPY_SOURCES` is not defined.
+
+### Background
```
gilles-peskine-arm, line 8: I think there's too much detail about what was there before. It's useful to have some explanations, because some readers will be programmers who have to upgrade some part of the application that was written by a long-gone colleague and that they don't understand. But we don't need to explain everything about the old API, especially the parts that haven't changed (e.g. the platform sources are exactly the same as before, what's changed is how you say whether you want to use them).

docs/4.0-migration-guide/platform-entropy-gathering.md
```diff
@@ -0,0 +1,63 @@
...
+- NV seed
+  - Disabled by default, it could be enabled through `MBEDTLS_ENTROPY_NV_SEED`.
+  - It allowed the user to store an entropy seed on a non-volatile (NV) memory
+    (ex: filesystem).
+  - This is crucial on systems that do not have a cryptographic entropy source
+    available through the previous 2 options.
+
+Up to Mbed TLS 3.6 all these 3 build symbols could be enabled independently.
+
+### Reasons to change
```
gilles-peskine-arm, line 34: I don't think a detailed list of reasons is useful in the migration guide. That belongs in an architecture document. The migration guide is a practical document for programmers using the library.

docs/4.0-migration-guide/platform-entropy-gathering.md
```diff
@@ -0,0 +1,63 @@
...
+
+### What changes in TF-PSA-Crypto 1.0?
+
+The following build symbols are removed:
+
+- `MBEDTLS_ENTROPY_HARDWARE_ALT`,
+- `MBEDTLS_ENTROPY_MIN_HARDWARE`,
+- `MBEDTLS_NO_PLATFORM_ENTROPY`.
+
+By default entropy module uses Platform entropy source to gather entropy data.
```
gilles-peskine-arm, line 51: We need to be careful with “by default”: for some users, it means with the default config file, but for other users, it means if you don't define any configuration symbol (empty config file). So we should clarify, e.g. by writing “In the default configuration, the entropy module …”.

docs/4.0-migration-guide/platform-entropy-gathering.md
```diff
@@ -0,0 +1,63 @@
...
+However the user can enable `MBEDTLS_PLATFORM_GET_ENTROPY_ALT` to define their
+own custom implementation of the entropy gathering function and this will
+exclude the Platform one (alternate implementation).
+The prototype of this function can be found in `mbedtls/platform.h` and it's as
```
gilles-peskine-arm, line 55: This is describe a whole new feature, but that's not what is happening and that's not the goal of the migration guide. The target reader already has code that implements the old `mbedtls_hardware_poll`. So what we should explain here is, if you had an application that enabled `MBEDTLS_ENTROPY_HARDWARE_ALT`, how to change an implementation of `mbedtls_hardware_poll` into an implementation of `mbedtls_platform_get_entropy`.

ChangeLog.d/9618.txt
```diff
@@ -0,0 +1,16 @@
+Removals
+   * MBEDTLS_NO_PLATFORM_ENTROPY and previously deprecated
+     MBEDTLS_ENTROPY_HARDWARE_ALT are removed. See documentation of
+     MBEDTLS_PLATFORM_GET_ENTROPY_ALT for a description on how entropy module
+     gathers entropy data.
```
gilles-peskine-arm, line 5: Minor: “and **the** previously deprecated”, “See **the** documentation”, “how **the** entropy module”...

docs/4.0-migration-guide/platform-entropy-gathering.md
```diff
@@ -0,0 +1,63 @@
...
+follows:
+
+```c
+int mbedtls_platform_get_entropy(void *data, unsigned char *output, size_t len, size_t *olen);
```
gilles-peskine-arm, line 59: The prototype is wrong (in fact it lists the arguments for the old `mbedtls_hardware_poll`).

docs/4.0-migration-guide/platform-entropy-gathering.md
```diff
@@ -0,0 +1,63 @@
...
+```
+
+See documentation of this function for futher details about input parameters and
```
gilles-peskine-arm, line 62: There's no need to go into detail, but the migration guide does need to explain the new parameter `entropy_content`. It's likely to be a bit mysterious to many readers.

include/psa/crypto_config.h
```diff
@@ -1057,18 +1057,17 @@
...
- * Your function must be called \c mbedtls_hardware_poll(), have the same
- * prototype as declared in library/entropy_poll.h, and accept NULL as first
- * argument.
- *
- * Uncomment to use your own hardware entropy collector.
+ * By default Mbed TLS uses platform-specific sources such as getrandom(),
+ * /dev/urandom or BCryptGenRandom() to gather entropy data. If these functions
+ * are not available for some reason (ex: working on a baremetal project), the
+ * following symbol allows the user to define a custom callback function named
+ * `mbedtls_platform_get_entropy()` that Mbed TLS will use to gather entropy
```
gilles-peskine-arm, line 1066: To make the function name into a link:
```suggestion
 * mbedtls_platform_get_entropy() that Mbed TLS will use to gather entropy
```

include/psa/crypto_config.h
```diff
@@ -1057,18 +1057,17 @@
...
+ * data. Public header `mbedtls/platform.h` provides the prototype for this
```
gilles-peskine-arm, line 1067: To make the header name into a link:
```suggestion
 * data. The public header mbedtls/platform.h provides the prototype for this
```

include/psa/crypto_config.h
```diff
@@ -1057,18 +1057,17 @@
...
- * Uncomment this macro to let Mbed TLS use your own implementation of a
- * hardware entropy collector.
- *
- * Your function must be called \c mbedtls_hardware_poll(), have the same
- * prototype as declared in library/entropy_poll.h, and accept NULL as first
- * argument.
- *
- * Uncomment to use your own hardware entropy collector.
+ * By default Mbed TLS uses platform-specific sources such as getrandom(),
+ * /dev/urandom or BCryptGenRandom() to gather entropy data. If these functions
```
gilles-peskine-arm, line 1063: Minor:
```suggestion
 * By default Mbed TLS uses platform-specific sources such as `getrandom()`,
 * `/dev/urandom` or `BCryptGenRandom()` to gather entropy data. If these functions
```

include/psa/crypto_config.h
```diff
@@ -1057,18 +1057,17 @@
...
+ * are not available for some reason (ex: working on a baremetal project), the
```
gilles-peskine-arm, line 1064: “for some reason” seems weird to me. We know the reason.
```suggestion
 * the library does not provide an implementation for your platform, the
```

include/psa/crypto_config.h
```diff
@@ -1057,18 +1057,17 @@
...
  *
- * Uncomment this macro to let Mbed TLS use your own implementation of a
- * hardware entropy collector.
- *
- * Your function must be called \c mbedtls_hardware_poll(), have the same
- * prototype as declared in library/entropy_poll.h, and accept NULL as first
- * argument.
- *
- * Uncomment to use your own hardware entropy collector.
+ * By default Mbed TLS uses platform-specific sources such as getrandom(),
```
gilles-peskine-arm, line 1062: The product name is now TF-PSA-Crypto, not Mbed TLS.

(We'll probably make a pass to change those near release time anyway.)

docs/4.0-migration-guide/platform-entropy-gathering.md
gilles-peskine-arm, line 62: There's no need to go into detail, but the migration guide does need to explain the new parameter `entropy_content`. It's likely to be a bit mysterious to many readers.

gilles-peskine-arm (changes requested):
I'd like more changes to the documentation, especially the migration guide.

docs/4.0-migration-guide/platform-entropy-gathering.md
valeriosetti: That's part for the "Background" section indeed where I was trying to give the user an overview of the situation before my changes. I will add a note to clearly state that nothing has changed for NV seed

bjwtaylor (approved):

docs/4.0-migration-guide/platform-entropy-gathering.md
```diff
@@ -0,0 +1,35 @@
...
+                                 size_t *output_len, size_t *entropy_content);
+```
+
+This new prototype is placed in the public header `mbedtls/platform.h` instead
+of the private `library/entropy_poll.h` one as it was previously.
+
+The new `entropy_content` parameter measures (in bit) the amount of entropy
+contained in the returned `output` buffer. For the time being only full entropy
+sources are allowed (i.e. `entropy_content = 8 * *output_len`), but this can
+change in the future.
```
gilles-peskine-arm, line 25: For many readers, “full entropy” is a fuzzy concept at best. So we should attempt to explain it. Here's my proposal.

For the time being, TF-PSA-Crypto only supports callbacks that provide full entropy. That means that the content of the output buffer must be uniformly random, allowing you to report that `entropy_content = 8 * *output_len`. Typically `*output_len == output_size`, but if it is smaller, the library will call the callback again in a loop.

If you implement this callback for hardware that delivers partial entropy, a typical technique is to gather enough entropy then hash the result. Hashing does not increase the entropy, but it distributes it throughout the buffer. For example, suppose your hardware has a 32-bit register which is documented as having a single bit of entropy each time it is read after calling a priming function. The following code snippet collects 8 bits of entropy and outputs it in one byte.

```
int mbedtls_platform_get_entropy(unsigned char *output, size_t output_size,
                                 size_t *output_len, size_t *entropy_content) {
    uint32_t reads[8];
    uint8_t hash[32];
    int ret;
    for (int i = 0; i < 8; i++) {
        ret = myhardware_prime_magic_random_register();
        if (ret != 0) return PSA_ERROR_INSUFFICIENT_ENTROPY;
        reads[i] << *MYHARDWARE_MAGIC_RANDOM_REGISTER_ADDRESS;
    }
    size_t hash_len;
    ret = psa_hash_compute(PSA_ALG_SHA_256, (uint_t*) reads, sizeof(reads), hash, sizeof(hash), &hash_len);
    if (ret != 0) return PSA_ERROR_INSUFFICIENT_ENTROPY;
    output[0] = hash[0];
    *output_len = 1;
    *entropy_content = 8;
    return 0;
}
```

tests/suites/test_suite_entropy.function
```diff
@@ -565,3 +566,19 @@ exit:
...
 }
 /* END_CASE */
+
+/* BEGIN_CASE depends_on:MBEDTLS_PLATFORM_GET_ENTROPY_ALT */
+void entropy_test_platform_get_entropy_alt(void)
+{
+    uint8_t data[MBEDTLS_ENTROPY_POLL_PLATFORM_MIN];
+    size_t olen;
+
+    TEST_ASSERT(mbedtls_entropy_poll_platform(NULL, data, sizeof(data), &olen) == 0);
```
gilles-peskine-arm, line 576: We should validate the value of `olen`.

```suggestion
    TEST_ASSERT(mbedtls_entropy_poll_platform(NULL, data, sizeof(data), &olen) == 0);
    TEST_LE_U(1, olen);
    TEST_LE_U(olen, sizeof(data));
```

gilles-peskine-arm, line 576: Do we have a test somewhere (presumably in `test_suite_entropy.function`) to validate what happens if the callback reports full entropy but with a smaller output size, i.e. `*olen < output_size`? There's a loop in `entropy.c` that should call the function again until it provides enough entropy, but I'm not sure our tests are up to date with the change we've made to the callback: we'd need to let the test callback know to report less entropy, and I don't see this being done.

gilles-peskine-arm (changes requested):
I have more comments in the migration guide. The rest looks good to me, except that reviewing the migration guide has made me revisit a couple of places in the documentation and test coverage.

tests/suites/test_suite_entropy.function
```diff
@@ -567,20 +572,41 @@ exit:
...
+
+    /* Return less data than required -> callback will be called more than once. */
+    mbedtls_entropy_init(&ctx);
+    mbedtls_test_platform_get_entropy_set_output_len(
+        TEST_PLATFORM_GET_ENTROPY_DATA_LENGTH_PER_CALL);
+    TEST_ASSERT(mbedtls_entropy_func(&ctx, data, sizeof(data)) == 0);
+    mbedtls_entropy_free(&ctx);
+    call_count = mbedtls_test_platform_get_entropy_get_call_count();
+    TEST_EQUAL(call_count, TEST_PLATFORM_GET_ENTROPY_EXP_CALL_COUNT);
 }
```
gilles-peskine-arm, line 611: Sorry I missed this earlier: if a test assertion fails, it might leave the entropy callback test in a non-default state, which will likely have weird effects on subsequent tests. Please reset the state in the `exit:` block, that's how we normally deal with test state.

gilles-peskine-arm (changes requested):
Thanks for the updates, looks good to me now except for one little thing I missed earlier.

gilles-peskine-arm (dismissed):
LGTM

bjwtaylor (approved):

valeriosetti:
Remaining CI failures are:
```
[2025-04-29T14:35:20.193Z] test/selftest.c:229:14: error: implicit declaration of function 'mbedtls_platform_entropy_poll' is invalid in C99 [-Werror,-Wimplicit-function-declaration]
[2025-04-29T14:35:20.193Z]     result = mbedtls_platform_entropy_poll(NULL,
```
which is reasonable since the main repo is tested on the `development` branch. We can rely on the CI of https://github.com/Mbed-TLS/mbedtls/pull/10090 to validate this PR.

gilles-peskine-arm:
It's problematic to merge something in the crypto repository that breaks the CI, since it means crypto will be broken on all branches that don't have this PR merged yet.

Can you please add a transition function `mbedtls_platform_entropy_poll` that's just good enough for how it's used in `selftest.c`, with a comment stating that it's going to be removed soon? And might as well make a PR that reverts that commit, to be merged in a few weeks.

valeriosetti:
> Can you please add a transition function mbedtls_platform_entropy_poll that's just good enough for how it's used in selftest.c, with a comment stating that it's going to be removed soon? And might as well make a PR that reverts that commit, to be merged in a few weeks.

@gilles-peskine-arm I just created https://github.com/Mbed-TLS/TF-PSA-Crypto/issues/273 to track this

valeriosetti:
@gilles-peskine-arm although I added the stub function for `mbedtls_platform_entropy_poll` there are remaining failures on the CI related to the fact that the main branch actually needs some more fixes on its codebase to make this PR in (configuration header files, config.py, etc - see https://github.com/Mbed-TLS/mbedtls/pull/10090 for the full list).
I'm not sure I can make the CI fully happy in this PR.,,

gilles-peskine-arm (approved):
LGTM

valeriosetti:
Sorry for the last minute fix, but since I had to [modify](https://github.com/Mbed-TLS/mbedtls/pull/10090/commits/1afedacfea918c47ff55f845a22e95d38d84f836) the PR on the main repo to make its CI happy, I preferred to fix the name of the test component being added also in this PR (see [new commit](https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/212/commits/390da553beaf3f32468e69d565e4c8bb72feeece)).

gilles-peskine-arm (approved):
LGTM

configs/crypto-config-ccm-aes-sha256.h
```diff
@@ -24,5 +24,6 @@
 #define MBEDTLS_PSA_CRYPTO_C
 #define MBEDTLS_CTR_DRBG_C
 #define MBEDTLS_ENTROPY_C
+#define MBEDTLS_PLATFORM_C
```
gilles-peskine-arm, line 27: This shouldn't be necessary. `MBEDTLS_PLATFORM_C` should only be needed to support runtime configuration of platform features. Neither `MBEDTLS_PLATFORM_GET_ENTROPY_ALT` nor `!MBEDTLS_PLATFORM_GET_ENTROPY_ALT` should require `MBEDTLS_PLATFORM_C`.

I think the natural way to fix this is to move the code from `platform.c` to `platform_util.c`.

At this point, given that we're planning to do a disruptive combined merge tonight, I propose to merge as is, and do a follow-up that does the move and removes the added enabling of `MBEDTLS_PLATFORM_C` in both TF-PSA-Crypto and mbedtls.

gilles-peskine-arm: https://github.com/Mbed-TLS/TF-PSA-Crypto/issues/282

bjwtaylor (approved):

gilles-peskine-arm:
As previously discussed, I'm going ahead and force-merging this pull request. This will break the CI on TF-PSA-Crypto where it consumes the head of mbedtls, to be fixed by merging https://github.com/Mbed-TLS/mbedtls/pull/10090.

PR #216, opened 2025-03-21
asn1parse: Require minimal-length encodings of lengths

DemiMarie:
## Description

X.690 explicitly states that DER must use the shortest possible encoding of a length.  For instance, 04 81 00 is not a valid DER encoding of the empty OCTET STRING.  Instead, 04 00 must be used.  With the exception of PKCS#7, Mbed TLS only parses DER, and it is reasonable to require that even PKCS#7 be DER-encoded.

## PR checklist

- [x] **changelog** provided
- [x] **mbedtls development PR** not required because: no change to mbedtls
- [x] **mbedtls 3.6 PR** not required because: it is unclear whether this should be backported to stable releases
- [x] **mbedtls 2.28 PR** not required because: it is unclear whether this should be backported to stable releases
- [x] **tests**  provided

### DIFF
--- a/ChangeLog.d/minimal-length-lengths.txt
+++ b/ChangeLog.d/minimal-length-lengths.txt
@@ -1,0 +1,2 @@
+Bugfix
+  * Invalid overlong encodings of ASN.1 DER lengths are now rejected.
--- a/drivers/builtin/src/asn1parse.c
+++ b/drivers/builtin/src/asn1parse.c
@@ -29,12 +29,16 @@ int mbedtls_asn1_get_len(unsigned char **p,
                          const unsigned char *end,
                          size_t *len)
 {
+    unsigned char lenlen;
+
     if ((end - *p) < 1) {
         return MBEDTLS_ERR_ASN1_OUT_OF_DATA;
     }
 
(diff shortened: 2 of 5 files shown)


gilles-peskine-arm:
Indeed, we do not fix sloppiness in LTS branches if it looks harmless, which seems to be the case here. So we will not backport this patch to 3.6.

This patch is definitely desirable for TF-PSA-Crypto. It's unrelated to https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/184: that issue is solely about INTEGER payload encoding, not about the length encoding.

mpg:
> It's unrelated to https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/184: that issue is solely about INTEGER payload encoding, not about the length encoding.

Ah, good point, I has missed that, thanks.

> PKCS#7 actually allows arbitrary BER, but that also includes indefinite-length encodings, which are already unsupported. A full BER parser would need to be implemented differently.

Agreed, our current parser is not for BER, it's for "sloppy DER". I don't think we have or ever had any intention of parsing arbitrary BER.

PR #240, opened 2025-03-29
Added OID of Ed448 ECDSA signature algorithm

frankencode:
Trust roots are more and more frequently signed with Ed448. Adding the related algorithm identifiers (see RFC 9481) is needed to be able to load certificates signed with Ed448.

## Description

One of my customers signs the trust roots for his embedded appliances with Ed448 signature algorithm. This results in an unknown OID error when loading such certificates. Adding the related algorithm identifier to the OID database fixes that issue.

## PR checklist

- [ ] **changelog** provided | not required because: minimal change
- [ ] **mbedtls development PR** provided Mbed-TLS/mbedtls# | not required because: ?
- [ ] **mbedtls 3.6 PR** provided #[10100 ](https://github.com/Mbed-TLS/mbedtls/pull/10100)
- **tests**  provided | not required because: minimal change

### DIFF
--- a/drivers/builtin/src/oid.c
+++ b/drivers/builtin/src/oid.c
@@ -457,6 +457,12 @@ static const oid_sig_alg_t oid_sig_alg[] =
         MBEDTLS_MD_SHA512,   MBEDTLS_PK_ECDSA,
     },
 #endif /* PSA_WANT_ALG_SHA_512 */
+#if defined(PSA_WANT_ALG_SHA3_256)
+    {
+        OID_DESCRIPTOR(MBEDTLS_OID_ED448,            "Ed448",                "Ed448 with SHA3-256"),
+        MBEDTLS_MD_SHA3_256, MBEDTLS_PK_ECDSA
+    },
+#endif /* PSA_WANT_ALG_SHA3_256 */
 #endif /* PSA_HAVE_ALG_SOME_ECDSA */
 #if defined(MBEDTLS_RSA_C)
     {


minosgalanakis:
Hi and thank you for your contribution. To comply with the project's  [DCO](https://github.com/Mbed-TLS/mbedtls/blob/development/dco.txt) all contributions need to be signed 'git commit -s'. 

Since that is a single commit, could you please do a  `git commit --amend --signoff` ?

frankencode:
Thanks for tip. I've added the sign-off.

drivers/builtin/src/oid.c
```diff
@@ -457,6 +457,12 @@ static const oid_sig_alg_t oid_sig_alg[] =
         MBEDTLS_MD_SHA512,   MBEDTLS_PK_ECDSA,
     },
 #endif /* PSA_WANT_ALG_SHA_512 */
+#if defined(PSA_WANT_ALG_SHA3_256)
+    {
+        OID_DESCRIPTOR(MBEDTLS_OID_ED448,            "Ed448",                "Ed448 with SHA3-256"),
```
gilles-peskine-arm, line 462: Shouldn't we add Ed25519 as well?

Can you please add a comment explaining why we would have an OID for a signature algorithm that the library doesn't support? The use case of a trusted root is valid, but not at all obvious.

gilles-peskine-arm, line 462: Please add a test case that (1) successfully parses a certificate with this OID and (2) fails to verify its signature. (I don't know if this is doable with the existing test functions.) I'm worried that there could be a bug in the library if we assume somewhere that a valid OID means a valid algorithm.

gilles-peskine-arm (changes requested):
Ok for the change (including in 3.6), but I would like a test case.

drivers/builtin/src/oid.c
frankencode: As stated in the description. In particular I have several home appliance manufacturers which are trying to join the HomeConnect network. When trying to load the trust root of the HomeConnect cloud servers on the WiFi module of the home appliance (usually a Cortext-M4) they get the following error:
```
I (4561) Connecting to wss://***.homeconnectegw.com:443/ws?objectid=***..
I (4581) Waiting for connection...
E (4611) mbedtls_x509_crt_parse of CA cert returned -0x262E
E (4611) ESP_ERR_MBEDTLS_X509_CRT_PARSE_FAILED
E (4621) transport_base: Failed to open a new connection
....
```
I found that his error code just means that, when trying to load the ca_cert for the TLS connection, mbedtls bails out, because it does not know the particular OID of the signature algorithm used for the trusted root certificate. Of course this checksum is never verified for the trusted root. So I decided to just add the OID to the table of known OIDs to fix the problem and so far I had no issue with that. Of course I fully expect mbedtls to fail verifying that checksum when it actually is asked to do so (might need another test for this!-). Alternatively any other idea to fix that problem would be great to hear about!

frankencode: OK. These tests I would add to the mbedtls project?! For instance to the `test_suite_x509parse` ?! Could you provide my with some pointers? I barely understand how to build and run these tests.

frankencode:
OK, after trying to approach the problem again from another direction I found a better solution. I've managed to remove the unsupported signatures (and their OIDs;-) from the affected certificates and do not need a patch for mbedtls anymore.  Thanks anyways, good to have mbedtls being opensource. In general I think it is also a good idea to stop processing certificates when an unknown OID is encountered. Please keep that behavior, THX.

Issue #227, opened 2025-03-23
Privatize functions in low-level asymmetric crypto headers


gilles-peskine-arm:
Make functions in asymmetric cryptography headers private, keeping types exposed. See https://github.com/Mbed-TLS/TF-PSA-Crypto/issues/219#issuecomment-2742932065

The following headers are in scope:
```
bignum.h
ecdh.h
ecdsa.h
ecjpake.h
ecp.h
```

Note that `lms.h` remains public, since this mechanism is not in PSA yet.

Note that `rsa.h` is out of scope, since it is becoming fully private.

Prerequisites: https://github.com/Mbed-TLS/mbedtls/issues/10082, https://github.com/Mbed-TLS/TF-PSA-Crypto/issues/219

Issue #226, opened 2025-03-23
Privatize functions in cipher mode headers


gilles-peskine-arm:
Make functions in cipher mode headers private, keeping types exposed. See https://github.com/Mbed-TLS/TF-PSA-Crypto/issues/219#issuecomment-2742932065

The following headers are in scope:
```
block_cipher.h
ccm.h
cmac.h
gcm.h
```

Note that `nist_kw.h` remains public, since this mechanism is not in PSA yet.

Prerequisites: https://github.com/Mbed-TLS/mbedtls/issues/10082, https://github.com/Mbed-TLS/TF-PSA-Crypto/issues/219

Issue #225, opened 2025-03-23
Privatize functions in cipher primitive headers


gilles-peskine-arm:
Make functions in cipher primitive headers private, keeping types exposed. See https://github.com/Mbed-TLS/TF-PSA-Crypto/issues/219#issuecomment-2742932065

The following headers are in scope:
```
aes.h
aria.h
camellia.h
des.h
chacha20.h
chachapoly.h
poly1305.h
```

Prerequisites: https://github.com/Mbed-TLS/mbedtls/issues/10082, https://github.com/Mbed-TLS/TF-PSA-Crypto/issues/219

Issue #224, opened 2025-03-23
Privatize functions in low-level hash headers


gilles-peskine-arm:
Make functions in low-level hash private, keeping types exposed. See https://github.com/Mbed-TLS/TF-PSA-Crypto/issues/219#issuecomment-2742932065

The following headers are in scope:
```
md5.h
ripemd160.h
sha1.h
sha256.h
sha3.h
sha512.h
```

Prerequisites: https://github.com/Mbed-TLS/mbedtls/issues/10082, https://github.com/Mbed-TLS/TF-PSA-Crypto/issues/219

PR #243, opened 2025-03-31
Privatize functions in drivers/builtin/include/mbedtls header files

felixc-arm:
Adds the guard `#if defined(MBEDTLS_DECLARE_PRIVATE_IDENTIFIERS)` around all functions in the files below so that they cannot be accessed by user programs (as they are now private).
~~Also adds this guard to a couple of structs/types that are explicitly marked as internal (in ecp.h, ecdsa.h & block_cipher.h).~~

List of files (for ease of reference, matches the files specified in the linked issues):
- aes.h
- aria.h
- bignum.h
- camellia.h
- cipher.h
- ccm.h
- chacha20.h
- chachapoly.h
- cipher.h
- cmac.h
- des.h
- ecdh.h
- ecdsa.h
- ecjpake.h
- ecp.h
- gcm.h
- md5.h
- poly1305.h
- ripemd160.h
- sha1.h
- sha256.h
- sha3.h
- sha512.h

resolves https://github.com/Mbed-TLS/TF-PSA-Crypto/issues/220, resolves https://github.com/Mbed-TLS/TF-PSA-Crypto/issues/224, resolves https://github.com/Mbed-TLS/TF-PSA-Crypto/issues/225, resolves https://github.com/Mbed-TLS/TF-PSA-Crypto/issues/226, resolves https://github.com/Mbed-TLS/TF-PSA-Crypto/issues/227

## PR checklist

- [ ] **changelog** not required because: Changelog and other documentation to be added with this issue: https://github.com/Mbed-TLS/TF-PSA-Crypto/issues/232
- [ ] **mbedtls development PR** not required because: tf-psa-crypto-only change
- [ ] **mbedtls 3.6 PR** not required because: new API for 4.0
- **tests**  not required because: no internal change to PSA crypto/mbedtls (I guess could have a test that uses the private functions and is expected to fail to compile, but I'm not sure how possible that is with the framework & it would require a 'wrong' example program so may be confusing?)

### DIFF
--- a/drivers/builtin/include/mbedtls/aes.h
+++ b/drivers/builtin/include/mbedtls/aes.h
@@ -84,6 +84,8 @@ typedef struct mbedtls_aes_xts_context {
 } mbedtls_aes_xts_context;
 #endif /* MBEDTLS_CIPHER_MODE_XTS */
 
+#if defined(MBEDTLS_DECLARE_PRIVATE_IDENTIFIERS)
+
 /**
  * \brief          This function initializes the specified AES context.
  *
@@ -577,6 +579,8 @@ MBEDTLS_CHECK_RETURN_CRITICAL
 int mbedtls_aes_self_test(int verbose);
 
 #endif /* MBEDTLS_SELF_TEST */
+
+#endif /* MBEDTLS_DECLARE_PRIVATE_IDENTIFIERS */
 
 #ifdef __cplusplus
 }
--- a/drivers/builtin/include/mbedtls/aria.h
+++ b/drivers/builtin/include/mbedtls/aria.h
@@ -52,6 +52,8 @@ typedef struct mbedtls_aria_context {
 }
 mbedtls_aria_context;
 
+#if defined(MBEDTLS_DECLARE_PRIVATE_IDENTIFIERS)
+
 /**
  * \brief          This function initializes the specified ARIA context.
  *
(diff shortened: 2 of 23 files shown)


mpg:
> resolves #220, resolves #224, resolves #225, resolves #226, resolves #227

I'm sorry, but I don't think doing a single PR for all of them was a good idea. Even though on the surface it looks like the resulting PR is still small (in number of lines change) there is quite a lot to review because the question is exactly what should be hidden - and in some cases there might be non-trivial discussions. We want to avoid too large reviews for several reasons, including quality (experience shows we miss more things in larger PRs) and planning (spreading the load, making incremental progress).

In general, we should not address more than one issue with a single PR. When we're tempting to do so, we should generally ask for opinions - if possible from the person who created the issues, as there's usually a reason why the work was split that way. As always there are exceptions (for example dependencies were missed when splitting tasks, so it turns out they were split in an inconvenient way), but I don't think that's one of them.

As a result, I'm closing this PR, please re-raise one PR per issue in order to facilitate review. (As a reminder, we have a general rule to prioritize reviewer convenience over author convenience, since review is much more often the bottleneck.)

felixc-arm:
Sure, no worries :slightly_smiling_face:

gilles-peskine-arm:
> the question is exactly what should be hidden - and in some cases there might be non-trivial discussions

Specifically, I think we always want to exclude functions. We mostly want to include types but perhaps not all of them, although that's not a big deal. But macros could go either way and getting that wrong could have subtle consequences (e.g. depending on header inclusion order, different `.c` files could end up having incompatible opinions on a data structure's layout, because they aren't getting the same `#define`s).

PR #245, opened 2025-04-02
Add fuzz progs

bjwtaylor:
## Description

Add fuzz programs, which have been removed from mbedtls to tf-psa-crypto. Contributes to https://github.com/Mbed-TLS/mbedtls/issues/10067 depends https://github.com/Mbed-TLS/mbedtls/pull/10115

This PR is part of a chain which needs to be merged in the following order:

1. https://github.com/Mbed-TLS/mbedtls/pull/10115
2. https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/245
3. https://github.com/Mbed-TLS/mbedtls/pull/10205

## PR checklist

- [ ] **changelog** not required
- [ ] **mbedtls development PR** provided Mbed-TLS/mbedtls https://github.com/Mbed-TLS/mbedtls/pull/10115
- [ ] **mbedtls 3.6 PR** not required because: No backports
- [ ] **mbedtls 2.28 PR** not required because: No backports
- **tests**  not required because: No changes


bjwtaylor:
I've looked in to doing this as a non dependent merge, however it would probably require splitting in to 3 PR's as you cannot remove without re-adding in tf-psa-crypto as there are dependencies on the moved files. You also cannot add before you remove as you then generate a duplicate binary issue in the build. It's possible to do it with some renaming or a 3 stage merge, however this may not be worthwhile. Let me know what people think.

valeriosetti (dismissed):

programs/fuzz/CMakeLists.txt
```diff
@@ -0,0 +1,47 @@
...
+
+    # This emulates "if ( ... IN_LIST ... )" which becomes available in CMake 3.3
+    list(FIND executables_with_common_c ${exe} exe_index)
+    if(${exe_index} GREATER -1)
+        list(APPEND exe_sources fuzz_common.c)
+    endif()
+
+    add_executable(${exe} ${exe_sources} $<TARGET_OBJECTS:tf_psa_crypto_test>)
+    set_base_compile_options(${exe})
+    target_include_directories(${exe} PRIVATE ../../core/
```
felixc-arm, line 37: Is `../../core` needed here? - it wasn't needed when the programs were in `mbedtls` but something might have changed that I'm missing...

felixc-arm (commented):
Generally LGTM, although I think the programs need to be added back to `tf-psa-crypto/programs/.gitignore`.

Also I guess in response to the comment on this patch from a month and a half ago, it seems that we are are doing 'one extra patch to avoid interdependent patches that break the CI', so presumably the corresponding `mbedtls` patch should be split in two - one that removes or renames these programs, merge this `TF-PSA-Crypto` PR, and then another that consumes the two programs that would then be in `tf-psa-crypto` - or another option that avoids the two-way dependencies, I haven't thought through the process too much :slightly_smiling_face:.

programs/fuzz/CMakeLists.txt
bjwtaylor: Agree, I'm not sure we need it. I've removed it and the local build succeeds, let's see if it causes complaints in the CI though.

valeriosetti (dismissed):

felixc-arm (commented):
LGTM apart from one minor comment issue :+1:

felixc-arm (approved):
LGTM, thanks!

valeriosetti (approved):

programs/fuzz/fuzz_onefile.c
```diff
@@ -0,0 +1,81 @@
...
+    const char *argv0 = argv[0] == NULL ? "PROGRAM_NAME" : argv[0];
+
+    if (argc != 2) {
+        fprintf(stderr, "Usage: %s REPRODUCER_FILE\n", argv0);
+        return 1;
+    }
+    //opens the file, get its size, and reads it into a buffer
+    #ifdef WIN32
+    errno_t err = 0;
+    err = fopen_s(&fp, argv[1], "rb");
```
gilles-peskine-arm, line 28: Why would we bother with `fopen_s` in this program? This is a test program where security is irrelevant. We don't bother with it elsewhere in the library. Not that `fopen_s` improves on `fopen` in a relevant way anyway.

bjwtaylor: So, this was a while back. Though I seem to remember it had something todo with being windows compatible?

Issue #223, opened 2025-03-23
Move public legacy headers to /include


gilles-peskine-arm:
Move the public legacy headers from `/drivers/builtin/include/mbedtls` to `/include/mbedtls`.

This is just a file move, since they're moving to the same relative path under a different directory that is also on the include path.

See https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/145 for a study of which headers should remain public. The list is:

```
asn1.h
asn1write.h
base64.h
constant_time.h
lms.h
memory_buffer_alloc.h
nist_kw.h
pem.h
pk.h
platform.h
platform_time.h
platform_util.h
psa_util.h
threading.h
```

ronald-cron-arm:
If we move them to `include/` like `include/asn1.h` then we will need to change how they are included from something like `#include "mbedtls/asn1.h"` to `#include "asn1.h"`.
If we move them to `include/mbedtls` then no code to change for their inclusion but as they define public APIs of TF-PSA-Crypto they would fit better into `include/tf-psa-crypto`.
Looking at #145 it is not clear me what has been decided yet.

mpg:
Indeed I'm not sure the exact location of the public headers has been explicitly discussed yet. From this issue's description "they're moving to the same relative path under a different directory" it seems what Gilles had in mind was `include/mbedtls` (even if I agree that's not immediately clear from the first sentence).

I agree that `include/asn1.h` would not be a good idea, and I see how `include/tf-psa-crypto` could make sense.

However I'm inclined to stick with Gilles's suggestion of `include/mbedtls`, at least for now: that's the lowest-effort thing we can do that clearly marks those headers as public. I doesn't prevent us from moving them to `include/tf-psa-crypto` later (but before 1.0 of course) if we can find the time and think it's worth it.

ronald-cron-arm:
These headers define the TF-PSA-Crypto APIs they should be in `include/tf-psa-crypto`, `include/mbedtls` is just going to be confusing. I understand the lowest-effort argument but when I look at https://github.com/Mbed-TLS/mbedtls/issues/10087 it does not seem consistent then.

mpg:
[...]

PR #247, opened 2025-04-04
Move public legacy headers to /include/mbedtls

felixc-arm:
Moves public headers from `/drivers/builtin/include/mbedtls` to `/include/mbedtls` to make it clear that they are public.
Resolves https://github.com/Mbed-TLS/TF-PSA-Crypto/issues/223, although there may be further work to move these headers somewhere else, e.g. `/include/tf-psa-crypto`.

List of files moved for reference:
```
asn1.h
asn1write.h
base64.h
constant_time.h
lms.h
memory_buffer_alloc.h
nist_kw.h
pem.h
pk.h
platform.h
platform_time.h
platform_util.h
psa_util.h
threading.h
```

## PR checklist

- [ ] **changelog** not required because: no change in user-facing functionality
- [ ] **framework PR** provided: https://github.com/Mbed-TLS/mbedtls-framework/pull/156
- [ ] **mbedtls development PR** provided: https://github.com/Mbed-TLS/mbedtls/pull/10122
- [ ] **mbedtls 3.6 PR** not required because: work for 4.0/1.0
- **tests**  not required because: no functional changes


gilles-peskine-arm:
There are probably build scripts that need to be updated. For example `include/CMakeLists.txt` has
```
    file(GLOB tf-psa-crypto_headers "tf-psa-crypto/*.h")
    file(GLOB mbedtls_crypto_headers "../drivers/builtin/include/mbedtls/*.h")
```
and presumably there needs to be a line for `"mbedtls/*.h"`. I don't know whether it should be one of these variables or yet another one: I don't know what the difference is between those variables.

felixc-arm:
The companion PR https://github.com/Mbed-TLS/mbedtls/pull/10122 has now passed the CI

bjwtaylor (dismissed):

valeriosetti (changes requested):
The PR looks mostly OK to me. I only left 1 question

include/CMakeLists.txt
```diff
@@ -3,7 +3,7 @@ option(INSTALL_TF_PSA_CRYPTO_HEADERS "Install TF PSA Crypto headers." ON)
 if(INSTALL_TF_PSA_CRYPTO_HEADERS)
     file(GLOB psa_headers "psa/*.h")
     file(GLOB tf-psa-crypto_headers "tf-psa-crypto/*.h")
-    file(GLOB mbedtls_crypto_headers "../drivers/builtin/include/mbedtls/*.h")
+    file(GLOB mbedtls_crypto_headers "../drivers/builtin/include/mbedtls/*.h" "mbedtls/*.h")
```
valeriosetti, line 6: Question: if `include/mbedtls/` is going to hold all the public header files, why do we need to keep installing also the private ones?

gilles-peskine-arm: There are three kinds of definitions:

* Public — things that applications can use.
* Exposed — not part of the documented API, but needed to implement documented APIs. Typical example: context types, where `mbedtls_pk_context pk; mbedtls_pk_init(&pk);` needs to work so we have to have the definition in an installed header, but we don't want to document the fields of `mbedtls_pk_context`.
* Internal — only needs to be visible when building the library itself (and invasive tests, and programs that haven't been updated yet).

See https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/145 for more explanations.

valeriosetti: Oh OK, thanks for the link. This was the [detail](https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/145/files#diff-6282df272fdb6697fc70d40912d945b4ab7b2e9071a96d956103c6319289b49dR361) I was missing and that solves my doubts.

valeriosetti (dismissed):

gilles-peskine-arm:
The CI is unhappy here. Is the companion mbedtls pull request needed to make it happy?

felixc-arm:
Yep, there's https://github.com/Mbed-TLS/mbedtls/pull/10122 for mbedtls & https://github.com/Mbed-TLS/mbedtls-framework/pull/156 for the framework. Both of those should be merge-able without any dependencies, so I guess the process would be merge framework -> update mbedtls PR's tf-psa-crypto pointer to development head rather than this patch & update framework pointer to the merge commit -> wait for mbedtls to pass CI without this tf-psa-crypto commit -> merge mbedtls PR -> update framework pointer to the merge commit for this tf-psa-crypto PR which should then make this CI green -> merge this PR.

Does this seem right?

gilles-peskine-arm:
@felixc-arm That's great! Please do this then. When it's just a matter of splitting PR like this, it's less disruptive than requiring branch updates to be synchronized — that tends to break many pull requests for unrelated work in progress.

gilles-peskine-arm:
There's a merge conflict in the framework submodule, please update it.

There are also CI failures which look like they might be related to this change.

felixc-arm:
@gilles-peskine-arm Yep, just waiting on https://github.com/Mbed-TLS/mbedtls/pull/10122 to go in so that the `mbedtls` changes are in place, then this patch should be green. (+ I'll update the framework pointer at that point as well)

valeriosetti (dismissed):

bjwtaylor (dismissed):

felixc-arm:
I've added another commit to no longer render the non-public headers based on the discussion here: https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/238#discussion_r2068061550

I've checked by manually following the commands in `apidoc_full.sh`, i.e. (from TF-PSA-Crypto root)
```
python3 scripts/config.py realfull
cd doxygen
rm -rf build-apidoc-full
mkdir build-apidoc-full
cd build-apidoc-full
cmake -DCMAKE_BUILD_TYPE:String=Check -DGEN_FILES=ON ../../
make tfpsacrypto-apidoc
```
Then go to TF-PSA-Crypto-root/apidoc and open `index.html` in a browser

And it all looks correct to me. The corresponding change in `mbedtls` will happen once this is merged and I do a clean-up patch that removes the unneeded conditionals (i.e. the Doxygen change will be in this cleanup patch as well).

felixc-arm:
Hmm... there are (Doxygen) errors/warnings due to explicit Doxygen links between headers in `tf-psa-crypto/include` and types/macros in `tf-psa-crypto/drivers/builtin/include/mbedtls`. The only realistic fix I can find is to just unlink these references in `tf-psa-crypto/include` (and `mbedtls/include` when appropriate) and leave them as plain-text, but I just want to check that this is a reasonable solution as it would leave the documentation in a bit of a weird position, i.e. referencing types/macros that are not themselves documented/are non-public.

If there are plans to redo/update the documentation to remove references to non-public types/macros then this should be fine, otherwise it might need a bit of thought...

Open to suggestions (including just reverting this commit and dealing with it later...) :slightly_smiling_face: 
cc @gilles-peskine-arm

gilles-peskine-arm:
That's a really good question! Every complaint from Doxygen is saying that we are making a promise to users that they have no way of understanding. Some of them should be resolved by changing the documentation — for example the references to `MBEDTLS_RSA_SALT_LEN_ANY` in `pk.h` are only in documentation for private fields, which shouldn't really be documented even today. But others need an API change — for example `MBEDTLS_ERR_ECP_IN_PROGRESS` should become an internal synonym of `PSA_OPERATION_INCOMPLETE`.

Given that https://github.com/Mbed-TLS/TF-PSA-Crypto/commit/63b57173ecd53e87596a60008c332c869a133245 has been approved and has passed the CI, I propose to go back to it and merge it (if it hasn't become stale), then handle the documentation privatization as a separate task. WDYT @felixc-arm @mpg ?

mpg:
Sounds like a plan. Indeed those doxygen errors are actual issues that we want to address (as opposed to just silence), so let's do it - but proceed step by step, starting with getting this PR merged without the doxygen change for now.

valeriosetti (approved):

gilles-peskine-arm:
The failure in https://mbedtls.trustedfirmware.org/blue/organizations/jenkins/mbed-tls-tf-psa-crypto-multibranch/detail/PR-247-head/9/pipeline/1211 doesn't look like a glitch:

```
/var/lib/build/core/psa_crypto.c:21:10: fatal error: 'psa_crypto_driver_wrappers.h' file not found
#include "psa_crypto_driver_wrappers.h"
         ^
/var/lib/build/core/psa_crypto.c:21:10: fatal error: 'psa_crypto_driver_wrappers.h' file not found
#include "psa_crypto_driver_wrappers.h"
         ^
1 error generated.
core/CMakeFiles/tfpsacrypto_static.dir/build.make:71: recipe for target 'core/CMakeFiles/tfpsacrypto_static.dir/psa_crypto.c.o' failed
make[2]: *** [core/CMakeFiles/tfpsacrypto_static.dir/psa_crypto.c.o] Error 1
CMakeFiles/Makefile2:329: recipe for target 'core/CMakeFiles/tfpsacrypto_static.dir/all' failed
make[1]: *** [core/CMakeFiles/tfpsacrypto_static.dir/all] Error 2
make[1]: *** Waiting for unfinished jobs....
[ 31%] Building C object core/CMakeFiles/tfpsacrypto.dir/psa_crypto_client.c.o
1 error generated.
core/CMakeFiles/tfpsacrypto.dir/build.make:71: recipe for target 'core/CMakeFiles/tfpsacrypto.dir/psa_crypto.c.o' failed
make[2]: *** [core/CMakeFiles/tfpsacrypto.dir/psa_crypto.c.o] Error 1
make[2]: *** Waiting for unfinished jobs....
CMakeFiles/Makefile2:290: recipe for target 'core/CMakeFiles/tfpsacrypto.dir/all' failed
make[1]: *** [core/CMakeFiles/tfpsacrypto.dir/all] Error 2
make: *** [all] Error 2
Makefile:138: recipe for target 'all' failed
^^^^test_tf_psa_crypto_shared: build/test: shared libraries: make -> 2^^^^
```

It is weird that the internal CI passed though. A non-deterministic failure when building could be a missing build dependency. On the CI, we run `make -j2` (at least in builds initiated by `make`, I'm not completely sure this applies to CMake builds). With parallel builds, sometimes, missing build dependencies are only a problem if the build is executed in a particular order, and are not a problem with sequential builds.

This pull request moves files in a way that could affect dependencies — for example maybe somewhere dependencies are calculated in a way that doesn't take `/include/mbedtls` into account. Can you please check carefully whether that might be the case?

Experimentally, you may want to try doing build locally with various levels of parallelism and see if it breaks. This can be very sensitive to the level of parallelism, the number of CPUs, how fast the builds are and other factors. For example we used to have a missing dependency that would in practice never be a problem for `make -j2` if there was an odd number of files to build, but caused problems frequently if there was an even number! (Or vice versa, I don't remember — but the parity of the number of files was what made `make -j2` work or not in practice.)

felixc-arm:
I'll have a look, but I do remember seeing this exact failure on another patch (found it!: https://mbedtls.trustedfirmware.org/blue/organizations/jenkins/mbed-tls-framework-multibranch/detail/PR-153-head/16/pipeline/2066 for https://github.com/Mbed-TLS/mbedtls-framework/pull/153) that is unrelated, so I don't think it is this specific patch that is causing it at least.

felixc-arm:
Sure, opened https://github.com/Mbed-TLS/TF-PSA-Crypto/issues/286 - and I'll rerun the OpenCI now I have permissions to hopefully ensure it's only an intermittent failure... :slightly_smiling_face:

felixc-arm:
The Windows CI seemed to have issues with rerunning after https://github.com/Mbed-TLS/mbedtls-test/pull/203 was merged as it was trying to rerun Windows 2013 jobs whilst looking for 'v141' which I think is Visual Studio 2017 which wasn't installed.
So I've rebased it (again... :slightly_smiling_face:) to hopefully re-trigger with the Windows 2017 jobs instead and get everything green.

bjwtaylor (approved):

felixc-arm:
I see, good to know the 'build now' button will do a fresh job, cheers!

Issue #211, opened 2025-03-18
Can `psa_mac_` APIs internally use an external cipher driver?


Harshal5:
As far as I can understand for the `psa_mac_` APIs to internally use an external cipher driver we would need to write a custom MAC driver. Writing a custom MAC driver would involve almost rewriting the `mbedtls_psa_mac_` APIs and just replacing `mbedtls_cipher_` API calls with `psa_cipher_` so that the external cipher driver could be used. 

Instead is it possible that to update the `mbedtls_cipher_cmac_` API's usage of `mbedtls_cipher_` with `psa_cipher_` APIs? Thus allowing the usage of an external cipher driver with `psa_mac_` API calls.

Harshal5:
So while migrating from the `mbedcrypto` module of mbedtls-3.6.2 to TF-PSA-Crypto 1.0 if we introduce the driver dispatch layer for AES, we would not be able to disable `MBEDTLS_AES_C` because the CMAC implementation would need it for successful execution (We were considering to disable `MBEDTLS_AES_C` to minimize the code size impact).
It looks like the `mbedtls` migration increases our code size significantly, what recommendations would you suggest for reducing this code size impact?

gilles-peskine-arm:
Disabling built-in mechanisms is possible in Mbed TLS 3.6 in most common configurations, but not all. I don't know the details off the top of my head. This is documented in the [driver-only builds guide](https://github.com/Mbed-TLS/mbedtls/blob/mbedtls-3.6.2/docs/driver-only-builds.md). For CMAC, it seems that you need to provide a CMAC driver if you want to remove `MBEDTLS_AES_C`. We plan to lift this limitation and other similar ones eventually, but it won't happen in the 3.6.x branch.

Harshal5:
@gilles-peskine-arm I have raised PRs to facilitate thee CMAC implementation to call PSA functions for the block cipher in ECB mode, including driver dispatch if there's a driver for that block cipher. PTAL!!

PR #249, opened 2025-04-08
feat(drivers): Builtin CMAC driver use PSA cipher interface

Harshal5:
## Description

Current issue:

As far as I can understand for the `psa_mac_` APIs to internally use an external cipher driver we would need to write a custom complete MAC driver. Writing a custom MAC driver would involve almost rewriting the `mbedtls_psa_mac_` APIs.

Solution:

Instead we could update the builtin driver's `mbedtls_psa_mac_` API usage of `mbedtls_cipher_` with `psa_cipher_` APIs. This would allow the usage of an external cipher driver with `psa_mac_` API calls without rewriting the CMAC code.

## Related

Fixes https://github.com/Mbed-TLS/TF-PSA-Crypto/issues/211

## PR checklist

- [ ] **framework PR** provided: https://github.com/Mbed-TLS/mbedtls-framework/pull/155
- [ ] **mbedtls 3.6 PR** not required 
- [ ] **mbedtls 2.28 PR** not required 
- **tests**  not required because: Current tests follow the updated code workflow


Harshal5:
@gilles-peskine-arm I have now updated the code workflow now, could you PTAL once?

The major change in the current version of the workflow as compared to the earlier one is that instead of adding the `PSA_KEY_USAGE_ENCRYPT` attribute to the CMAC key, I have now created a new instance/copy of CMAC key with type AES and the encrypt attribute for allowing us to pass the key to the cipher APIs. This key is created in the `_mac_start()` API and the key id is stored in the operation context, finally is destroyed when `_mac_abort()` API is called for the operation context (`_mac_finish` also calls `_mac_abort` internally).

Therefore, now we do not need the "_hack_" of adding the encrypt attribute to the CMAC key thus solving the incorrect key policy issue. Also, even if the CMAC API layer holds a lock on the CMAC key, the cipher API layer can use the key for performing encrypt operations and can hold a lock as well.

Here is the current workflow for an PSA CMAC operation that can internally use an external cipher driver as well:
`psa_mac_` (PSA API) -> `psa_mac_driver_wrapper_` (driver dispatch) -> `mbedtls_psa_mac_` (builtin driver) -> `psa_cipher_` (PSA API) -> `psa_cipher_driver_wrapper_` (driver dispatch) -> `mbedtls_psa_cipher_` (builtin driver).

Having this change would also help disabling `MBEDTLS_AES_C` when using a custom AES driver layer thus reducing the code size significantly.

Let me know what do you think about this approach!

Harshal5:
@gilles-peskine-arm @yanesca Could you PTAL?

PR #252, opened 2025-04-09
Move X.509 OIDs out of crypto

gilles-peskine-arm:
Move the definitions of OIDs that are only needed X.509 out of `oid.c` and `oid.h` out of TF-PSA-Crypto, and into the X.509 library in Mbed TLS.

Note: this pull request does not work against the head of Mbed TLS. After it is merged, TF-PSA-Crypto can only be consumed by an Mbed TLS where https://github.com/Mbed-TLS/mbedtls/pull/10125 is merged. As a consequence, the TF-PSA-Crypto CI will fail. Arranging for a compatibility period would be doable, but quite a bit of extra work.

## PR checklist

- [x] **changelog** not required because: internal interfaces only (`oid.h` isn't yet officially a private interface, but [it will be](https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/145))
- [x] **development PR** https://github.com/Mbed-TLS/mbedtls/pull/10125
- [x] **TF-PSA-Crypto PR** here
- [x] **3.6 PR** not required because: product split


drivers/builtin/include/mbedtls/oid.h
```diff
@@ -470,39 +443,6 @@
 extern "C" {
 #endif
```
gilles-peskine-arm, line 445: A note for potential future improvements: the functions declared here are only used inside TF-PSA-Crypto, except `mbedtls_oid_get_md_alg()` which is also used in X.509 and PKCS7 in Mbed TLS.

gilles-peskine-arm: Actually `mbedtls_oid_get_md_alg()` is not used in crypto! It's only called from `test_suite_md` which validates the OID associated with each hash algorithm. So `mbedtls_oid_get_md_alg()` should move to X.509 as well.

gilles-peskine-arm:
I plan to rework the history to reach largely the same content, as described in https://github.com/Mbed-TLS/mbedtls/issues/10124#issuecomment-2858492592. The main change is the macros will be duplicated across the two repositories, instead of creating `oid_definition_helpers.h`.

PR #254, opened 2025-04-10
Don't call mbedtls_test_hook_error_add()

gilles-peskine-arm:
Fix an omission in https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/167 which is likely to cause test failures when we change more functions to return PSA error codes.

## PR checklist

- [x] **changelog** provided | not required because: test stuff
- [x] **mbedtls development PR** not required because: will be automatically consumed by mbedtls, needs no further work there
- **tests**  provided

### DIFF
--- a/drivers/builtin/include/mbedtls/error_common.h
+++ b/drivers/builtin/include/mbedtls/error_common.h
@@ -62,8 +62,6 @@ extern "C" {
  * \param low       Low-level error code, i.e. error code returned by
  *                  a lower-level function.
  *                  This can be 0 to just return a high-level error.
- * \param file      file where this error code combination occurred.
- * \param line      line where this error code combination occurred.
  */
 static inline int mbedtls_error_add(int high, int low)
 {
--- a/drivers/builtin/include/mbedtls/error_common.h
+++ b/drivers/builtin/include/mbedtls/error_common.h
@@ -35,7 +35,7 @@ extern "C" {
  *        more details.
  */
 #define MBEDTLS_ERROR_ADD(high, low) \
-    mbedtls_error_add(high, low, __FILE__, __LINE__)
+    mbedtls_error_add(high, low)
 
 #if defined(MBEDTLS_TEST_HOOKS)
 /**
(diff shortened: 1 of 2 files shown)


mpg (approved):
LGTM, thanks!

This is resolving the most pressing issue, but not cleaning everything up. I think at some point we should remove all traces of `mbedtls_test_hook_error_add` and `mbedtls_test_err_add_check()` from tf-psa-crypto and mbedtls/development, and guard all occurrences in mbedtls-framework with "If mbedtls version < 4". Doesn't have to be now, but then we probably want an issue to track it?

gilles-peskine-arm:
> I think at some point we should remove all traces of mbedtls_test_hook_error_add and mbedtls_test_err_add_check() from tf-psa-crypto

Isn't that what I did? Oh, I forgot to remove the declaration. Updated.

> and mbedtls/development

I thought there'd be no impact, but there's a now-useless variable psasim, I'll make a follow-up for that.

> and guard all occurrences in mbedtls-framework with "If mbedtls version < 4".

Why? I'd prefer not to have explicit version dependencies if it's not necessary. That will disrupt existing branches, for example. The framework just offers the hook, and it's harmless if a branch doesn't consume that hook.

mpg (approved):
LGTM

mpg:
> > and guard all occurrences in mbedtls-framework with "If mbedtls version < 4".
> 
> Why? I'd prefer not to have explicit version dependencies if it's not necessary. That will disrupt existing branches, for example. The framework just offers the hook, and it's harmless if a branch doesn't consume that hook.

You're right, let's not add a guard if we don't need to.

drivers/builtin/include/mbedtls/error_common.h
```diff
@@ -76,17 +65,8 @@ extern void (*mbedtls_test_hook_error_add)(int, int, const char *, int);
  * \param file      file where this error code combination occurred.
  * \param line      line where this error code combination occurred.
  */
-static inline int mbedtls_error_add(int high, int low,
-                                    const char *file, int line)
+static inline int mbedtls_error_add(int high, int low)
```
mpg, line 68: Clang (in the CI) points out we should remove `file` and `line` from the documentation as well.

mpg (changes requested):

mpg (approved):

bjwtaylor (approved):

PR #260, opened 2025-04-17
Move the inclusion of crypto_sizes.h and crypto_struct.h in crypto.h

ronald-cron-arm:
## Description
Move the inclusion of crypto_sizes.h and crypto_struct.h in crypto.h
That way when API are declared, the types they used are defined.
This should resolve the issues related to psa_xyz_init functions returning a structure described in Mbed-TLS/mbedtls#7087.

## PR checklist
- [x] **changelog** provided
- [x] **mbedtls development PR** not required because: crypto-only change not affecting mbedtls
- [x] **mbedtls 3.6 PR** provided Mbed-TLS/mbedtls#10196
- **tests**  not required because: the change is validated by the current tests. We can't afford the effort required for non-regression testing right now. Covered by  https://github.com/Mbed-TLS/mbedtls/issues/1767.

### DIFF
--- a/include/psa/crypto.h
+++ b/include/psa/crypto.h
@@ -59,6 +59,18 @@ extern "C" {
  * of integral types defined in "crypto_types.h". */
 #include "crypto_values.h"
 
+/* The file "crypto_sizes.h" contains definitions for size calculation
+ * macros whose definitions are implementation-specific. */
+#include "crypto_sizes.h"
+
+/* The file "crypto_struct.h" contains definitions for
+ * implementation-specific structs that are declared above. */
+#if defined(MBEDTLS_PSA_CRYPTO_STRUCT_FILE)
+#include MBEDTLS_PSA_CRYPTO_STRUCT_FILE
+#else
+#include "crypto_struct.h"
+#endif
+
 /** \defgroup initialization Library initialization
  * @{
  */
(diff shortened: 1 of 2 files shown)


include/psa/crypto.h
```diff
@@ -39,6 +39,18 @@ extern "C" {
  * algorithms, key types, policies, etc. */
 #include "crypto_types.h"
 
+/* The file "crypto_sizes.h" contains definitions for size calculation
+ * macros whose definitions are implementation-specific. */
+#include "crypto_sizes.h"
```
gilles-peskine-arm, line 44: `crypto_sizes.h` uses macros from `crypto_values.h`, so it would make more sense to include `crypto_values.h` first. I don't think it actually matters for the sake of the C language, because `crypto_sizes.h` itself only defines macros, it doesn't expand them. But then `crypto_struct.h` does expand macros from `crypto_sizes.h`, and that can result in expanding macros from `crypto_values.h`. It works because `crypto_struct.h` (indirectly) includes `crypto_driver_common.h` which includes `crypto_values.h`. But it would make more sense to include `crypto_values.h` at the logical place in `crypto.h`.

ronald-cron-arm: Right, I missed that, thanks. I've changed it and while at it I've tried to fix the similar issues in crypto_extra.h pointed to by #7087.

gilles-peskine-arm (dismissed):
LGTM for the code change. I don't know if it actually fixes https://github.com/Mbed-TLS/mbedtls/issues/7087, but at worst it's harmless.

If this does fix https://github.com/Mbed-TLS/mbedtls/issues/7087 then we need a changelog entry, and preferably a non-regression test (which would resolve https://github.com/Mbed-TLS/mbedtls/issues/1767).

gilles-peskine-arm:
We have a [report that this does fix 7087](https://github.com/Mbed-TLS/mbedtls/issues/7087#issuecomment-2859193386). Can you please add a changelog entry and make a backport to 3.6?

aslze:
> We have a [report that this does fix 7087](https://github.com/Mbed-TLS/mbedtls/issues/7087#issuecomment-2859193386). Can you please add a changelog entry and make a backport to 3.6?

Can confirm. I had the same issue and just replacing "crypto_extra.h" and "crypto.h" in version 3.6.3 fixed it.

irwir:
@ronald-cron-arm
Since you already rearranging declarations, would it be possible to improve consistency as requested in the [issue #7841](https://github.com/Mbed-TLS/mbedtls/issues/7481)?

ronald-cron-arm:
I've removed the changes related to crypto_extra.h as they are going to be done in #258.

irwir:
Apparentely, there is no need for any extensive changes; deleting the declarations already resolves the issue.

gilles-peskine-arm (approved):
LGTM

davidhorstmann-arm (approved):
LGTM, thanks!

PR #265, opened 2025-04-23
Privatize functions and some types in hash headers

felixc-arm:
Privatizes all functions in the following files by guarding them with `MBEDTLS_DECLARE_PRIVATE_IDENTIFIERS`

`md5.h`
`ripemd160.h`
`sha1.h`
`sha256.h`
`sha3.h`
`sha512.h`

Also privatizes the type `mbedtls_sha3_id` in `sha3.h` - All of the context types in these files are found in `mbedtls_psa_hash_operation_t` in the file `tf-psa-crypto/include/psa/crypto_builtin_primitives.h` and so are kept unprivatized/internal.
Otherwise there are no other types/macros in these files.

Resolves https://github.com/Mbed-TLS/TF-PSA-Crypto/issues/224

## PR checklist

- [ ] **changelog** not required because: will come later encompassing all privatization using `MBEDTLS_DECLARE_PRIVATE_IDENTIFIERS`
- [ ] **mbedtls development PR** not required because: crypto change only
- [ ] **mbedtls 3.6 PR** not required because: 4.0/1.0 only change

### DIFF
--- a/drivers/builtin/include/mbedtls/sha3.h
+++ b/drivers/builtin/include/mbedtls/sha3.h
@@ -28,6 +28,7 @@ extern "C" {
 /** SHA-3 input data was malformed. */
 #define MBEDTLS_ERR_SHA3_BAD_INPUT_DATA                 -0x0076
 
+#if defined(MBEDTLS_DECLARE_PRIVATE_IDENTIFIERS)
 /**
  * SHA-3 family id.
  *
@@ -41,6 +42,7 @@ typedef enum {
     MBEDTLS_SHA3_384, /*!< SHA3-384 */
     MBEDTLS_SHA3_512, /*!< SHA3-512 */
 } mbedtls_sha3_id;
+#endif /* MBEDTLS_DECLARE_PRIVATE_IDENTIFIERS */
 
 /**
  * \brief          The SHA-3 context structure.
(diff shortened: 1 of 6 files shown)


drivers/builtin/include/mbedtls/sha3.h
```diff
@@ -41,6 +42,7 @@ typedef enum {
     MBEDTLS_SHA3_384, /*!< SHA3-384 */
     MBEDTLS_SHA3_512, /*!< SHA3-512 */
 } mbedtls_sha3_id;
+#endif /* MBEDTLS_DECLARE_PRIVATE_IDENTIFIERS */
```
bjwtaylor, line 45: Minor point: I'm assuming that the middle section here can't be privatised, in which case it may be worth moving the enum def down as it doesn't appear to be needed in the struct. Then you can scrap the second if defined.

bjwtaylor (approved):

drivers/builtin/include/mbedtls/sha3.h
felixc-arm: Yeah, I was working with the 'keep everything in place and just add the guards to minimize changes' thought, but I'll do the change here if I need to update this patch for another reason.

valeriosetti (approved):

PR #277, opened 2025-05-02
Fix bug that caused build to break with -DUSE_SHARED_TF_PSA_CRYPTO_LIBRARY=ON

felixc-arm:
This bug caused errors like:
```
/usr/bin/ld: ../../drivers/builtin/libbuiltin.so: undefined reference to `psa_asymmetric_encrypt'
/usr/bin/ld: ../../drivers/builtin/libbuiltin.so: undefined reference to `psa_crypto_driver_pake_get_user'
collect2: error: ld returned 1 exit status
make[2]: *** [programs/test/CMakeFiles/which_aes.dir/build.make:101: programs/test/which_aes] Error 1
make[1]: *** [CMakeFiles/Makefile2:1127: programs/test/CMakeFiles/which_aes.dir/all] Error 2
```
When building with `-DUSE_SHARED_TF_PSA_CRYPTO_LIBRARY=ON`, e.g. the test `component_test_tf_psa_crypto_shared` in `component-build-system.sh` should have picked it up, I found it locally on Ubuntu 22.

It wasn't caught by the CI, presumably because the CI was running it on Ubuntu 16 and the older version of `ld` didn't notice it...

Originally introduced in https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/199

## PR checklist

- [ ] **changelog** not required because: Bug fix in 4.0/1.0 work, not present in releases
- [ ] **mbedtls development PR** not required because: tf-psa-crypto change only
- [ ] **mbedtls 3.6 PR** not required because: tf-psa-crypto change only
- **tests**  not required because: Can't easily run same test on newer Ubuntu (?)

### DIFF
--- a/programs/test/CMakeLists.txt
+++ b/programs/test/CMakeLists.txt
@@ -16,6 +16,7 @@ target_link_libraries(benchmark PRIVATE ${tfpsacrypto_target} ${CMAKE_THREAD_LIB
 
 # which_aes
 tfpsacrypto_build_program_common(which_aes)
+target_sources(which_aes PRIVATE $<TARGET_OBJECTS:tf_psa_crypto_test>)
 target_link_libraries(which_aes PRIVATE ${tfpsacrypto_target})
 
 # dlopen


felixc-arm:
The new job has failed as expected on commit `ac140e5` : https://mbedtls.trustedfirmware.org/blue/organizations/jenkins/mbed-tls-tf-psa-crypto-multibranch/detail/PR-277-head/4/pipeline/1199 - so the new job works in catching the error, hopefully after applying the bug fix the CI (including this new job) will be green...

felixc-arm:
@valeriosetti Hmm I think there's something strange going on in `tf-psa-crypto-outcomes.csv` - I can't see any results in that file for tests that run on non-16 Ubuntu (e.g. `tf-psa-crypto-all_u18-tf_psa_crypto_check_code_style` & `tf-psa-crypto-all_u18-tf_psa_crypto_test_psa_compliance`). *Actually I think it's probably because those components don't run the "actual" tests, they run python scripts that check other things, and the new component doesn't run tests either it just checks the build, so they don't have any test results that need to be put in that file.

I see the new test is also not in that file, however it is definitely being run on the CI, e.g. [the last job before the windows jobs](https://mbedtls.trustedfirmware.org/blue/organizations/jenkins/mbed-tls-tf-psa-crypto-multibranch/detail/PR-277-head/5/pipeline/1199) is `tf-psa-crypto-all_u22-build_tf_psa_crypto_shared_newer_ld_gcc`, and it also shows up in the [timestamps.json artifact](https://mbedtls-builds.trustedfirmware.org/static-files/dTXvcRVs5c9EIoAveIQiiBSAB_v6GTGVoXFHd1d0N1AxNzQ3ODM2MDM1Njg1Ojk6YW5vbnltb3VzOmpvYi9tYmVkLXRscy10Zi1wc2EtY3J5cHRvLW11bHRpYnJhbmNoL2pvYi9QUi0yNzctaGVhZC81L2FydGlmYWN0/timestamps.json) (hopefully the links don't die before I finish writing this...)

I was basing this work on a reply I got from @mpg:

> If Ubuntu 22 is enough to reproduce the issue, we have it in the CI. If you can write an all.sh component the reproduces the issue, then you can force it to run on a recent enough Ubuntu by using a support_ function. You can have a look at existing support functions for inspiration. If we want to require Ubuntu >= 22 we can probably take inspiration from support_test_cmake_out_of_source but I guess here it would probably cleaner to request a recent enough version of ld if you've identified that to be the culprit.

So it's definitely running on Ubuntu 22 (also running & failing in the link in my first comment as I hadn't put in the CMake fix yet), but yes it is a bit counter-intuitive in the way you have to force it to use a newer OS or tool.

valeriosetti:
> So it's definitely running on Ubuntu 22 (also running & failing in the link in my first comment as I hadn't put in the CMake fix yet), but yes it is a bit counter-intuitive in the way you have to force it to use a newer OS or tool.

Oh, you are right, I didn't see that line for `tf-psa-crypto-all_u22-build_tf_psa_crypto_shared_newer_ld_gcc` before (sorry!) so the test is definitely executed. Therefore my comment above is invalid.

I only have a minor proposal left: instead of naming the component `build_tf_psa_crypto_shared_newer_ld_gcc` can't we pick something more specific than `newer`? `newer` is really relative to the current status and it's not really future proof to me. Wdyt?

--- Question to other reviewers ---
I wonder how this actually works: I checked `all-core.sh` and it seems to me that the `support_` function is acting as a "filter" as I mentioned before (it's used to generate the `SUPPORTED_COMPONENTS` list). So how can the CI determine which tests are to be executed in which Docker image? Or put differently, how can each Docker image >= 18 no re-execute tests which are already done for the `u16` case?

programs/test/CMakeLists.txt
```diff
@@ -16,7 +16,8 @@ target_link_libraries(benchmark PRIVATE ${tfpsacrypto_target} ${CMAKE_THREAD_LIB
 
 # which_aes
 tfpsacrypto_build_program_common(which_aes)
-target_link_libraries(which_aes PRIVATE ${tfpsacrypto_target})
+target_sources(which_aes PRIVATE $<TARGET_OBJECTS:tf_psa_crypto_test>)
```
ronald-cron-arm, line 19: I think it would be better to not do that as this rather hides than fixes the issue. The core issue seems that we have two separate libraries, in the C compiler/linker sense, for the PSA core and the driver builtin that reference each other and it is linker dependent how this is handled. We should probably have only one library and in the CMake scripts to define the builtin library as an object library. I will create an issue for that.

mpg:
> I wonder how this actually works: I checked `all-core.sh` and it seems to me that the `support_` function is acting as a "filter" as I mentioned before (it's used to generate the `SUPPORTED_COMPONENTS` list). So how can the CI determine which tests are to be executed in which Docker image? Or put differently, how can each Docker image >= 18 no re-execute tests which are already done for the `u16` case?

The magic for that lives in the mbedtls-test repo,  specifically [this function](https://github.com/Mbed-TLS/mbedtls-test/blob/main/vars/common.groovy#L227). Basically it runs `all.sh --list-all-components` somewhere to get a complete list of components, then it has an ordered list of platforms, runs `all.sh --list-components` on each platform to know what each platforms supports, then for each component it picks the first platform in the list that supports it.

That's most of what the pre-test-checks stage does (after loading and/or rebuilding the Docker images, and before running the actual tests). If one component from `--list-all-components` has no platform that supports it, the run is aborted with an error. Otherwise, we move to the testing stage where for each component, `all.sh <component_name>` is executed on the platform that was determined for this component earlier.

These days the list has Ubuntu 16 at the front, so all `all.sh` components run on Ubuntu 16 unless their support function tells it's not possible, then other platforms from the list are tried, until a suitable one is found - in this case Ubuntu 22 will be the first one.

(This is for Linux, I think FreeBSD is handled differently: it runs a hardcoded subset of `all.sh` that also runs on Linux. Also, even on Linux there are some more subtleties, like components with `_release_` in their name being excluded from PR jobs, etc.)

I hope that helps. Feel free to browse the code and/or ask more questions obviously!

mpg:
> Actually I think it's probably because those components don't run the "actual" tests, they run python scripts that check other things, and the new component doesn't run tests either it just checks the build, so they don't have any test results that need to be put in that file.

Yes, currently only test results are recorded in the outcomes file, not build results, which is annoying for all kinds of reasons, including this. As usual, Gilles has [a PR for that](https://github.com/Mbed-TLS/mbedtls-framework/pull/129), which we have not been reviewing so it's not been merged and we keep running in the issue... If you feel like reviewing it, it would help :)

gilles-peskine-arm:
> Hmm I think there's something strange going on in tf-psa-crypto-outcomes.csv - I can't see any results in that file for tests that run on non-16 Ubuntu (e.g. tf-psa-crypto-all_u18-tf_psa_crypto_check_code_style & tf-psa-crypto-all_u18-tf_psa_crypto_test_psa_compliance). *Actually I think it's probably because those components don't run the "actual" tests, they run python scripts that check other things, and the new component doesn't run tests either it just checks the build, so they don't have any test results that need to be put in that file.

Right. Currently the outcome file only records outcomes from running unit tests, `ssl-opt.sh` and `compat.sh`. Other things, such as builds, are not recorded (there's a [stalled PR for that](https://github.com/Mbed-TLS/mbedtls-framework/pull/129)).

> I checked all-core.sh and it seems to me that the support_ function is acting as a "filter" as I mentioned before

Right, this is documented [in `all.sh`](https://github.com/Mbed-TLS/mbedtls-framework/blob/23970e03b6ed58f50ab14e73d9b19218de7d147f/scripts/all-core.sh#L95) (and only there I think, it would be useful to have a more comprehensive and discoverable document somewhere).

programs/test/CMakeLists.txt
ronald-cron-arm: Here is the issue: https://github.com/Mbed-TLS/TF-PSA-Crypto/issues/300

PR #285, opened 2025-05-07
[tf-psa-crypto] Remove MBEDTLS_USE_PSA_CRYPTO from PK module

### LINKED ISSUE #283: Remove MBEDTLS_USE_PSA_CRYPTO from PK module
### Suggested enhancement

Since MBEDTLS_USE_PSA_CRYPTO will be assumed to be always enabled its usage in the PK module can be removed.

### Justification

TF-PSA-Crypto needs this because it removes dead code and simplifies code reading.

valeriosetti:
## Description

Resolves #283

## PR checklist

- [x] **changelog** not required because: internal
- [x] **development PR** not required because: crypto only
- [x] **TF-PSA-Crypto PR** provided HERE
- [x] **3.6 PR** not required because: 1.0 only
- **tests**  not required because: removing dead code

### DIFF
--- a/tests/suites/test_suite_pk.function
+++ b/tests/suites/test_suite_pk.function
@@ -15,7 +15,7 @@
 #include <limits.h>
 #include <stdint.h>
 
-/* Needed only for test case data under #if defined(MBEDTLS_USE_PSA_CRYPTO),
+/* Needed only for test case data using PSA API macros/types because
  * but the test code generator requires test case data to be valid C code
  * unconditionally (https://github.com/Mbed-TLS/mbedtls/issues/2023). */
 #include "psa/crypto.h"
@@ -27,12 +27,6 @@
 
 /* Needed for the definition of MBEDTLS_PK_WRITE_PUBKEY_MAX_SIZE. */
 #include "pkwrite.h"
-
-#if defined(MBEDTLS_RSA_C) ||                                           \
-    defined(MBEDTLS_ECDSA_C) ||                                         \
-    defined(MBEDTLS_USE_PSA_CRYPTO)
-#define PK_CAN_SIGN_SOME
-#endif
 
 /* MBEDTLS_TEST_PK_PSA_SIGN is enabled when:
  * - The build has PK_[PARSE/WRITE]_C for RSA or ECDSA signature.
@@ -609,7 +603,7 @@ exit:
  * END_DEPENDENCIES
  */
 
-/* BEGIN_CASE depends_on:MBEDTLS_USE_PSA_CRYPTO */
+/* BEGIN_CASE */
 void pk_psa_utils(int key_is_rsa)
 {
     mbedtls_pk_context pk, pk2;
@@ -712,7 +706,7 @@ exit:
 }
 /* END_CASE */
 
-/* BEGIN_CASE depends_on:MBEDTLS_USE_PSA_CRYPTO */
+/* BEGIN_CASE */
 void pk_can_do_ext(int opaque_key, int key_type, int key_usage, int key_alg,
                    int key_alg2, int curve_or_keybits, int alg_check, int usage_check,
                    int result)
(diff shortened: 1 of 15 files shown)


tests/suites/test_suite_pk.function
```diff
@@ -28,12 +28,6 @@
 /* Needed for the definition of MBEDTLS_PK_WRITE_PUBKEY_MAX_SIZE. */
 #include "pkwrite.h"
 
-#if defined(MBEDTLS_RSA_C) ||                                           \
-    defined(MBEDTLS_ECDSA_C) ||                                         \
-    defined(MBEDTLS_USE_PSA_CRYPTO)
-#define PK_CAN_SIGN_SOME
```
bjwtaylor, line 34: Whats the thinking behind the removal of #define PK_CAN_SIGN_SOME? The above statement with MBEDTLS_USE_PSA_CRYPTO always defined should always evaluate to true, so is there another reason this needs to be remove?

tests/suites/test_suite_pk.function
```diff
@@ -944,34 +938,22 @@ exit:
...
     mbedtls_pk_init(&alt);
     PSA_INIT();
 
-#if defined(MBEDTLS_USE_PSA_CRYPTO)
-    /* mbedtls_pk_check_pair() returns either PK or ECP error codes depending
-       on MBEDTLS_USE_PSA_CRYPTO so here we dynamically translate between the
-       two */
-    if (ret == MBEDTLS_ERR_ECP_BAD_INPUT_DATA) {
-        ret = MBEDTLS_ERR_PK_BAD_INPUT_DATA;
-    }
```
bjwtaylor, line 964: Whats the thinking behind the removal of #define block? The above statement with MBEDTLS_USE_PSA_CRYPTO always defined should always evaluate to true, so is there another reason this needs to be removed?

tests/suites/test_suite_pk.function
```diff
@@ -1222,7 +1183,7 @@ exit:
 }
 /* END_CASE */
 
-/* BEGIN_CASE depends_on:PSA_WANT_ALG_SHA_256:PK_CAN_SIGN_SOME */
```
bjwtaylor, line 1225: Tied to the previous comment, though is there some reason for removing PK_CAN_SIGN_SOME

bjwtaylor (commented):
Generally looks good, just had a few questions on some the removals that didn't seem related to the issue. Suspect they are fine though.

tests/suites/test_suite_pk.function
valeriosetti: Since `MBEDTLS_USE_PSA_CRYPTO` this `#if defined()` check always evaluates to true so `PK_CAN_SIGN_SOME`. Therefore I can replace all potential checks of `PK_CAN_SIGN_SOME` with `true`

tests/suites/test_suite_pk.function
valeriosetti: In the past (i.e. up to 3.6) `mbedtls_pk_check_pair()` could return either `MBEDTLS_ERR_ECP_BAD_INPUT_DATA` or `MBEDTLS_ERR_PK_BAD_INPUT_DATA` based on `MBEDTLS_USE_PSA_CRYPTO` (as specified in the comment that I removed in that block). Since now `MBEDTLS_USE_PSA_CRYPTO` is always set, this error code conversion would always take place. Therefore instead of doing this extra step, I set the proper error code in `test_suite_pk.data` on the test `Check pair #2 (EC, bad)`.

tests/suites/test_suite_pk.function
bjwtaylor: No, makes sense.

bjwtaylor (approved):

mpg:
Again conflicts due to `pk.h` being move. Should not be complex to resolve, but perhaps a bit tedious if `git rebase` fails to understand what's going on.

gilles-peskine-arm:
I'd expect `git rebase` or `git merge` to work. Modern and even not-so-modern versions of git have no problem with a file being moved on one side and modified on the other. I guess it's a limitation of the old fork of git that runs on GitHub.

tests/suites/test_suite_pk.function
```diff
@@ -15,7 +15,7 @@
 #include <limits.h>
 #include <stdint.h>
 
-/* Needed only for test case data under #if defined(MBEDTLS_USE_PSA_CRYPTO),
+/* Needed only for test case data using PSA API macros/types because
```
mpg, line 18: Minor: "because but" (over the line break). Actually I'd remove the comment entirely, I think these days we don't need to justify including `psa/crypto.h`.

mpg (approved):
LGTM, thanks!

valeriosetti:
- 1st forced push:
    - address @mpg [comment](https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/285#discussion_r2090867871)
    - remove update of the `framework`

- 2nd forced push: rebase on top of `development`to solve the conflict

mpg (approved):
LGTM, based on the output of `echo "*/pk.h" > orderfile && git -c diff.orderfile=orderfile range-diff development 52782305fd80547eea575fa198d8f52d751a4ec2 public/pr/285` and manually reviewing the final commit.

bjwtaylor (approved):

Issue #286, opened 2025-05-07
Intermittent CI failure in job test_tf_psa_crypto_shared


felixc-arm:
The CI job `tf-psa-crypto-all_u16-test_tf_psa_crypto_shared` has occasional failures (in both OpenCI & Internal CI) with the build error:

```
[ 31%] Building C object core/CMakeFiles/tfpsacrypto.dir/psa_crypto.c.o
/var/lib/build/core/psa_crypto.c:21:10: fatal error: 'psa_crypto_driver_wrappers.h' file not found
 #include "psa_crypto_driver_wrappers.h"
         ^
 /var/lib/build/core/psa_crypto.c:21:10: fatal error: 'psa_crypto_driver_wrappers.h' file not found
 #include "psa_crypto_driver_wrappers.h"
         ^
1 error generated.
core/CMakeFiles/tfpsacrypto_static.dir/build.make:71: recipe for target 'core/CMakeFiles/tfpsacrypto_static.dir/psa_crypto.c.o' failed
make[2]: *** [core/CMakeFiles/tfpsacrypto_static.dir/psa_crypto.c.o] Error 1
CMakeFiles/Makefile2:329: recipe for target 'core/CMakeFiles/tfpsacrypto_static.dir/all' failed
make[1]: *** [core/CMakeFiles/tfpsacrypto_static.dir/all] Error 2
make[1]: *** Waiting for unfinished jobs....
 31%] Building C object core/CMakeFiles/tfpsacrypto.dir/psa_crypto_client.c.o
1 error generated.
core/CMakeFiles/tfpsacrypto.dir/build.make:71: recipe for target 'core/CMakeFiles/tfpsacrypto.dir/psa_crypto.c.o' failed
make[2]: *** [core/CMakeFiles/tfpsacrypto.dir/psa_crypto.c.o] Error 1
make[2]: *** Waiting for unfinished jobs....
CMakeFiles/Makefile2:290: recipe for target 'core/CMakeFiles/tfpsacrypto.dir/all' failed
make[1]: *** [core/CMakeFiles/tfpsacrypto.dir/all] Error 2
make: *** [all] Error 2
Makefile:138: recipe for target 'all' failed
 ^^^^test_tf_psa_crypto_shared: build/test: shared libraries: make -> 2^^^^
```

[...]

PR #288, opened 2025-05-07
Prepare crypto for the OID module split

gilles-peskine-arm:
Prepare for `mbedtls/oid.h` to be in Mbed TLS instead of TF-PSA-Crypto. We want this because crypto OIDs don't need to be public, but some X.509 OIDs should be. This is an early step on the way to https://github.com/Mbed-TLS/mbedtls/issues/10124, which needs to be done before the bulk of the work in mbedtls.

In crypto, the OID module will become fully private, so API breaks are fine here.

This is just a small pull request to enable the mbedtls work. Enforcing the private nature of `mbedtls/oid.h` will be done in a follow-up, once mbedtls no longer needs `mbedtls/oid.h` to exist in TF-PSA-Crypto.

## PR checklist

- [x] **changelog** provided
- [x] **mbedtls development PR** not required because: just crypto stuff
- **tests**  provided (this must not break mbedtls)

### DIFF
--- a/drivers/builtin/include/mbedtls/crypto_oid.h
+++ b/drivers/builtin/include/mbedtls/crypto_oid.h
@@ -7,8 +7,8 @@
  *  Copyright The Mbed TLS Contributors
  *  SPDX-License-Identifier: Apache-2.0 OR GPL-2.0-or-later
  */
-#ifndef MBEDTLS_OID_H
-#define MBEDTLS_OID_H
+#ifndef MBEDTLS_CRYPTO_OID_H
+#define MBEDTLS_CRYPTO_OID_H
 #include "mbedtls/private_access.h"
 
 #include "tf-psa-crypto/build_info.h"
@@ -692,4 +692,4 @@ int mbedtls_oid_get_pkcs12_pbe_alg(const mbedtls_asn1_buf *oid, mbedtls_md_type_
 }
 #endif
 
-#endif /* oid.h */
+#endif /* crypto_oid.h */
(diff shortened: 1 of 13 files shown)


drivers/builtin/include/mbedtls/crypto_oid.h
```diff
@@ -0,0 +1,698 @@
+/**
+ * \file oid.h
```
felixc-arm, line 2: Minor nit - I guess this should be `crypto_oid.h`, otherwise LGTM.

I also noticed that the failure mentioned before (https://github.com/Mbed-TLS/TF-PSA-Crypto/issues/286) cropped up again (on the internal CI this time)...

drivers/builtin/include/mbedtls/crypto_oid.h
```diff
@@ -0,0 +1,698 @@
...
+
+#include "mbedtls/md.h"
+
+/* OID is not found. This error is in the process of being removed from
+ * the public API, Hide it from scripts/generate_errors.pl, which
+ * is in the mbedtls repository, and would complain because it doesn't
+ * recognize the header name "crypto_oid.h". */
+#define MBEDTLS_ERR_OID_NOT_FOUND                         (-0x002E)
+/* output buffer is too small */
+#define MBEDTLS_ERR_OID_BUF_TOO_SMALL (PSA_ERROR_BUFFER_TOO_SMALL)
```
gilles-peskine-arm, line 33: 7a022a6ddfe8e04af7f8595087c62cd18234fe97 broke uncrustify's heuristics about alignment. Don't ask me why

```
#define MBEDTLS_ERR_OID_BUF_TOO_SMALL                     PSA_ERROR_BUFFER_TOO_SMALL
```
doesn't work but adding parentheses does.

felixc-arm (approved):

Issue #281, opened 2025-05-06
Ensure Doxygen documentation doesn't render files in drivers/builtin/include/mbedtls


felixc-arm:
Change `tfpsacrypto.doxyfile.in` in `TF-PSA-Crypto` & `mbedtls.doxyfile` in `mbedtls` so that files in `drivers/builtin/include/mbedtls` are no longer rendered.

This task depends on https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/247 being merged so that the public headers are under `/include/mbedtls` and are therefore rendered.

This work will also include changing Doxygen comments in other files so that the documentation builds without errors/warnings and does not contain any references to (newly) non-public content (i.e. macros/types/functions). See https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/247#issuecomment-2851543251.

This task will likely need an `mbedtls` PR and a `TF-PSA-Crypto` PR, although they should be independent of eachother.

To build the Doxygen docs (at least for TF-PSA-Crypto as there is no CI job at the moment):
from `tf-psa-crypto` root
```
python3 scripts/config.py realfull
cd doxygen
rm -rf build-apidoc-full
mkdir build-apidoc-full
cd build-apidoc-full
cmake -DCMAKE_BUILD_TYPE:String=Check -DGEN_FILES=ON ../../
make tfpsacrypto-apidoc
```
and then open `apidoc/index.html` in a browser. (n.b. The warnings that will cause the CI to fail are not very obvious in the wall of build logs.)

PR #294, opened 2025-05-15
Remove links to private functions/types/macros in public Doxygen comments

felixc-arm:
Part of implementing https://github.com/Mbed-TLS/TF-PSA-Crypto/issues/281, ~depends on https://github.com/Mbed-TLS/mbedtls/pull/10183~ (merged)

Changes the `mbedtls` errors `MBEDTLS_ERR_ECP_IN_PROGRESS` & `MBEDTLS_ERR_RSA_VERIFY_FAILED` to be aliases of the `PSA` errors `PSA_OPERATION_INCOMPLETE` & `PSA_ERROR_INVALID_SIGNATURE`.
This is so that Doxygen comments in public headers no-longer reference functions/types/macros in non-public headers:

Places that referenced the above `mbedtls` errors now reference the equivalent `PSA` errors.
Places that referenced `MBEDTLS_MD_NONE` have been removed - these were all in comments of functions that seem to be getting replaced by PSA equivalents and the reference is not super important, although I see that in `0e-plans.md` a subset of `md.h` might become public. If this is the case then this change is not needed (although of course we would need to wait for the `md.h` change to do this one).

Other changes should be more self-explanatory. 

## PR checklist

- [ ] **changelog** provided (for the error alias change)
- [ ] **mbedtls development PR** provided: HERE
- [ ] **mbedtls 3.6 PR** not required because: 4.0/1.0 work only
- **tests**  not required because: Documentation changes

### DIFF
--- a/core/psa_crypto.c
+++ b/core/psa_crypto.c
@@ -325,11 +325,11 @@ static int psa_is_dh_key_size_valid(size_t bits)
 
 psa_status_t mbedtls_to_psa_error(int ret)
 {
-    /* PSA error codes (except PSA_SUCCESS) exist in the space between
-     * PSA_ERROR_GENERIC_ERROR and PSA_OPERATION_INCOMPLETE, so if we are
-     * passed a PSA error code then just return it. */
-    if (ret <= PSA_ERROR_GENERIC_ERROR && ret >= PSA_OPERATION_INCOMPLETE) {
-        return (psa_status_t)ret;
+    /* Only legacy error codes need to be translated.
+     * Those are either a low-level error code (-127..-2)
+     * or a high-level error code (<= -0x1000). */
+    if (ret > -0x1000 && ret < -0x80) {
+        return (psa_status_t) ret;
     }
     /* Mbed TLS error codes can combine a high-level error code and a
      * low-level error code. The low-level error usually reflects the


felixc-arm:
I also had to add a bit to `mbedtls_to_psa_error` to account for the fact that `PSA_ERROR_xxx`s may be input - The way I checked if an error was `MBEDTLS` or `PSA` was the easiest way I could see, but if there's a nicer way to distinguish `MBEDTLS` and `PSA` errors just by using the numerical error code value then I'll update that patch.

core/psa_crypto.c
```diff
@@ -325,6 +325,12 @@ static int psa_is_dh_key_size_valid(size_t bits)
 
 psa_status_t mbedtls_to_psa_error(int ret)
 {
+    /* PSA error codes (except PSA_SUCCESS) exist in the space between
+     * PSA_ERROR_GENERIC_ERROR and PSA_OPERATION_INCOMPLETE, so if we are
+     * passed a PSA error code then just return it. */
+    if (ret <= PSA_ERROR_GENERIC_ERROR && ret >= PSA_OPERATION_INCOMPLETE) {
+        return (psa_status_t)ret;
+    }
```
gilles-peskine-arm, line 333: Using specific error names is fragile because we might add another status that's e.g. `PSA_OPERATION_INCOMPLETE - 1`.

All legacy error codes are either in the range -127..-2 or -32767..-1024. See [`error .h`](https://github.com/Mbed-TLS/mbedtls/blob/mbedtls-3.6.3/include/mbedtls/error.h#L25) up to 3.x. There was a gap between -1024 and -128 (0 as the high-level module ID) that never got used. Going forward, if we add new error values, they'll be in the unified error space from the start. So I think we can assume that only values in these legacy ranges need to be translated.

So I would suggest
```suggestion
    /* Only legacy error codes need to be translated.
     * Those are either a low-level error code (-127..-2)
     * or a high-level error code (<= -0x1000).
     */
    if (ret > -0x1000 && ret < -0x80) {
        return (psa_status_t)ret;
    }
```

felixc-arm: Makes sense, cheers!

felixc-arm:
Rebased to fix GCC 15 tests, should be good to review. (Just to make it 100% clear this PR doesn't actually make any changes to what files doxygen renders, this is just pre-work to remove any errors/warnings that would occur - the change to doxygen file rendering will come (for both `TF-PSA-Crypto` and `mbedtls` after this PR is merged.)

include/mbedtls/pk.h
```diff
@@ -476,16 +476,15 @@ int mbedtls_pk_can_do_ext(const mbedtls_pk_context *ctx, psa_algorithm_t alg,
  *                    based on the #mbedtls_pk_type_t type of \p pk,
  *                    the chosen \p usage and other factors:
  *                      - #MBEDTLS_PK_RSA whose underlying
- *                        #mbedtls_rsa_context has the padding mode
- *                        #MBEDTLS_RSA_PKCS_V15:
+ *                        context uses the PKCS#1 v1.5 padding mode:
```
mpg, line 479: Minor nit on commit structure (no need to rewrite history now, just a though for future PR): as a reviewer, I would prefer for a commit to contain either manual changes (such as this one) or "mechanical" changes (such as substituting `PSA_ERROR_INVALID_SIGNATURE` etc below), rather than a mixture of both. No biggie, just a thought for next time.

include/mbedtls/pk.h
```diff
@@ -968,9 +952,6 @@ int mbedtls_pk_parse_key(mbedtls_pk_context *ctx,
  *                  with mbedtls_pk_init() or reset with mbedtls_pk_free(). If you need a
  *                  specific key type, check the result with mbedtls_pk_can_do().
  *
- * \note            For compressed points, see #MBEDTLS_ECP_PF_COMPRESSED for
```
mpg, line 971: Note: I was going to say in this case we should copy the relevant information here: parsing compressed points on curve SECP224R1 and SECP224K1 is not supported, but OTOH we're planning to remove support for those curves in 1.0, and one of them is already not supported in PSA, so we probably don't need to bother.

mpg (approved):
LGTM, thanks!

bjwtaylor (approved):

PR #297, opened 2025-05-20
Remove encrypt/decrypt from PK

mpg:
## Description

Remove encryption/decryption features from PK, in order to avoid having to spend time on it in #298.

~**Depends on:** https://github.com/Mbed-TLS/mbedtls/pull/10186 for a clean CI.~ (merged)

## PR checklist

- [x] **changelog** provided
- [x] **mbedtls development PR** provided Mbed-TLS/mbedtls#10186
- [x] **mbedtls 3.6 PR**  not required because: 4.0-only removal
- **tests**  not required because: feature removal


felixc-arm (approved):
LGTM :+1:

valeriosetti (approved):
LGTM so approving. I know it depends on https://github.com/Mbed-TLS/mbedtls/pull/10186 so I count on `need-preceding-pr` (and failing CI) to block this one from being merged.
I will re-check and re-approve once the dependency will be merged ;)

mpg:
Made a zero-diff force-push (just added an blank line to the last commit's message). With the new github rules, it doesn't invalidate reviews. So this should be good to go as soon as the CI is happy.

valeriosetti:
While working on https://github.com/Mbed-TLS/TF-PSA-Crypto/issues/298 I found that in `pk.c` there are still functions like `psa_algorithm_for_rsa()` which can potentially return an encryption/decryption algorithm to be associated to a PK wrapped RSA key before importing it into PSA. Since this PR is about removing support for encryption/decryption using PK, should we keep that function as-is, or should it be updated accordingly?

gilles-peskine-arm:
`psa_algorithm_for_rsa` is used by `mbedtls_pk_get_psa_attributes` and there's no plan to change that. We're removing encryption from PK, and `mbedtls_pk_get_psa_attributes` together with `mbedtls_pk_import_into_psa` is how you do encryption in PSA with a key that you constructed via PK. So it's even more useful now.

valeriosetti:
Oh right, I think I missed a detail yesterday while working on my [previous comment](https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/297#issuecomment-2913137150). My reasoning was like: "if I cannot do encryption/decryption with PK, then I should not try to setup a PK context for that kind of operation in the first place, right?". What I missed is that one of the main benefits in having PK nowadays is that it provides ways to parse keys that are not available yet in PSA. Therefore on the practical side I can parse a key with PK, import it into PSA and then do encryption/decryption with PSA API.
Sorry for the noise @gilles-peskine-arm

PR #301, opened 2025-06-03
scripts: driver_templates: call driver init/free

ilie-halip-nxp:
If a driver declares "init" or "free" entry points, these should be called during the PSA initialization/uninitialization.

While here, fix a minor whitespace issue to make the generated code look slightly more readable (the entry_point_name macro was adding a newline between the function name and open paranthesis).

## Description

I'm in the process of writing an opaque driver that is using a secure element for some crypto operations. I found that the driver dispatch code is sometimes not generated properly to direct requests to my driver. As an example, the [PSA driver interface](https://github.com/ilie-halip-nxp/TF-PSA-Crypto/blob/development/docs/proposed/psa-driver-interface.md#driver-initialization) mentions that if a driver declares an "init" entry point, it would get called during initialization - but this is not the case, as the `psa_driver_wrapper_init()` is hardcoded to only initialize the test drivers.

There are other such cases where the driver wrapper function doesn't have the dispatch code (`psa_driver_wrapper_mac_compute()`, `psa_driver_wrapper_aead_encrypt()` are only 2 other such examples), but I wanted to start with this simple init/free PR for some feedback, I can submit other PRs for those functions as well.

Notably, for init/free I'm not using the `OS-template-opaque.jinja` common template because in init/free there's no switch for the key location.

## PR checklist

- [ ] **changelog** provided
- [ ] **mbedtls development PR** not required because: change is internal to TF-PSA-Crypto
- [ ] **mbedtls 3.6 PR** not required because: this is a slight behavior change
- [ ] **tests**  not required because: test drivers are already initialized/freed

### DIFF
--- a/scripts/data_files/driver_templates/psa_crypto_driver_wrappers.h.jinja
+++ b/scripts/data_files/driver_templates/psa_crypto_driver_wrappers.h.jinja
@@ -56,7 +56,7 @@
     {{ capability.names[entry_point]}}
     {% else -%}
     {{driver.prefix}}_{{driver.type}}_{{entry_point}}
-    {% endif -%}
+    {%- endif -%}
 {% endmacro %}
 /* END-Common Macro definitions */
 
@@ -74,6 +74,18 @@ static inline psa_status_t psa_driver_wrapper_init( void )
         return( status );
 #endif
 
+{% with entry_point = "init" -%}
+{% for driver in drivers -%}
+{% for capability in driver.capabilities if entry_point in capability.entry_points -%}
+#if ({% if capability['mbedtls/c_condition'] is defined -%}{{ capability['mbedtls/c_condition'] }} {% else -%} {{ 1 }} {% endif %})
+    status = {{ entry_point_name(capability, entry_point, driver) }}( );
+    if( status != PSA_SUCCESS )
+        return( status );
+#endif
+{% endfor -%}
+{% endfor -%}
+{% endwith %}
+
     (void) status;
     return( PSA_SUCCESS );
 }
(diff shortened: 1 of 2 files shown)


scripts/data_files/driver_templates/psa_crypto_driver_wrappers.h.jinja
```diff
@@ -56,7 +56,7 @@
     {{ capability.names[entry_point]}}
     {% else -%}
     {{driver.prefix}}_{{driver.type}}_{{entry_point}}
-    {% endif -%}
+    {%- endif -%}
```
gilles-peskine-arm, line 59: This changes e.g.
```
#if (defined(PSA_CRYPTO_DRIVER_TEST) )
            status = mbedtls_test_transparent_import_key
                (attributes,
                                data,
                                data_length,
```
to
```
#if (defined(PSA_CRYPTO_DRIVER_TEST) )
            status = mbedtls_test_transparent_import_key(attributes,
                                data,
                                data_length,
```

It isn't our normal code style either way, but I don't mind the incremental improvement.

However, if you make a backport to the 3.6 long-term support branch, please don't change the presentation there. In the LTS branch, we try to minimize upgrade pain, including the pain for people who patch the generated file manually and whose patch would no longer apply cleanly.

gilles-peskine-arm (approved):
LGTM, with the caveat that this is untested (we currently don't have the infrastructure to properly test driver wrapper generation).

If you want we can also add this to the 3.6 LTS branch. We generally do accept driver integration improvements in 3.6 if they aren't disruptive.

ilie-halip-nxp:
I'm preparing similar changes for other functions that don't have the proper driver dispatch code generated.

Any comments/suggestions from other reviewers?

bjwtaylor (approved):

ilie-halip-nxp:
Not needed any more, can be closed.

Issue #282, opened 2025-05-06
Don't require MBEDTLS_PLATFORM_C for entropy


gilles-peskine-arm:
https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/212 causes `MBEDTLS_PLATFORM_C` to be required when using the platform default entropy sources (`MBEDTLS_PLATFORM_GET_ENTROPY_ALT` disabled). It shouldn't: `MBEDTLS_PLATFORM_C` is only supposed to cover runtime configurability of platform functions, not default implementations.

We'd prefer to go ahead and merge https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/212 in this state because it's a disruptive merge (requiring https://github.com/Mbed-TLS/mbedtls/pull/10090 to be merged in order to get a passing CI on crypto), and if we do rework on 212 then we'll have to change our planned merge time.

Fix this in a follow-up by moving the code from `platform.c` to `platform_util.c`. Also remove the requirements on `MBEDTLS_PLATFORM_C` added by https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/212 and https://github.com/Mbed-TLS/mbedtls/pull/10090.

PR #302, opened 2025-06-03
Remove requirment on MBEDTLS_PLATFORM_C for entropy

felixc-arm:
`TF-PSA-Crypto` part of https://github.com/Mbed-TLS/TF-PSA-Crypto/issues/282

Removes the requirement that `MBEDTLS_PLATFORM_C` be enabled to use default entropy sources (i.e. when `MBEDTLS_PLATFORM_GET_ENTROPY_ALT` is disabled), introduced in https://github.com/Mbed-TLS/mbedtls/pull/10090 & https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/212.

Moves the code related to this from `platform.c` to `platform_util.c` so that it is no longer guarded by `#if defined(MBEDTLS_PLATFORM_C)`. Also removes `#define MBEDTLS_PLATFORM_C` from places where it was added due to this. 

## PR checklist

- [x] **changelog** not required because: bugfix caught before release
- [x] **mbedtls development PR** provided: https://github.com/Mbed-TLS/mbedtls/pull/10204
- [x] **mbedtls 3.6 PR** not required because: 4.0/1.0 only bugfix


gilles-peskine-arm (changes requested):
`MBEDTLS_PLATFORM_C` should be taken back out from one more config file.

Issue #291, opened 2025-05-12
Remove X.509 OID stuff from crypto


gilles-peskine-arm:
* Make `mbedtls/crypto_oid.h` an internal header `.../src/crypto_oid.h`.
* Remove the parts of `oid.c` that are not consumed by crypto. Any block of code guarded by `1 /* OID_INFO_STRINGS */` should also be removed. Remove the corresponding unit tests.
* Remove the parts of `crypto_oid.h` that are not consumed by crypto (function declarations, string macros, `MBEDTLS_ERR_OID_BUF_TOO_SMALL`).
* Make `MBEDTLS_ERR_OID_NOT_FOUND` an alias of `PSA_ERROR_NOT_SUPPORTED`.
* Remove the transition file `mbedtls/oid.h`.
* Remove the compilation option `MBEDTLS_OID_C`. Instead, guard each remaining function by whatever guards the place where it's consumed.
* Write a changelog entry for the removal of the OID module from crypto.

This is the crypto side of https://github.com/Mbed-TLS/mbedtls/pull/10125, which you can refer to for inspiration. It's a little simpler because we no longer need to care about cross-module .

Prerequisite: https://github.com/Mbed-TLS/mbedtls/issues/10124 — after that task is done, Mbed TLS no longer consumes those things.

Follow-up: https://github.com/Mbed-TLS/mbedtls/issues/10172

PR #311, opened 2025-06-09
Remove x509 OID code from TF-PSA-Crypto

felixc-arm:
Resolves https://github.com/Mbed-TLS/TF-PSA-Crypto/issues/291.

Removes any OID-related code that is unused in TF-PSA-Crypto.
Essentially removes code that can be found in `mbedtls/library/x509_oid.h`, as this is the x509-related OID code which is not needed in `TF-PSA-Crypto`.

n.b. I didn't change anything in the long list of macros which define all of the OID values at the top of `crypto_oid.h`, but I can see that some could definitely be removed, so if I should change/remove some of these them let me know :slightly_smiling_face: 

## PR checklist

- [ ] **changelog** provided
- [ ] **mbedtls development PR** not required because: pre-work done in https://github.com/Mbed-TLS/mbedtls/pull/10173
- [ ] **mbedtls 3.6 PR** not required because: 4.0/1.0 change only
- **tests**  not required because: removing unused code, no functional changes

### DIFF
--- a/drivers/builtin/src/oid.c
+++ b/drivers/builtin/src/oid.c
@@ -102,6 +102,10 @@
         return MBEDTLS_ERR_OID_NOT_FOUND;                                    \
     }
 
+/* Note: while the data is shared, ideally individual functions that are used
+ * only for writing or only for parsing should depend specifically on that.
+ * See https://github.com/Mbed-TLS/TF-PSA-Crypto/issues/317
+ */
 #if defined(MBEDTLS_PK_PARSE_C) || defined(MBEDTLS_PK_WRITE_C)
 /*
  * For PublicKeyInfo (PKCS1, RFC 5480)
@@ -267,6 +271,10 @@ FN_OID_GET_OID_BY_ATTR1(mbedtls_oid_get_oid_by_ec_grp_algid,
 #endif /* PSA_WANT_KEY_TYPE_ECC_PUBLIC_KEY */
 #endif /* MBEDTLS_PK_PARSE_C || MBEDTLS_PK_WRITE_C */
 
+/*
+ * Note: the optimal dependency would also include CIPHER_C, see
+ * https://github.com/Mbed-TLS/TF-PSA-Crypto/issues/317
+ */
 #if defined(MBEDTLS_PKCS5_C) && defined(MBEDTLS_ASN1_PARSE_C)
 /*
  * For HMAC digestAlgorithm


ChangeLog.d/remove-x509-oid.txt
```diff
@@ -0,0 +1,4 @@
+Removals
+   * The OID module has been removed, along with the compilation option
+     MBEDTLS_OID_C. There are no longer any interfaces to look up values by
+     OID or OID enum values.
```
gilles-peskine-arm, line 4: As worded, the first sentence suggests that TF-PSA-Crypto can no longer parse anything that contains an OID. The OID module still exists in the implementation, but there's no longer a direct interface to it.

The OID module offered four different things:

* `#include <mbedtls/oid.h>` — removed.
* Compilation option `MBEDTLS_OID_C` — removed, now implicitly enabled as needed.
* Functions `mbedtls_oid_xxx()` (if the header is included and the compilation option is enabled) — removed.
* Functionality of being able to parse things with OIDs — kept (for things that are relevant to the crypto libray; things that are only relevant to X.509 are now invisible).

You can see [how I worded it in mbedtls](https://github.com/Mbed-TLS/mbedtls/blob/development/ChangeLog.d/oid.txt) (a little different there because mbedtls still has `oid.h`, with functions for OID representation conversion that are only relevant for X.509).

drivers/builtin/src/oid.c
```diff
@@ -9,9 +9,7 @@
 
 #include "common.h"
```
gilles-peskine-arm, line 11: Now all OID tables are always present in the build. That's a waste of code size. We should guard each function (and the table it consumes) according to where it's used.

For example, `oid_pk_alg` is used for `mbedtls_oid_get_pk_alg()` and `mbedtls_oid_get_oid_by_pk_alg()`. `mbedtls_oid_get_pk_alg()` is only used in `pkparse.c`, guarded by `defined(MBEDTLS_PK_PARSE_C)`, so its definition should also be guarded by `defined(MBEDTLS_PK_PARSE_C)`. Likewise `mbedtls_oid_get_oid_by_pk_alg()` in `pkwrite.c` guarded by `defined(MBEDTLS_PK_WRITE_C)`. And as a consequence the table `oid_pk_alg` should be guarded by `defined(MBEDTLS_PK_PARSE_C) || defined(MBEDTLS_PK_WRITE_C)`.

We can resolve this after 1.0, but I'd prefer to make a reasonable effort now.

drivers/builtin/src/crypto_oid.h
```diff
@@ -28,31 +28,7 @@
  * the public API, Hide it from scripts/generate_errors.pl, which
  * is in the mbedtls repository, and would complain because it doesn't
  * recognize the header name "crypto_oid.h". */
```
gilles-peskine-arm, line 30: This comment is obsolete since the definition of `MBEDTLS_ERR_OID_NOT_FOUND` is no longer visible to `generate_errors.pl`.

drivers/builtin/src/crypto_oid.h
```diff
@@ -28,31 +28,7 @@
...
-#define MBEDTLS_OID_X509_EXT_SUBJECT_DIRECTORY_ATTRS     (1 << 7)
-#define MBEDTLS_OID_X509_EXT_BASIC_CONSTRAINTS           (1 << 8)
-#define MBEDTLS_OID_X509_EXT_NAME_CONSTRAINTS            (1 << 9)
-#define MBEDTLS_OID_X509_EXT_POLICY_CONSTRAINTS          (1 << 10)
-#define MBEDTLS_OID_X509_EXT_EXTENDED_KEY_USAGE          (1 << 11)
-#define MBEDTLS_OID_X509_EXT_CRL_DISTRIBUTION_POINTS     (1 << 12)
-#define MBEDTLS_OID_X509_EXT_INIHIBIT_ANYPOLICY          (1 << 13)
-#define MBEDTLS_OID_X509_EXT_FRESHEST_CRL                (1 << 14)
-#define MBEDTLS_OID_X509_EXT_NS_CERT_TYPE                (1 << 16)
+#define MBEDTLS_ERR_OID_NOT_FOUND                         PSA_ERROR_NOT_SUPPORTED
```
gilles-peskine-arm, line 32: We should remove unused macros at some point. But since they're internal and don't cost anything other than a tiny amount of extra effort when reading the file, it's not urgent at all.

drivers/builtin/src/oid.c
```diff
@@ -29,13 +27,8 @@
 /*
  * Macro to generate mbedtls_oid_descriptor_t
  */
-#if 1 /* OID_INFO_STRINGS */
-#define OID_DESCRIPTOR(s, name, description)  { ADD_LEN(s), name, description }
-#define NULL_OID_DESCRIPTOR                   { NULL, 0, NULL, NULL }
-#else
 #define OID_DESCRIPTOR(s, name, description)  { ADD_LEN(s) }
```
gilles-peskine-arm, line 30: Now the name and description are always unused. We might want to clean this up later. There's no need to do it now, and arguably having the name is a bit of documentation that we might want to keep. But please add a note saying that the name and description are present for historical reasons and no longer used.

gilles-peskine-arm (changes requested):
LGTM except that the OID tables and functions should have compilation guards, and the changelog entry.

felixc-arm:
Rebase to fix unrelated CI failure.

davidhorstmann-arm (commented):
No further comments beyond what Gilles has raised

drivers/builtin/src/crypto_oid.h
mpg: No action required here then, I think?

gilles-peskine-arm (approved):
LGTM

Issue #229, opened 2025-03-23
Move Everest headers to a private subdirectory


gilles-peskine-arm:
Move headers from `drivers/everest/include/everest` to `drivers/everest/include/tf-psa-crypto/private/everest`. Adjust build scripts accordingly.

The goal is that you have to type `/private/` when `#include`ing a private header. See https://github.com/gilles-peskine-arm/TF-PSA-Crypto/blob/7cce26bc7993b7df39d9a43db75d02c4dde86511/docs/architecture/0e-plans.md#privatization-of-everest-headers for more context.

Follow up: https://github.com/Mbed-TLS/mbedtls/issues/10226

gilles-peskine-arm:
@bjwtaylor Note that I'm not sure if we're actually going ahead with this task? It's the Everest part of https://github.com/Mbed-TLS/mbedtls/issues/10087 which has been demoted, so presumably it should be demoted too. @mpg ?

mpg:
Let's go the opposite way and promote 10087 again - I don't want to spend my time arguing against doing something that I agree is the right thing to do in principle, especially when part of it have been started already.

PR #319, opened 2025-06-16
[tf-psa-crypto] Move Everest headers to a private subdirectory (2/2)

valeriosetti:
## Description

- Resolves https://github.com/Mbed-TLS/TF-PSA-Crypto/issues/229.
- PR 2 out of 2
- Supersedes https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/235

## PR checklist

- [x] **changelog** not required because: nothing visible to the end user
- [x] **mbedtls development PR** provided https://github.com/Mbed-TLS/mbedtls/pull/10225
- [x] **mbedtls 3.6 PR** not required because: only development
- **tests** not required because: no change in the code


drivers/everest/CMakeLists.txt
```diff
@@ -6,20 +6,20 @@ add_library(${everest_target}
   library/Hacl_Curve25519_joined.c)
 
 set_base_compile_options(${everest_target})
-target_include_directories(${everest_target}
-  PUBLIC $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>
-         $<BUILD_INTERFACE:${TF_PSA_CRYPTO_DIR}/include>
-         $<BUILD_INTERFACE:${TF_PSA_CRYPTO_DIR}/drivers/builtin/include>
+  target_include_directories(${everest_target}
+  PUBLIC $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include/tf-psa-crypto/private>
```
gilles-peskine-arm, line 10: This doesn't look right. The point of this task is so that e.g. `#include "everest/everest.h"` no longer works, and instead you need `#include "tf-psa-crypto/private/everest/everest.h"`. So `private` should not be part of any directory on the include list. But `ecdh.h` should change.

gilles-peskine-arm (changes requested):
The goal of this task is that if you include Everest headers, you need to write `#include ".../private/..."`, to make it clear that you're using a non-public interface. Adding a subdirectory of `.../private` to a public include path defeats that.

I'm not sure about private include paths, when building Everest itself: there it's ok to put `.../private` on the include path, but it's probably simpler overall to use the same include path as in public and change `everest.c` to `#include ".../private/..."` as well.

drivers/everest/CMakeLists.txt
valeriosetti: Oh I see. This was not something mentioned in the [original issue description](https://github.com/Mbed-TLS/TF-PSA-Crypto/issues/229) and I had to dig trough #235 [comments](https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/235#discussion_r2073050303) to find a mention to it. Plus it's also market with the :rocket: so my thought was that everything was already addressed and I only had to take care of reorganizing PRs.
Apparently this is not the case and there's more changes required. I will try to address this as well

valeriosetti: I should have addressed this part now.

gilles-peskine-arm (approved):
LGTM at d2bf33cb1ef91b8acb3bf094f7fa220b5a80d6b8 for content

gilles-peskine-arm:
DCO is unhappy because a revert commit is missing a signoff line.

I think it would make sense to simplify this PR's history. The final patch is pretty small for 23 commits!

mpg (approved):
LGTM, thanks!

mpg:
Aw, there are conflicts with the framework pointer. I suggest we wait until https://github.com/Mbed-TLS/mbedtls/pull/10225 has gone through the merge queue, then update this one to resolve the conflict - that way we'll also get a new CI run based on the new mbed TLS development including m10225, which will hopefully make it green.

mpg:
@valeriosetti The Mbed TLS PR has been merged, so now would be  good time to resolve conflicts here.

mpg (approved):
LGTM, thanks!

PR #321, opened 2025-06-19
Remove DSA constants

### LINKED ISSUE #7990: Remove mentions of DSA from Mbed TLS
The PSA headers and tests mention DSA (as in, finite-field DSA — not elliptic curves algorithms like ECDSA and EdDSA). In 2023 I can confidently state that we are not going to support it in Mbed TLS. Furthermore, supporting DSA in the PSA crypto API would require extensions to the API and not just an algorithm encoding, because DSA has unusual characteristics: it has partially independent sizes for the private key and the public key so key generation needs two bit-sizes and not just one; and it requires domain parameters for which we no longer plan to add an API (and in Mbed TLS [we are going to remove the partial implementation of domain parameters](https://github.com/Mbed-TLS/mbedtls/issues/6495)).

The goal of this task is to remove the definitions of key types and algorithms for DSA as well as related macros and tests. However note these limitations:

* Technically, it's an API break to remove the publicly documented macros, so we may need to keep them until 4.0. We can freeze the macros and remove associated tests though (as long as we don't run into some test completeness issue). We can also remove the handling of DSA in `crypto_sizes.h`.
* The DSA definitions are just as useless in 2.28 as in 3.x. So we may want to do the same thing in 2.28 and 3.x. If we don't modify 2.28, we may want to keep the code to skip DSA in `generate_psa_tests.py`.

gilles-peskine-arm:
Remove constants for DSA (the old, non-ECC kind). It was never implemented in Mbed TLS, and was not included in the standard PSA Crypto API.

We should do this before 1.0, to avoid any doubt about DSA possibly making its way into TF-PSA-Crypto 1.x.

Resolves https://github.com/Mbed-TLS/mbedtls/issues/7990

## PR checklist

- [x] **changelog** not required because: removing hints of a feature that was never implemented
- [x] **framework PR** not required (the framework does have code to handle DSA constant for `psa_constant_names`, but it's still relevant to 3.6)
- **tests**  provided


valeriosetti (dismissed):

bjwtaylor (dismissed):

gilles-peskine-arm:
We missed this in 1.0.0. Is it still ok to merge? Technically it removes some public, documented macros, so it's an API break. But they were useless macros.

bjwtaylor (dismissed):
I think the CI is still moaning, however otherwise looks good.

valeriosetti (approved):

gilles-peskine-arm:
I rebased on top of `development`. Let's try to get this at least in 1.1 LTS. [Save a bit of code size in applications](https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/687#discussion_r2797781069).

valeriosetti:
I restarted the CI because on the previous run I failed while setting up

gilles-peskine-arm:
Subsumed by https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/824.

gilles-peskine-arm:
Subsumed by https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/824.

PR #325, opened 2025-06-20
Privatize content in Everest headers

### LINKED ISSUE #230: Privatize content in Everest headers
Make functions in asymmetric cryptography headers private, keeping types exposed. See https://github.com/Mbed-TLS/TF-PSA-Crypto/issues/219#issuecomment-2742932065

Prerequisites: https://github.com/Mbed-TLS/mbedtls/issues/10082, https://github.com/Mbed-TLS/TF-PSA-Crypto/issues/219

TODO: do we actually need this? Everest types only appear via ECDH contexts. But ECDH contexts are only exposed for interruptible key exchange (aka restartable ECDH), and Everest doesn't do restartable. It is possible to have both restartable ECC and Everest enabled in the same build, however. The type `mbedtls_ecdh_context_everest` appears in the unnamed union `mbedtls_ecdh_context::ctx`, even though this field of the union is never used when `mbedtls_ecdh_context` is used restartably. Is it easier to sort this out now, or to selectively privatize parts of Everest?

Depends on https://github.com/Mbed-TLS/TF-PSA-Crypto/issues/229

valeriosetti:
## Description

Resolves #230.

## PR checklist

- [x] **changelog** not required because: these functions were already meant to be private and we're only enforcing it here, so nothing changes from the end user point of view.
- **tests** not required because: no code change

### DIFF
--- a/drivers/everest/include/tf-psa-crypto/private/everest/Hacl_Curve25519.h
+++ b/drivers/everest/include/tf-psa-crypto/private/everest/Hacl_Curve25519.h
@@ -15,7 +15,11 @@
 
 #include "kremlib.h"
 
+#if defined(MBEDTLS_DECLARE_PRIVATE_IDENTIFIERS)
+
 void Hacl_Curve25519_crypto_scalarmult(uint8_t *mypublic, uint8_t *secret, uint8_t *basepoint);
+
+#endif /* MBEDTLS_DECLARE_PRIVATE_IDENTIFIERS */
 
 #define __Hacl_Curve25519_H_DEFINED
 #endif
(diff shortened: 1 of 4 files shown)


felixc-arm (approved):
LGTM given we're not too bothered about privatizing macros/types in Everest: https://github.com/Mbed-TLS/TF-PSA-Crypto/issues/230#issuecomment-2857903311

One thing that's not really related - I see there's a `Hacl_Curve25519.h` in `private/everest/vs2013` that seems identical (apart from the comment on how it was generated) to the now privatized one in `private/everest`. The only place I can see this file being used is in `mbedtls/scripts/generate_visualc_files.pl`, but given that in this file the directory `private/everest` is included before `private/everest/vs2013` I don't think `vs2013/Hacl_Curve25519.h` is ever used (as `everest/Hacl_Curves25519.h` will be found first), so could possibly be removed? Either way doesn't really affect this PR.

gilles-peskine-arm:
The merge ban has been lifted. Merging.

Issue #323, opened 2025-06-20
TF-PSA-Crypto version macro


gilles-peskine-arm:
TF-PSA-Crypto should define version number macros, similar to what Mbed TLS has as `MBEDTLS_VERSION_MAJOR`, `MBEDTLS_VERSION_MINOR`, `MBEDTLS_VERSION_PATCH`. We should also provide strings similar to `MBEDTLS_VERSION_NUMBER`, `MBEDTLS_VERSION_STRING`, `MBEDTLS_VERSION_STRING_FULL`.

TBD: what header provides these macros? In Mbed TLS, the macros are defined in `mbedlts/build_info.h` so that they are always available, but the historical interface and the one that's still kind of intended is `mbedtls/version.h`. I think we should make it official that the TF-PSA-Crypto macros are exposed through `#include <tf-psa-crypto/version.h>`, regardless of where they are defined (presumably `tf-psa-crypto/build_info.h`).

(Note that this is different from `PSA_CRYPTO_API_VERSION_MAJOR` and `PSA_CRYPTO_API_VERSION_MINOR`, which indicate support for API versions.)

Related, but out of scope here: [how projects can know the version of Mbed TLS **or** TF-PSA-Crypto](https://github.com/Mbed-TLS/mbedtls/issues/10022), if they want crypto and support both Mbed TLS 3.6 and TF-PSA-Crypto 1.x.

Actions here:

* Define the macros.
* Document the macros and how to access them.

Follow-ups:

* Make sure that the release process includes updating the TF-PSA-Crypto version.

PR #367, opened 2025-07-08
Add TF-PSA-Crypto version macros and functions

davidhorstmann-arm:
Introduce new version macros for TF-PSA-Crypto, analogous to the ones in Mbed TLS. Introduce new convenience functions as well to access the values of these macros (although they may also be accessed directly).

## PR checklist

- [x] **changelog** provided
- [x] **mbedtls development PR** not required because: Changes are only for TF-PSA-Crypto
- [x] **mbedtls 3.6 PR** not required because: Changes are only for TF-PSA-Crypto 1.x
- **tests**  provided

### DIFF
--- a/core/tf_psa_crypto_version.c
+++ b/core/tf_psa_crypto_version.c
@@ -8,6 +8,8 @@
 #include "tf-psa-crypto/version.h"
 #include <string.h>
 
+#if defined(TF_PSA_CRYPTO_VERSION)
+
 unsigned int tf_psa_crypto_version_get_number(void)
 {
     return TF_PSA_CRYPTO_VERSION_NUMBER;
@@ -24,3 +26,5 @@ void tf_psa_crypto_version_get_string_full(char *string)
     memcpy(string, TF_PSA_CRYPTO_VERSION_STRING_FULL,
            sizeof(TF_PSA_CRYPTO_VERSION_STRING_FULL));
 }
+
+#endif /* TF_PSA_CRYPTO_VERSION */
--- a/core/tf_psa_crypto_version.c
+++ b/core/tf_psa_crypto_version.c
@@ -16,16 +16,14 @@ unsigned int tf_psa_crypto_version_get_number(void)
     return TF_PSA_CRYPTO_VERSION_NUMBER;
 }
 
-void tf_psa_crypto_version_get_string(char *string)
+const char *tf_psa_crypto_version_get_string(void)
 {
-    memcpy(string, TF_PSA_CRYPTO_VERSION_STRING,
-           sizeof(TF_PSA_CRYPTO_VERSION_STRING));
+    return TF_PSA_CRYPTO_VERSION_STRING;
 }
 
-void tf_psa_crypto_version_get_string_full(char *string)
+const char *tf_psa_crypto_version_get_string_full(void)
 {
-    memcpy(string, TF_PSA_CRYPTO_VERSION_STRING_FULL,
-           sizeof(TF_PSA_CRYPTO_VERSION_STRING_FULL));
+    return TF_PSA_CRYPTO_VERSION_STRING_FULL;
 }
 
 #endif /* TF_PSA_CRYPTO_VERSION */
--- a/core/tf_psa_crypto_version.c
+++ b/core/tf_psa_crypto_version.c
@@ -9,7 +9,6 @@
 
 #if defined(TF_PSA_CRYPTO_VERSION)
 #include "tf-psa-crypto/version.h"
-#include <string.h>
 
 unsigned int tf_psa_crypto_version_get_number(void)
 {
--- a/core/tf_psa_crypto_version.c
+++ b/core/tf_psa_crypto_version.c
@@ -1,7 +1,7 @@
 /*
  *  Version information
  *
- *  Copyright The TF-PSA-Crypto Contributors
+ *  Copyright The Mbed TLS Contributors
  *  SPDX-License-Identifier: Apache-2.0 OR GPL-2.0-or-later
  */
 
--- a/core/tf_psa_crypto_version.c
+++ b/core/tf_psa_crypto_version.c
@@ -5,6 +5,7 @@
  *  SPDX-License-Identifier: Apache-2.0 OR GPL-2.0-or-later
  */
 
+#include "tf-psa-crypto/build_info.h"
 #include "tf-psa-crypto/version.h"
 #include <string.h>
 
--- a/core/tf_psa_crypto_version.c
+++ b/core/tf_psa_crypto_version.c
@@ -5,7 +5,7 @@
  *  SPDX-License-Identifier: Apache-2.0 OR GPL-2.0-or-later
  */
 
-#include "tf-psa-crypto/tf_psa_crypto_version.h"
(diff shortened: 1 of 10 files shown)


include/tf-psa-crypto/version.h
```diff
@@ -0,0 +1,66 @@
...
+#define TF_PSA_CRYPTO_VERSION_PATCH  0
+
+/**
+ * The single version number has the following structure:
+ *    MMNNPP00
+ *    Major version | Minor version | Patch version
+ */
+#define TF_PSA_CRYPTO_VERSION_NUMBER        0x01000000
+#define TF_PSA_CRYPTO_VERSION_STRING        "1.0.0"
+#define TF_PSA_CRYPTO_VERSION_STRING_FULL   "TF-PSA-Crypto 1.0.0"
```
ronald-cron-arm, line 33: As I said, I would move that to `tf-psa-crypto/build_info.h`. Otherwise, the CI is not happy with "Copyright The TF-PSA-Crypto Contributors".

davidhorstmann-arm: Ah sorry I missed that. I was going to say - [from the issue](https://github.com/Mbed-TLS/TF-PSA-Crypto/issues/323#issue-3162741076):

> I think we should make it official that the TF-PSA-Crypto macros are exposed through #include <tf-psa-crypto/version.h>, regardless of where they are defined (presumably tf-psa-crypto/build_info.h).

I can't see how to do this except by only defining the macros in `version.h`. If it's okay to have them available everywhere then I will move them to `build_info.h`.

davidhorstmann-arm: > Otherwise, the CI is not happy with "Copyright The TF-PSA-Crypto Contributors".

Ah I did not realise that we did copyright checks. I'll update that.

gilles-peskine-arm: > the CI is not happy with "Copyright The TF-PSA-Crypto Contributors"

As a derivative of Mbed TLS, TF-PSA-Crypto remains encumbered by its copyright. The copyright owners of TF-PSA-Crypto are, collectively:

* Arm;
* Everyone who contributed to Mbed TLS or its predecessor project under a legal agreement that transferred their intellectual property to Arm (CLA);
* Everyone who has contributed to Mbed TLS since we dropped the CLA requirement, and until TF-PSA-Crypto forked;
* Everyone who has contributed directly to TF-PSA-Crypto.

We can put “Copyright The TF-PSA-Crypto Contributors” on brand new files, but we have to be careful, for example if we copy code from other files then we'd have to change the line to also mention the Mbed TLS contributors. So I think we should avoid doing that for now.

davidhorstmann-arm:
It looks like `check_names.py` does not always respond to `//no-check-names` exceptions, so I'll have to fix this on the framework side. I'll also try to fix the use of `//no-check-names` as well, so it works in the future.

include/tf-psa-crypto/version.h
```diff
@@ -0,0 +1,51 @@
+/**
+ * \file tf-psa-crypto/version.h
```
ronald-cron-arm, line 2: This does not seem to be enough to appease `check_doxygen_warning()`. Looking at `build_info.h` and `check_config.h` it is likely that in `include/mbedtls/version.h` we should have:
```
 * \file mbedtls/version.h
```

davidhorstmann-arm:
(Force-pushed to get one framework-update commit. I initially updated it to point to the head of Mbed-TLS/mbedtls-framework#185 but then that PR was merged so I updated it to point to the tip of `main`)

ronald-cron-arm (dismissed):
LGTM

core/tf_psa_crypto_version.c
```diff
@@ -0,0 +1,31 @@
...
+{
+    memcpy(string, TF_PSA_CRYPTO_VERSION_STRING,
+           sizeof(TF_PSA_CRYPTO_VERSION_STRING));
+}
+
+void tf_psa_crypto_version_get_string_full(char *string)
+{
+    memcpy(string, TF_PSA_CRYPTO_VERSION_STRING_FULL,
+           sizeof(TF_PSA_CRYPTO_VERSION_STRING_FULL));
+}
```
bensze01, line 29: I know that this is the interface we use in Mbed TLS, but is this really the best way to go about it?
Our users have no way of knowing how large the buffer needs to be to contain our version strings, aside from making educated guesses.

String literals have a static storage duration in both [C](https://en.cppreference.com/w/c/language/string_literal.html) and [C++](https://en.cppreference.com/w/cpp/language/string_literal.html), so returning a pointer to a string literal should be fine when compiled with any compiler of either language.

```suggestion
char *tf_psa_crypto_version_get_string(void)
{
    return TF_PSA_CRYPTO_VERSION_STRING;
}

char *tf_psa_crypto_version_get_string_full(void)
{
    return TF_PSA_CRYPTO_VERSION_STRING_FULL;
}
```

@gilles-peskine-arm Thoughts?

bensze01 (changes requested):

core/tf_psa_crypto_version.c
gilles-peskine-arm: I agree. Having a fixed-size string limits us. We can't rename the product. We only have room for a certain number of digits, no way to add a suffix like `beta` and we should be careful with numbers. I'm all for just returning a pointer to a global constant.

gilles-peskine-arm: I've even [added it to the 4.0 shoulds](https://github.com/Mbed-TLS/mbedtls/issues/10308). It's not very important, but it's very small.

gilles-peskine-arm: > for the sake of consistency

I don't see any need for consistency here. If there was a `get_version_of_crypto_library()` function that already existed in Mbed TLS then TF-PSA-Crypto should maintain backward compatibility with that, to allow applications to work with either product. But there wasn't, there was only “get version of Mbed TLS” and TF-PSA-Crypto doesn't offer a version of Mbed TLS. So “get version of the crypto library” is a new thing anyway.

gilles-peskine-arm: Yes, I understood that, sorry for being unclear. My point is that the only benefit I see for consistency is to simplify things for callers who want to get the crypto version in code that needs to work with both 3.6 and 1.0, and having similar prototypes with just different function names and buffer sizes doesn't really simplify that. If anything, callers might just change the function name and forget to make the buffer large enough for `"TF-PSA-Crypto 1.23.4\000"`.

bensze01: Since David is on Leave now, I would be willing to take this one over, and raise an equivalent PR updating to the version functions in 4.0.

ronald-cron-arm: David is still there today.

bensze01: Ah, ok. My bad.

davidhorstmann-arm: That's a really good point, thanks!

bensze01 (approved):
LGTM!

ronald-cron-arm (approved):
LGTM, thanks.

PR #388, opened 2025-07-23
[tf-psa-crypto] Define MBEDTLS_PK_ALG_ECDSA (2/2)

### LINKED ISSUE #345: Define MBEDTLS_PK_ALG_ECDSA
Define and document `MBEDTLS_PK_ALG_ECDSA` as the default variant of ECDSA used by PK:

```
#if defined(MBEDTLS_ECDSA_DETERMINISTIC)
#define MBEDTLS_PK_ALG_ECDSA(hash_alg) PSA_ALG_DETERMINISTIC_ECDSA(hash_alg)
#else
#define MBEDTLS_PK_ALG_ECDSA(hash_alg) PSA_ALG_ECDSA(hash_alg)
#endif
```

This is equivalent to the internal `MBEDTLS_PK_PSA_ALG_ECDSA_MAYBE_DET`. Replace it by `MBEDTLS_PK_ALG_ECDSA` and remove its definition.

Also update `mbedtls_pk_get_psa_attributes` to use `MBEDTLS_PK_ALG_ECDSA`.

valeriosetti:
## Description

Resolves #345.
This is PR 2 out of 2 of this series

## PR checklist

- [ ] **changelog** provided: done
- [ ] **mbedtls development PR** provided: https://github.com/Mbed-TLS/mbedtls/pull/10329
- **tests** not required because: no functional code change

### DIFF
--- a/include/mbedtls/pk.h
+++ b/include/mbedtls/pk.h
@@ -565,9 +565,7 @@ int mbedtls_pk_can_do_ext(const mbedtls_pk_context *ctx, psa_algorithm_t alg,
  *                        if \p usage is ENCRYPT/DECRYPT.
  *                      - #MBEDTLS_PK_ECDSA or #MBEDTLS_PK_ECKEY
  *                        if \p usage is SIGN/VERIFY:
- *                        #PSA_ALG_DETERMINISTIC_ECDSA(#PSA_ALG_ANY_HASH)
- *                        if #MBEDTLS_PK_ALG_ECDSA is set,
- *                        otherwise #PSA_ALG_ECDSA(#PSA_ALG_ANY_HASH).
+ *                        #MBEDTLS_PK_ALG_ECDSA.
  *                      - #MBEDTLS_PK_ECKEY_DH or #MBEDTLS_PK_ECKEY
  *                        if \p usage is DERIVE:
  *                        #PSA_ALG_ECDH.


ChangeLog.d/issue345.txt
```diff
@@ -0,0 +1,4 @@
+Features
+   * Helper symbol MBEDTLS_PK_ALG_ECDSA is added to the public "mbedtls/pk.h"
+     header file. It allows to automatically select the proper ECDSA algorithm,
```
felixc-arm, line 3: Tiny grammar point - missing something between 'allows' and 'to', e.g.
```suggestion
     header file. It allows the automatic selection of the proper ECDSA algorithm,
```

felixc-arm (commented):
LGTM apart from small grammar point :+1:

gilles-peskine-arm:
There's a merge conflict

felixc-arm (approved):
LGTM, thanks!

docs/architecture/pk-4.md
```diff
@@ -264,11 +264,9 @@ There is currently no way to access the underlying PSA key of a PK context. A nw
 
 PK sometimes needs to choose between ECDSA variants, when it builds PSA attributes for an ECC key:
 
-* To import an ECC key during parsing, and to sign with a built-in ECC key, when `MBEDTLS_PK_USE_PSA_EC_DATA` is enabled. This uses the macro `MBEDTLS_PK_PSA_ALG_ECDSA_MAYBE_DET` defined in `pk_internal.h`.
+* To import an ECC key during parsing, and to sign with a built-in ECC key, when `MBEDTLS_PK_USE_PSA_EC_DATA` is enabled. This uses the macro `MBEDTLS_PK_ALG_ECDSA` defined in `mbedtls/pk.h`.
 * In `mbedtls_pk_get_psa_attributes()` to choose the default policy for an ECC key used for signature.
 
-ACTION (https://github.com/Mbed-TLS/TF-PSA-Crypto/issues/345): change `MBEDTLS_PK_PSA_ALG_ECDSA_MAYBE_DET` to a public macro `MBEDTLS_PK_ALG_ECDSA`. Switch `mbedtls_pk_get_psa_attributes()` to it. This is nice but not critical for 1.0.
```
gilles-peskine-arm, line 270: Thanks for updating this in the spec, I had missed the inconsistency! But please keep the link to the task in the spec — don't remove this paragraph. The document is all the work we're doing for the project, not just the work we haven't done yet.

include/mbedtls/pk.h
```diff
@@ -262,6 +262,16 @@ typedef struct {
 typedef void mbedtls_pk_restart_ctx;
 #endif /* MBEDTLS_ECP_RESTARTABLE */
 
+/**
+ * PK helper that can be used to automatically pick the proper ECDSA algorithm
+ * variant (deterministic VS random) based on #MBEDTLS_ECDSA_DETERMINISTIC.
```
gilles-peskine-arm, line 267: In context, it's unclear what makes a variant “proper”. I would actually say that it's the other way round: `MBEDTLS_PK_ALG_ECDSA` exposes which variant the pk module uses by default.

As for `MBEDTLS_ECDSA_DETERMINISTIC`, I hadn't realized this was still a public macro. I had expected it to be removed from the config since it's supposed to be derived from `PSA_WANT_xxx`. In fact we now have a bug there:

* If the user enables both `PSA_WANT_ALG_DETERMINISTIC_ECDSA` then `MBEDTLS_ECDSA_DETERMINISTIC`, then all is fine.
* If the user enables `PSA_WANT_ALG_DETERMINISTIC_ECDSA` but not `MBEDTLS_ECDSA_DETERMINISTIC`, then `MBEDTLS_ECDSA_DETERMINISTIC` is auto-enabled internally and PK will use deterministic ECDSA, contradicting the documentation here.
* If the user enables `PSA_WANT_ALG_ECDSA` and `MBEDTLS_ECDSA_DETERMINISTIC` but not `PSA_WANT_ALG_DETERMINISTIC_ECDSA`, then PK will try to use deterministic ECDSA and fail.

So can you please:

* Remove `MBEDTLS_ECDSA_DETERMINISTIC` from `crypto_config.h`. It's never set explicitly in `all.sh` (it's unset in some components in mbedtls but that's ok), so I think that won't break the CI. No need for a separate changelog entry because that's all part of the `MBEDTLS_PSA_CRYPTO_CONFIG` always on thing.
* Change the documentation here say that PK currently uses deterministic ECDSA if available and randomized otherwise, but this choice may change in the future. (If we determine that the deterministic variant is weaker against some attack, we may want to make randomized the default.)
* Change the changelog entry for `MBEDTLS_PK_ALG_ECDSA` to just say that this is what PK picks. No need to explain how it picks.

ChangeLog.d/issue345.txt
```diff
@@ -0,0 +1,5 @@
+Features
+   * Helper symbol MBEDTLS_PK_ALG_ECDSA is added to the public "mbedtls/pk.h"
+     header file. It allows the automatic selection of the proper ECDSA
+     algorithm, deterministic or randomized, based on
+     #MBEDTLS_ECDSA_DETERMINISTIC.
```
gilles-peskine-arm, line 5: Not quite, see my comment on the definition.

gilles-peskine-arm (changes requested):
LGTM for the code, but not for the documentation.

I hadn't noticed we forgot a little piece of work of legacy config removal, which is to remove `MBEDTLS_ECDSA_DETERMINISTIC`. Please try it here. If the CI complains too much we might rethink the strategy.

include/mbedtls/pk.h
```diff
@@ -554,7 +566,7 @@ int mbedtls_pk_can_do_ext(const mbedtls_pk_context *ctx, psa_algorithm_t alg,
  *                      - #MBEDTLS_PK_ECDSA or #MBEDTLS_PK_ECKEY
  *                        if \p usage is SIGN/VERIFY:
  *                        #PSA_ALG_DETERMINISTIC_ECDSA(#PSA_ALG_ANY_HASH)
- *                        if #MBEDTLS_ECDSA_DETERMINISTIC is enabled,
+ *                        if #MBEDTLS_PK_ALG_ECDSA is set,
  *                        otherwise #PSA_ALG_ECDSA(#PSA_ALG_ANY_HASH).
```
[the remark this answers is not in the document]
valeriosetti: I apologize for the dumb question, but I don't fully get the point of this comment. What's wrong with the change I made?

felixc-arm (approved):
LGTM assuming CI passes :slightly_smiling_face:

PR #391, opened 2025-07-25
Avoid invalid gcc 14.3 warning about array bounds in mbedtls_xor

keith-packard:
## Description

The combination of the multi-byte loop with the single byte loop confuses GCC 14.3's array bounds checker. When the loop size is constant, check to see if it is a multiple of the multi-byte size and bail early. As this will be evaluated at compile time, there should be no run-time cost.

## PR checklist

- [X] **changelog** provided 
- [X] **mbedtls development PR** provided Mbed-TLS/mbedtls# | not required because: 
- [X] **mbedtls 3.6 PR provided https://github.com/Mbed-TLS/mbedtls/pull/10318
- **tests** not required because: no code generation change


core/tf_psa_crypto_common.h
```diff
@@ -208,6 +208,14 @@ static inline void mbedtls_xor(unsigned char *r,
         return;
     }
 #endif
+#if defined(MBEDTLS_COMPILER_IS_GCC) && __has_builtin(__builtin_constant_p)
```
davidhorstmann-arm, line 211: It looks like the CI is unhappy because `__has_builtin()` was added in [GCC 10](https://gcc.gnu.org/gcc-10/changes.html).

You should be able to use something like:
```c
#if defined(MBEDTLS_COMPILER_IS_GCC) && (__GNUC__ >= 10) && ...
```
To filter for GCC >= 10

keith-packard: Looks like someone got here ahead of me -- `MBEDTLS_HAS_BUILTIN` is defined a bit further down in the file and looks like it will do the trick nicely; it explicitly checks if __has_builtin is available. I had to move that up in this file to make it available where necessary; I tried to move it among similar-looking macros.

valeriosetti (approved):

ChangeLog.d/gcc-14.3.txt
```diff
@@ -0,0 +1,8 @@
+Changes
+   * Move definition of MBEDTLS_HAS_BUILTIN macro earlier in
+     core/tf_psa_crypto_common.h.
+
```
valeriosetti, line 4: That's a minor thing that I don't think it's worth being mentioned in changelog. But that's my opinion though, please wait for other reviewer's feedback before removing it

valeriosetti (changes requested):

ChangeLog.d/gcc-14.3.txt
davidhorstmann-arm: Ah sorry I did not specify, my bad! The ChangeLog-worthy thing here would be something like "Fix warning about array bounds in mbedtls_xor() when building with GCC 14.3". Would you be able to put something like that instead?

valeriosetti (approved):

davidhorstmann-arm (approved):
LGTM, thanks!

PR #392, opened 2025-07-25
Updates to the 0e plans

gilles-peskine-arm:
Updates to the specification for API changes based on the implementation work: change things that need changing, or clarify things that were unclear. I propose to keep accumulating changes here during development.

* Expand on `MBEDTLS_PSA_CRYPTO_RNG_HASH` and how to check it.
* Expand on DES removal.
* Expand on the `md.h` split.
* Tweak the value of `PSA_KEY_USAGE_DERIVE_PUBLIC`.

## PR checklist

- [x] **changelog** not required because: doc only
- [x] **mbedtls development PR** not required because: doc only
- **tests**  not required because: doc only

### DIFF
--- a/docs/architecture/0e-plans.md
+++ b/docs/architecture/0e-plans.md
@@ -1075,7 +1075,7 @@ This section is about the parameters of the DRBG (including how it communicates 
 
 Many of the [RNG options](#impacted-rng-options) are mostly meaningless now that the DRBG modules and the entropy module are no longer public. In this section, we determine which options are still relevant, either because they are algorithm choices or because they are a security/resources compromise.
 
-For notions of strength, we follow [NIST SP 800-90A r1](https://doi.org/10.6028/NIST.SP.800-90Ar1) (especially table 2 p. 38 for HMAC\_DRBG) and table 3 p. 49 for CTR\_DRBG), and [NIST SP 800-57 part 1 r1](https://doi.org/10.6028/NIST.SP.800-57pt1r5) §5.6.1.
+For notions of strength, we follow [NIST SP 800-90A r1](https://doi.org/10.6028/NIST.SP.800-90Ar1) (especially table 2 p. 38 for HMAC\_DRBG) and table 3 p. 49 for CTR\_DRBG), and [NIST SP 800-57 part 1 r5](https://doi.org/10.6028/NIST.SP.800-57pt1r5) §5.6.1.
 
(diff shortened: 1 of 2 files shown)


minosgalanakis (dismissed):
LGTM

docs/architecture/0e-plans.md
```diff
@@ -1283,7 +1283,14 @@ Reasons to do this:
...
+
+* Functions that are currently available when `MBEDTLS_MD_LIGHT` is defined and `MBEDTLS_MD_C` is not defined, will be declared unconditionally in `mbedtls/md.h`.
+* Functions that are currently available only when `MBEDTLS_MD_C`, will now be declared in `mbedtls/md.h` only when `MBEDTLS_DECLARE_PRIVATE_IDENTIFIERS` is defined.
+* `mbedtls_md_engine_t` (not a public interface, but used to define the public type `mbedtls_md_context_t`) should have no Doxygen documentation.
+* `MBEDTLS_ERR_MD_FILE_IO_ERROR` should be guarded by `MBEDTLS_DECLARE_PRIVATE_IDENTIFIERS` since it is only used by functions that are becoming private. (Or we could remove it, together with `mbedtls_md_file()`.)
+* The function `mbedtls_md_setup()` is only partially covered by `MBEDTLS_MD_C`. Specifically, the light configuration omits HMAC support. The whole code needs to stay (it's used by HMAC_DRBG and PKCS5), but the documentation should now state that the `hmac` argument must be 0. (Alternatively, we remove the `hmac` argument and split the function in two.)
+* Nothing changes in `md.c` (unless we go for more extensive work that removes the private features that we already don't need).
```
ronald-cron-arm, line 1293: Thanks for this. This is clearer to me where we want to go. Regarding:
```
The function `mbedtls_md_setup()` is only partially covered by `MBEDTLS_MD_C`. Specifically, the light configuration omits HMAC support. The whole code needs to stay (it's used by HMAC_DRBG and PKCS5), but the documentation should now state that the `hmac` argument must be 0. (Alternatively, we remove the `hmac` argument and split the function in two.)
```
I am in favour of removing the `hmac` argument in the public API, that's cleaner. Looking at the code it seems to me that we have the time to do it. Otherwise we could probably guard the "3.6 MD_C specific code" with MBEDTLS_PKCS5_C and MBEDTLS_HMAC_DRGB_C.

PR #395, opened 2025-07-28
Introduce MBEDTLS_PSA_RNG_RESEED_INTERVAL

minosgalanakis:
## Description

This PR implements the agreed [specification](https://github.com/Mbed-TLS/TF-PSA-Crypto/blob/development/docs/architecture/0e-plans.md)  and it part of a group with

Designed as part of the issue https://github.com/Mbed-TLS/TF-PSA-Crypto/issues/328

## PR checklist

Please remove the segment/s on either side of the | symbol as appropriate, and add any relevant link/s to the end of the line.
If the provided content is part of the present PR remove the # symbol.

- [ ] **changelog** provided | not required because: 
- [ ] **framework PR** provided Mbed-TLS/mbedtls-framework# | not required
- [ ] **mbedtls development PR** provided Mbed-TLS/mbedtls# | not required because: 
- [ ] **mbedtls 3.6 PR** provided Mbed-TLS/mbedtls# | not required because: 
- **tests**  provided | not required because: 

## Notes for the submitter

Please refer to the [contributing guidelines](https://github.com/Mbed-TLS/TF-PSA-Crypto/blob/development/CONTRIBUTING.md), especially the
checklist for PR contributors.

Help make review efficient:
* Multiple simple commits
  - please structure your PR into a series of small commits, each of which does one thing
* Avoid force-push
  - please do not force-push to update your PR - just add new commit(s)
* See our [Guidelines for Contributors](https://mbed-tls.readthedocs.io/en/latest/reviews/review-for-contributors/) for more details about the review process.

Issue #328, opened 2025-06-23
Simplify RNG options for TF-PSA-Crypto 1.0


gilles-peskine-arm:
In TF-PSA-Crypto 1.0, the modules implementing a random generator (entropy, HMAC_DRBG, CTR_DRBG) are now private interfaces. As a consequence, many of their configuration options are now irrelevant and possibly incomprehensible.

The goal of this task is to implement new simpler options, and deduce old options from the new ones.

* Add and document the new options.
* Remove the removed options from `crypto_config.h`. (Note that they remain as macros, just no longer configurable.)
* Where needed, set the old options from the new ones.
* Adapt tests accordingly. 
* Remove warnings about low-strength RNG from makefiles (now, if you set a low strength, it's explicit).
* Write a migration guide explaining how to move from the old option set to the new set. (This may be done in a separate PR.)

The list of options is based on the specification in https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/329.

[New options](https://github.com/Mbed-TLS/TF-PSA-Crypto/blob/development/docs/architecture/0e-plans.md#new-rng-options):

* `MBEDTLS_PSA_CRYPTO_RNG_STRENGTH` indicating the minimum strength of the RNG. Only 128 and 256 are meant to be useful values. Default to 256.
* `MBEDTLS_PSA_CRYPTO_RNG_HASH` indicating which hash algorithm to use for the entropy module, and for HMAC\_DRBG if configured. Default to SHA-256.
* `MBEDTLS_PSA_RNG_RESEED_INTERVAL`: inteeger, becomes the value of `MBEDTLS_CTR_DRBG_RESEED_INTERVAL` and `MBEDTLS_HMAC_DRBG_RESEED_INTERVAL`.

[Removed options](https://github.com/Mbed-TLS/TF-PSA-Crypto/blob/development/docs/architecture/0e-plans.md#removed-drbg-length-options):

* `MBEDTLS_ENTROPY_C`: now an internal detail. Enable it automatically if `MBEDTLS_PSA_CRYPTO_C` is enabled (so not in client-only builds) and `MBEDTLS_PSA_CRYPTO_EXTERNAL_RNG` is disabled.
* `MBEDTLS_CTR_DRBG_USE_128_BIT_KEY`: deduced from `MBEDTLS_PSA_CRYPTO_RNG_STRENGTH`.
[...]

PR #397, opened 2025-07-28
Introduce MBEDTLS_PSA_CRYPTO_RNG_STRENGTH

minosgalanakis:
## Description

This PR implements the agreed [specification](https://github.com/Mbed-TLS/TF-PSA-Crypto/blob/development/docs/architecture/0e-plans.md)  and it part of a group with #396 and #395 

Designed as part of the issue https://github.com/Mbed-TLS/TF-PSA-Crypto/issues/328

## PR checklist

- [ ] **mbedtls development PR** provided Mbed-TLS/mbedtls#10292

## Notes for the submitter

Help make review efficient:
* Multiple simple commits
* Avoid force-push


drivers/builtin/include/mbedtls/config_adjust_legacy_crypto.h
```diff
@@ -378,4 +378,15 @@
 #define MBEDTLS_SSL_HAVE_AEAD
 #endif
 
+/* Deduce MBEDTLS_CTR_DRBG_USE_128_BIT_KEY from MBEDTLS_PSA_CRYPTO_RNG_STRENGTH */
```
gilles-peskine-arm, line 381: Ok, then `MBEDTLS_CTR_DRBG_USE_128_BIT_KEY` should no longer be a configuration option. I guess this needs to wait until https://github.com/Mbed-TLS/mbedtls/pull/10292 is merged?

drivers/builtin/include/mbedtls/config_adjust_legacy_crypto.h
```diff
@@ -378,4 +378,15 @@
...
+#if defined(MBEDTLS_PSA_CRYPTO_RNG_STRENGTH)
+#if (MBEDTLS_PSA_CRYPTO_RNG_STRENGTH == 128)
+#define MBEDTLS_AES_ONLY_128_BIT_KEY
```
gilles-peskine-arm, line 384: No, we can't do that: AES can be used for things other than the RNG! The most obvious case for having 256-bit AES but a 128-bit RNG strength is if you want to have the fastest RNG that the library has to offer (CTR_DRBG with AES-128), but you still want to offer AES-256 for applications that insist on it.

`MBEDTLS_AES_ONLY_128_BIT_KEY` is incompatible with a 256-bit RNG strength _if the RNG is CTR\_DRBG_, but not if the RNG is something else. Checking that would be the job of `check_config.h`, not here.

What we need to set here is not `MBEDTLS_AES_ONLY_128_BIT_KEY` but `MBEDTLS_CTR_DRBG_USE_128_BIT_KEY` (and the logic here is correct for that).

gilles-peskine-arm (changes requested):
Looks good as the first step once `MBEDTLS_AES_ONLY_128_BIT_KEY` is no longer used instead of `MBEDTLS_CTR_DRBG_USE_128_BIT_KEY`.

minosgalanakis:
The comment's from @gilles-peskine-arm have been addressed in #395. This PR will be closed and will be merged with #395

bjwtaylor (approved):

PR #401, opened 2025-07-31
include: pk: use PSA_WANT instead of ECDSA_DETERMINISTIC for PK_ALG_ECDSA

valeriosetti:
## Description

MBEDTLS_ECDSA_DETERMINISTIC is only set when deterministic ECDSA is builtin. When an accelerating driver is present, that legacy symbol is not set and therefore PK will pick the wrong algorithm by default.

## PR checklist

- [ ] **changelog** not required because: no release was done with `MBEDTLS_PK_ALG_ECDSA`
- **tests** not required because: existing tests are  OK


gilles-peskine-arm (approved):
LGTM

felixc-arm (approved):

PR #415, opened 2025-08-08
New threading interface: prototype (crypto)

gilles-peskine-arm:
New threading interface. Resolves https://github.com/Mbed-TLS/mbedtls/issues/8455. Highlights:

* Decouple the platform abstraction from the application interface. This way we can make changes later, such as adding a state field to mutexes and condition variables to make their usage safer.
* Return a status from mutex initialization, since that can fail on some platforms. I ran out of time to implement it. It's conceptually simple but requires a lot of elbow grease if we want to do it right, namely, actually check when it fails. Maybe we could keep `void mbedtls_mutex_init`, and later deprecate it and add `int mbedtls_mutex_setup`.
* Add condition variables, to allow us to do synchronization patterns that we can't currently do (which was a major pain when implementing concurrency for the PSA key store).
* Have thread create/join provided as part of the platform abstraction, rather than require it to be pthread. It's optional if you only want to use the library, but mandatory to run the unit tests.

Status: for design review only. The history is a lot messier than I'd like, and I don't think the CI is nearly ready to pass. Please focus mainly on the documentation (public headers, migration guide).

## PR checklist

- [ ] **changelog** TODO
- [x] **framework PR** provided https://github.com/Mbed-TLS/mbedtls-framework/pull/198
- [ ] **TF-PSA-Crypto PR** https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/415
- [ ] **mbedtls development PR** TODO?
- [ ] **mbedtls 3.6 PR** not required because: no changes in 3.6
- **tests**  provided


gilles-peskine-arm:
Abandoned prototype. It's too ambitious. We'll do something simpler: https://github.com/Mbed-TLS/mbedtls/issues/8455#issuecomment-3234097267

PR #425, opened 2025-08-19
scripts: driver_templates: fix few driver wrappers

ilie-halip-nxp:
## Description

I'm writing an opaque driver that is using a secure element for some crypto operations, notably TLS1.3 and everything related to it. The SE does not allow exporting any keys (except public keys, of course), so all the crypto ops (AEAD, HMAC, HKDF derivation, etc) need to be ran on it using the available key handles.

I found that multiple operations are not being delegated to the driver when they should be. This is apparently due to the fact that the generated code from a few PSA driver wrappers is not correct and doesn't have the generated dispatch code. It looks as though these functions were hardcoded to call the test drivers only. This is essentially the same issue as in PR #301.

## PR checklist

- [ ] **changelog** provided 
- [ ] **mbedtls development PR** not required because: change is internal to TF-PSA-Crypto
- [ ] **mbedtls 3.6 PR** not required because: this is new functionality, backport not discussed yet
- [ ] **tests**  not required because: there's no proper test infrastructure for driver code generation

### DIFF
--- a/ChangeLog.d/driver-fix-few-wrappers.txt
+++ b/ChangeLog.d/driver-fix-few-wrappers.txt
@@ -1,0 +1,13 @@
+Changes
+    * These driver wrappers are fixed and now generate correct dispatch
+      code that calls the proper opaque driver's entry point. This
+      enables writing an opaque driver that can do TLS1.3 without exposing
+      any key.
+      - `get_key_buffer_size`
+      - `get_key_buffer_size_from_key_data`
+      - `generate_key`
+      - `key_agreement`
+      - `mac_sign_setup`, `mac_update`, `mac_sign_finish`, `mac_abort`
+      - `mac_compute`
+      - `aead_decrypt`, `aead_encrypt`
+      - `verify_hash`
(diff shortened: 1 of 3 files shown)


davidhorstmann-arm:
Thanks for this contribution!

This looks on the larger side of size-s. If I remember correctly, there is some dispute about the future of generated driver wrappers, so we need to think a bit about design issues before we can review and merge it.

gilles-peskine-arm:
> there is some dispute about the future of generated driver wrappers

We want to completely rewrite the generation part to use out `psa_wrapper` module instead of the external dependency on Jinja2. But we'll keep the existing JSON parsing code. And it will be a while before we can get around to the rewrite. So in the meantime, we should add driver wrapper code when there's demand for it.

ilie-halip-nxp:
This change is not really needed anymore on our side - is it ok close it?

gilles-peskine-arm:
@ilie-halip-nxp I'm sorry we didn't manage to review this in a timely fashion. We will resolve this in the long term with a script rewrite, but we're willing to take in patches that solve an immediate need — just overloaded with reviews. Feel free to close this if you don't need it, or to leave it open since it is an improvement.

ilie-halip-nxp:
No worries, I'll just go ahead and close this for now. Thanks!

PR #430, opened 2025-08-29
Create threading_internal.h

gilles-peskine-arm:
This is meant to hold threading-related definitions that are not public, but are used in the test framework. Create this in all branches, so that the framework can start doing `#include "threading_internal.h"`.

## PR checklist

- [x] **changelog** provided | not required because: internal
- [x] **framework PR** not required (will follow)
- [x] **mbedtls development PR** not required because: internal
- [x] **mbedtls 3.6 PR** provided Mbed-TLS/mbedtls#10375
- **tests**  not required because: refactoring only

### DIFF
--- a/drivers/builtin/src/threading_internal.h
+++ b/drivers/builtin/src/threading_internal.h
@@ -1,5 +1,5 @@
 /**
- * \file threadinginternal.h
+ * \file threading_internal.h
  *
  * \brief Threading interfaces used by the test framework
  */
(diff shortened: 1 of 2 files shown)


drivers/builtin/src/threading_internal.h
```diff
@@ -0,0 +1,28 @@
+/**
+ * \file threadinginternal.h
```
mpg, line 2: Missing underscore?

mpg (changes requested):
LGTM except for a typo.

davidhorstmann-arm (approved):
Looks good, but it seems a bit strange to have `threading.c` include the threading interfaces used by the test framework and then `threading_internal.h` includes the normal non-testing `threading.h`.

I don't think it's a blocker, but why is this?

PR #437, opened 2025-09-04
Create the 1.0 version of `md.h`

davidhorstmann-arm:
Fixes Mbed-TLS/mbedtls#8450

Move `md.h` to the public include path and make the part that is not guarded by `MBEDTLS_MD_LIGHT` private. Additionally, document that the `hmac` parameter of `mbedtls_md_setup()` must be set to zero as HMAC is no longer supported through MD.

## PR checklist

- [x] **changelog** provided
- [x] **mbedtls development PR** here
- [x] **mbedtls 3.6 PR** not required because: 4.x/1.x changes only
- **tests**  not required because: No new code to test


include/mbedtls/md.h
```diff
@@ -337,6 +339,8 @@ MBEDTLS_CHECK_RETURN_TYPICAL
 int mbedtls_md(const mbedtls_md_info_t *md_info, const unsigned char *input, size_t ilen,
                unsigned char *output);
 
+#if defined(MBEDTLS_DECLARE_PRIVATE_IDENTIFIERS)
```
ronald-cron-arm, line 342: Isn't the case that we should also in `md.c` replace the `#if defined(MBEDTLS_MD_C)` by `#if defined(MBEDTLS_DECLARE_PRIVATE_IDENTIFIERS)`? @mpg wdyt?

gilles-peskine-arm: No: `MBEDTLS_DECLARE_PRIVATE_IDENTIFIERS` guards things that are needed in the build, but not meant to be consumed by applications. These functions are still used in other library modules, possibly also in mbedtls as well, so they must be in the build.

mpg: Specifically regarding `setup()`, when it's called by an application (that doesn't define `PRIVATE_IDENTIFIERS` before including `md.h`) in a build with `MD_C` defined, the application should always be calling it with the `hmac` argument being `0`, but this will no longer be enforced by the code.

I'm not sure that's a big problem though. If the application is really trying to use HMAC, it will get compiler errors when trying to call `mbedtls_md_hmac()` functions. If it's not trying to use HMAC, but it's still passing `hmac == 1`, then (1) the application is doing something stupid and (2) the consequences are only and extra allocation that will never be used. So, I think that's quite acceptable.

gilles-peskine-arm: I guess the task description wasn't very clear if you weren't immersed in the history of md-light. I've made it more explicit in https://github.com/Mbed-TLS/TF-PSA-Crypto/blob/a84c1e70547bcbe5e380235d3e4a86032dab8670/docs/architecture/0e-plans.md#public-hash-only-mdh

gilles-peskine-arm (changes requested):
Mostly LGTM apart from a couple of macros that should be privatized or removed. And I agree that [splitting setup and erroring out on `mbedtls_md_setup(..., 1)`](https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/437#discussion_r2332334018) would be better, but not critical.

gilles-peskine-arm (dismissed):
LGTM

gilles-peskine-arm:
> these changes should be reflected in the .0 migration guide at some point. I'm not saying it has to be part of this PR necessarily, but if it isn't, we should create an issue to track the remaining work.

Last quarter, I wrote migration guide sections for every changelog entry, often combining related changes into one section. I think that's generally a good approach, so we don't need to update the migration guide here, and we don't need a new issue since that's already captured by the changelog entry enumeration method (and by the fact that there's a TODO marker).

mpg:
I take your point, but I'll note that in this case, https://github.com/Mbed-TLS/mbedtls/issues/9946 is not just "consolidate", it's more like "complete contents of several sections from TODO markers and/or ChangeLog entries and consolidate", which seems a bit much to me - I thought we liked small PRs?

mpg:
To clarify: I'm not advocating for having the migration guide in this PR - it's already big enough as it is. I'm also not arguing against your experience that writing a section based on multiple related ChangeLog entries is a good approach. It just seems more natural to me to have separate task for each section that needs writing ("write the MD section", "write the PK section", etc.) and then the final "consolidate" task (once we have all the sections, how do we structure the overall document?).

But that's not something we need to resolve now.

ronald-cron-arm:
> Apologies for the delay, I believe I have addressed everything here.

Just in case, did you address the CI failure?

ronald-cron-arm (approved):
LGTM, thanks.

davidhorstmann-arm:
> Just in case, did you address the CI failure?

Ah, I actually forgot to do that. I suppose I must have inadvertently fixed it by moving the declaration or definition of `mbedtls_md_hmac_setup()` as the CI has passed.

PR #460, opened 2025-09-16
[tf-psa-crypto] Align header include guards with their new locations in TF PSA Crypto

### LINKED ISSUE #399: Align header include guards with their new locations in TF PSA Crypto
### Suggested enhancement

Align the include guards embedded in our header files with the file locations in TF PSA Crypto

### Justification

We've inherited a lot of headers from Mbed TLS that still have include guards that still reflect their old filenames and locations in Mbed TLS 3.x. Our include guards by convention are derived from the location and filename of the header file, which reduces the chance of an accidental collision of include guards - which would cause build errors if both headers are included in the same source file.

Examples:
- Path: `core/alignment.h`
  - Include guard: `MBEDTLS_LIBRARY_ALIGNMENT_H`
  - 3.6 Path:  `library/alignment.h`
- Path: `drivers/builtin/src/psa_util_internal.h`
  - Include guard: `MBEDTLS_PSA_UTIL_INTERNAL_H`
   - 3.6 Path: `library/psa_util_internal.h`
- Path: `drivers/builtin/src/aesni.h`
  - Include guard: `MBEDTLS_AESNI_H`
  - 3.6 Path: `library/aesni.h`
  - 2.x Path: `include/mbedtls/aesni.h`
- Path: `drivers/everest/include/tf-psa-crypto/private/everest/everest.h`
  - Include guard: `MBEDTLS_EVEREST_H`
  - 3.6 path: `3rdparty/everest/include/everest/everest.h`

valeriosetti:
## Description

Resolves #399

Due to the large amount of files affected and the repetitive type of work, all the changes in this PR were generated by a Python script. This means that in case of any conflict with ongoing works, this PR can easily be updated.

The script is currently included in the 1st commit of this PR for reviewers' reference. I put it in `./scripts/` folder and I designed it in a way that can be used both for updating guards and also for checking them. Currently this check feature is not included in the CI, but I added that because I thought it could be a nice feature to have in the future.

## PR checklist

- [x] **changelog** not required because: this change should not affect end user 
- [x] **framework PR** ~provided https://github.com/Mbed-TLS/mbedtls-framework/pull/216~ not required
- **tests** not required because: no code change


mpg:
> Because they are very indented in the tree and I feared that this could cause some code style complain for going over the 100 columns.

Actually the C standard only requires the 63 first characters to be significant in a macro name, so I guess we should avoid macro names longer that 63 chars? Or are we confident that all implementations used in practice will not have this limit? (Unsurprisingly, [GCC has no limit](https://gcc.gnu.org/onlinedocs/cpp/Implementation-limits.html) and I'd expect the same from Clang.)

gilles-peskine-arm:
> For the time being build_psa_alt_headers is the only failing test in the CI, but this requires a fix in the mbedtls branch.

I haven't looked closely, but at a guess, the problem is that both projects have headers with the name name? That's a move in progress, the crypto files have already been created but [the mbedtls files are still there](https://github.com/Mbed-TLS/mbedtls/pull/10394). I wasn't prioritizing the removal because it seemed like a minor cleanup, but I'll raise the priority it since it's blocking something else. (Still below 1.0 work though.)

valeriosetti:
> > For the time being build_psa_alt_headers is the only failing test in the CI, but this requires a fix in the mbedtls branch.
> 
> I haven't looked closely, but at a guess, the problem is that both projects have headers with the name name? That's a move in progress, the crypto files have already been created but [the mbedtls files are still there](https://github.com/Mbed-TLS/mbedtls/pull/10394). I wasn't prioritizing the removal because it seemed like a minor cleanup, but I'll raise the priority it since it's blocking something else. (Still below 1.0 work though.)

Actually I think the problem is related to the fact `_alt.h` header files are generated form the original ones with:
`make -C tests ../framework/tests/include/alt-extra/psa/crypto_platform_alt.h ../framework/tests/include/alt-extra/psa/crypto_struct_alt.h`
and in the original ones I renamed `PSA_CRYPTO_PLATFORM_H` -> `TF_PSA_CRYPTO_INCLUDE_PSA_CRYPTO_PLATFORM_H`, but at the same time checks in `framework/tests/programs/query_included_headers.c` are still in the "old" format.

gilles-peskine-arm:
> If so, what about `TF_PSA_CRYPTO_<rel_path>_<filename>_H` where `rel_path` is the path through which the header file is usually included in the library (ex: `mbedtls/xxx.h`)?

That is what I would have expected a priori.

drivers/builtin/include/mbedtls/config_adjust_legacy_crypto.h
```diff
@@ -21,8 +21,8 @@
  *  SPDX-License-Identifier: Apache-2.0 OR GPL-2.0-or-later
  */
 
-#ifndef MBEDTLS_CONFIG_ADJUST_LEGACY_CRYPTO_H
-#define MBEDTLS_CONFIG_ADJUST_LEGACY_CRYPTO_H
+#ifndef TF_PSA_CRYPTO_MBEDTLS_CONFIG_ADJUST_LEGACY_CRYPTO_H
+#define TF_PSA_CRYPTO_MBEDTLS_CONFIG_ADJUST_LEGACY_CRYPTO_H
```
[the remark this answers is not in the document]
valeriosetti: I thought this was the [pattern we agreed on](https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/460#issuecomment-3302103636) for these guards, but apparently I missed something. Didn't we want a `TF_PSA_CRYPTO_` prefix for all the guards?

bjwtaylor (dismissed):

mpg:
Unfortunately this now has conflicts. Can you rebase? Thanks!

valeriosetti:
> Unfortunately this now has conflicts. Can you rebase? Thanks!

Yeah, I knew this, but as I mentioned to Ben yesterday on Slack since all the changes in this PR are automated through a Python script (so it's easy to re-run) and at the same it's easy to get conflicts, my preferred workflow was:
1. find an overall agreement on the proposed implementation;
2. rebase and fix conflicts;
3. re-approve shortly after the rebase in order to avoid conflicts.

@mpg if you agree with current design I will proceed with step (2)

mpg:
Ah yes, makes sense! I should be able to review later this morning.

scripts/c_header_guards.py
```diff
@@ -0,0 +1,157 @@
+#!/usr/bin/env python3
+"""
+Script to check/fix header guards in C header files.
+Converts guards to the format
+    TF_PSA_CRYPTO_<rel_path>_<filename>_H
+where:
+- 'TF_PSA_CRYPTO_' is the common hardcoded prefix;
+- 'rel_path' is the path used in C files to include such header file.
```
mpg, line 20: I don't think that's what the script implements. I think for public headers it's just `<rel_path>_<filename>_H` and the prefix is only added for private headers.

Note: I think that's also not exactly the rule [Gilles was suggesting](https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/460#issuecomment-3311375010): my understanding was that (even for private headers) we would only add a `TF_PSA_CRYPTO_` prefix if `<rel_path>` doesn't already start with one of `MBEDTLS_`, `PSA_` or `TF_PSA_CRYPTO`. Note that for all public headers, `rel_path` always starts with one of these, and for private headers sometimes it does (`drivers/builtin/include`) and sometimes not (`drivers/builtin/src`).

I'm not sure if the deviation from Gilles's suggestion was intentional?

Regardless we want the documentation to be aligned with the code :)

drivers/builtin/include/mbedtls/private/aes.h
```diff
@@ -25,8 +25,8 @@
  *  SPDX-License-Identifier: Apache-2.0 OR GPL-2.0-or-later
  */
 
-#ifndef MBEDTLS_AES_H
-#define MBEDTLS_AES_H
+#ifndef TF_PSA_CRYPTO_MBEDTLS_PRIVATE_AES_H
```
mpg: For example I think `MBEDTLS_PRIVATE_AES_H` would be good enough here.

scripts/c_header_guards.py
```diff
@@ -0,0 +1,157 @@
...
+import framework_scripts_path # pylint: disable=unused-import
+from mbedtls_framework import build_tree
+
+PREFIX = 'TF_PSA_CRYPTO_'
+
+# Dictionary containing the folders where header files are to be fixed/checked:
+# - key: path of header files to check;
+# - value: path that must be trimmed from the key value in order to get the
+#          final guard.
+FOLDERS_TO_CHECK = {
```
mpg, line 38: I'm also seeing headers in `core`.

scripts/c_header_guards.py
```diff
@@ -0,0 +1,157 @@
...
+    if adjusted_name.startswith(trim_path):
+        adjusted_name = adjusted_name[len(trim_path):]
+    # Remove suffix
+    if adjusted_name.endswith('.h'):
+        adjusted_name = adjusted_name[:-2]
+
+    adjusted_name = re.sub(r'[^a-zA-Z0-9]', '_', adjusted_name).upper()
+    guard_elements = []
+    # Add prefix only to non public header files
+    if file_path.startswith('drivers'):
```
mpg: Missing private headers in `core`.

Also, if we switch to the rule suggested by Gilles, the condition should be
```py
if not (adjusted_name.startswith('MBEDTLS_') or adjusted_name.startswith('PSA_') or adjusted_name.startswith('TF_PSA_CRYPTO_')):
```
That is, we no longer need to distinguish between private and public headers, we just check if the current name already has one of our 3 prefixes or not.

scripts/c_header_guards.py
```diff
@@ -0,0 +1,157 @@
+#!/usr/bin/env python3
+"""
+Script to check/fix header guards in C header files.
+Converts guards to the format
+    TF_PSA_CRYPTO_<rel_path>_<filename>_H
+where:
+- 'TF_PSA_CRYPTO_' is the common hardcoded prefix;
+- 'rel_path' is the path used in C files to include such header file.
+
+Usage: python3 fix_header_guards.py
```
mpg: The file is called `c_header_guards.py` now :)

scripts/c_header_guards.py
```diff
@@ -0,0 +1,157 @@
...
+    guard_elements = []
+    # Add prefix only to non public header files
+    if file_path.startswith('drivers'):
+        guard_elements.append('TF_PSA_CRYPTO')
+    guard_elements.extend([adjusted_name, 'H'])
+    return '_'.join(guard_elements)
+
+def find_line(lines: List[str], search_range, regex: str) -> Tuple[int, Match]:
+    """Try to match the regex expression on the range of lines given in input.
+    The first successful match is returned; an execption is returned if no
```
mpg: Typo: exception

scripts/c_header_guards.py
```diff
@@ -0,0 +1,157 @@
...
+            return i, match
+
+    raise ValueError(f'Error: {regex} line not found')
+
+def process_header(file_path, trim_path, fix) -> None:
+    """Check/fix header guard in a single file.
+    On failure an exception is thrown.
+    """
+    with open(file_path, 'r', encoding='utf-8') as f:
+        content = f.read()
```
mpg: Minor: a bit more compact than this plus `lines = content.split('\n')` below would be `lines = f.readlines()` here.

scripts/c_header_guards.py
```diff
@@ -0,0 +1,157 @@
...
+    return '_'.join(guard_elements)
+
+def find_line(lines: List[str], search_range, regex: str) -> Tuple[int, Match]:
+    """Try to match the regex expression on the range of lines given in input.
+    The first successful match is returned; an execption is returned if no
+    match is found in all the given lines.
+
+    Input params:
+        lines: content of the file given as list of strings;
+        search_range: range() to use when going through the list of lines;
```
mpg, line 77: Note: I like how you're using a range object as a parameter here, that's not something I would have considered, I've learned a new useful Python idiom today, thanks!

scripts/c_header_guards.py
```diff
@@ -0,0 +1,157 @@
...
+    lines = content.split('\n')
+
+    ifndef_line, match = find_line(lines, range(0, len(lines)),
+                                   r'#ifndef\s+(\w+)\b')
+    expected_line = f'#ifndef {correct_guard}'
+    if expected_line not in lines[ifndef_line]:
+        if fix:
+            lines[ifndef_line] = expected_line
+        else:
+            raise ValueError(f'Invalid guard format:\n'
```
mpg, line 108: Note: I'm not a big fan of using exceptions for normal errors, but I guess it's a matter of taste/style, and considering this script is not central, I'm not going to argue about this.

scripts/c_header_guards.py
```diff
@@ -0,0 +1,157 @@
...
+    # Write back to file
+    new_content = '\n'.join(lines)
+    with open(file_path, 'w', encoding='utf-8') as f:
+        f.write(new_content)
+
+def main():
+    current_path = os.getcwd()
+
+    arg_parser = argparse.ArgumentParser()
+    arg_parser.add_argument('--fix', action='store_true',
```
mpg, line 143: Ah, so there is a command-line option. That should probably be reflected in the "Usage" line in the script's docstring.

mpg (changes requested):
Looking pretty good, most of the few issues I found are mostly in docstring, I think the only one that matters in the code is that `core` is missing.

Also, since the script can check guards, not just fix them, I think the check should be added to `all.sh`, to catch any mistakes in future header files we create. Probably a new component in `tests/scripts/components-basic-checks.sh`.

scripts/c_header_guards.py
valeriosetti: > I'm not sure if the deviation from Gilles's suggestion was intentional?

No, I missed that part. OTOH since all we're changing here is guards for the header files, wouldn't it be better to use a fixed and simple approach (like the one that you just described in the 1st paragraph of your comment) instead of adding exceptions that make code harder to read?

> Regardless we want the documentation to be aligned with the code :)

Yes, I totally forgot to update it and I will do it now.

scripts/c_header_guards.py
valeriosetti: Let's find an agreement on the [pattern](https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/460#discussion_r2454888334) first and then if we agree to go this way I will update the script ;)

scripts/c_header_guards.py
valeriosetti: I included the "core/" folder in the search and check path of the script. The final result is at [this commit](https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/460/commits/940d834416388ce492806ac11de3ddf7ce9c472a).
Is it a reasonable result in your opinion?

I know that for some header files which already have `psa_crypto_` in the file name (ex: `core/psa_crypto_core.h`) the final result is a bit too long (for the file that I just mentioned it is `TF_PSA_CRYPTO_PSA_CRYPTO_CORE_H`), but I prefer to have a common pattern, rather than making exceptions. However I know that this is a matter of preference :)

mpg (approved):
LGTM, thanks!

bjwtaylor (approved):

PR #462, opened 2025-09-17
Remove our copy of the driver interface proposal

gilles-peskine-arm:
The document is now owned by PSA.

Note: I'm not sure if we _should_ remove our copy. If so, we shouldn't remove it just yet, we should at least wait until https://github.com/ARM-software/psa-api/pull/299 is merged.

Argument for keeping: we've published links to it. But these are links to the mbedtls repository, not to the crypto repository, so the mbedtls development branch is where people would look for it.

Argument for keeping: it describes what we're doing. The PSA spec may evolve and until we catch up with it, there would no longer be any documentation of what we're doing.

Argument for removal: if you want this document, it exists in the 3.6 LTS branch. (One change has been done since 3.6.0: 92673db5db061562dd6713ec76150b4904c7a122, in the random section, which we don't implement.)

Argument for removal: we don't implement the whole document anyway, and we don't have anything written about which parts we do implement. So as long as PSA doesn't make incompatible changes, the PSA document is as good as our.

## PR checklist

## Notes for the submitter

Help make review efficient:
* Multiple simple commits
* Avoid force-push


docs/proposed/psa-driver-interface.md
```diff
@@ -3,1465 +3,8 @@ PSA Cryptoprocessor Driver Interface
```
gilles-peskine-arm, line 3: If we remove the substance of this document, we should stop linking to it in `README.md` and elsewhere in our code base.

PR #468, opened 2025-09-23
Finalize pk.h

### LINKED ISSUE #205: Remove private pk functions from public documentation
Depends: new PK mostly done, so that it's clear exactly what is going into the 1.0 release.
 
Do a pass over `pk.h` to check for remaining obsolete statements and inconsistencies. In particular, check for mentions of API elements that have become private or have been removed.

Get rid any remaining mention of “opaque” in `pk.h`: it should now only refer to PSA keys that are in a secure element, not to all keys that are in PSA.

Also double-check that `mbedtls_pk_type_t` is no longer mentioned in any public documentation.

### LINKED ISSUE #347: Update documentation of mbedtls_pk_sign and mbedtls_pk_verify
Document that `mbedtls_pk_sign()` and `mbedtls_pk_verify()` are legacy functions, that perform the same algorithm that `mbedtls_pk_get_psa_attributes()` would pick in the key policy if given a sign or verify usage.

(Context: https://github.com/Mbed-TLS/TF-PSA-Crypto/blob/development/docs/architecture/pk-4.md#changes-to-private-key-selection-in-tls)

mpg:
## Description

Resolve #347 and resolve #205

**Depends on:** https://github.com/Mbed-TLS/mbedtls/pull/10418 for a clean CI (sigh).

## PR checklist

- [x] **changelog** provided
- [x] **mbedtls 3.6 PR** not required because: 1.0 only
- **tests**  not required because: docs only

### DIFF
--- a/include/mbedtls/pk.h
+++ b/include/mbedtls/pk.h
@@ -482,6 +482,16 @@ int mbedtls_pk_import_into_psa(const mbedtls_pk_context *pk,
  *                  independent from the original PSA key that it was generated
  *                  from.
  *
+ * \note            This function only copies the key material but discards
+ *                  policy information entirely. See \c
+ *                  mbedtls_pk_get_psa_attributes() for details on which
+ *                  algorithm is going to be used by PK for contexts set up with
+ *                  this function.
+ *
+ *                  If you want to retain the PSA policy, see \c
+ *                  mbedtls_pk_wrap_psa() - but then the PSA key needs to live
+ *                  at least as long as the PK context.
+ *
  * \param key_id    The key identifier of the key stored in PSA.
  * \param pk        The PK context that will be filled. It must be initialized,
  *                  but not set up.
@@ -502,6 +512,16 @@ int mbedtls_pk_copy_from_psa(mbedtls_svc_key_id_t key_id, mbedtls_pk_context *pk
  *                  Once this function returns the PK object will be completely
  *                  independent from the original PSA key that it was generated
  *                  from.
+ *
+ * \note            This function only copies the key material but discards
+ *                  policy information entirely. See \c
+ *                  mbedtls_pk_get_psa_attributes() for details on which
+ *                  algorithm is going to be used by PK for contexts set up with
+ *                  this function.
+ *
+ *                  If you want to retain the PSA policy, see \c
+ *                  mbedtls_pk_wrap_psa() - but then the PSA key needs to live
+ *                  at least as long as the PK context.
  *
  * \param key_id    The key identifier of the key stored in PSA.
  * \param pk        The PK context that will be filled. It must be initialized,
--- a/include/mbedtls/pk.h
+++ b/include/mbedtls/pk.h
@@ -19,18 +19,6 @@
 
 #include "mbedtls/md.h"
 
-#if defined(MBEDTLS_RSA_C)
-#include "mbedtls/private/rsa.h"
-#endif
-
-#if defined(MBEDTLS_ECP_C)
-#include "mbedtls/private/ecp.h"
-#endif
-
-#if defined(MBEDTLS_ECDSA_C)
-#include "mbedtls/private/ecdsa.h"
-#endif
-
 #if defined(MBEDTLS_PSA_CRYPTO_CLIENT)
 #include "psa/crypto.h"
 #endif
--- a/include/mbedtls/pk.h
+++ b/include/mbedtls/pk.h
@@ -240,32 +240,23 @@ void mbedtls_pk_restart_free(mbedtls_pk_restart_ctx *ctx);
 #endif /* MBEDTLS_ECP_RESTARTABLE */
 
 /**
- * \brief Initialize a PK context to wrap a PSA key.
+ * \brief Set up a PK context to wrap a PSA key pair.
  *
- * This function creates a PK context which wraps a PSA key. The PSA wrapped
- * key must be an EC or RSA key pair (DH is not suported in the PK module).
+ * The PSA key must be an EC or RSA key pair (FFDH is not suported in PK).
  *
- * Under the hood PSA functions will be used to perform the required
- * operations and, based on the key type, used algorithms will be:
- * * EC:
- *     * verify, verify_ext, sign, sign_ext: ECDSA.
- * * RSA:
- *     * sign: use the primary algorithm in the wrapped PSA key;
- *     * sign_ext: RSA PSS if the pk_type is #MBEDTLS_PK_SIGALG_RSA_PSS, otherwise
- *       it falls back to the sign() case;
- *     * verify, verify_ext: not supported.
- *
- * In order for the above operations to succeed, the policy of the wrapped PSA
- * key must allow the specified algorithm.
- *
- * PK contexts wrapping an EC keys also support \c mbedtls_pk_check_pair(),
- * whereas RSA ones do not.
+ * The resulting context can only perform operations that are allowed by the
+ * key's policy. Additionally, it currently has the following limitations:
+ * module).
+ * - restartable operations can't be used;
+ * - for RSA keys, signature verification is not supported, and neither is use
+ *   of \c mbedtls_pk_check_pair().
  *
  * \warning The PSA wrapped key must remain valid as long as the wrapping PK
  *          context is in use, that is at least between the point this function
  *          is called and the point mbedtls_pk_free() is called on this context.
  *
- * \param ctx The context to initialize. It must be empty (type NONE).
+ * \param ctx The context to set up. It must have been freshly initialized
+ *            but not set up yet.
  * \param key The PSA key to wrap, which must hold an ECC or RSA key pair.
  *
  * \return    \c 0 on success.
--- a/include/mbedtls/pk.h
+++ b/include/mbedtls/pk.h
@@ -509,19 +509,19 @@ int mbedtls_pk_copy_public_from_psa(mbedtls_svc_key_id_t key_id, mbedtls_pk_cont
 /**
  * \brief           Verify signature (including padding if relevant).
  *
- * \param ctx       The PK context to use. It must have been set up.
- * \param md_alg    Hash algorithm used.
- * \param hash      Hash of the message to sign
- * \param hash_len  Hash length
- * \param sig       Signature to verify
- * \param sig_len   Signature length
- *
  * \note            The signature algorithm used will be the one that would be
  *                  selected by \c mbedtls_pk_get_psa_attributes() called with a
  *                  usage of #PSA_KEY_USAGE_VERIFY_HASH - see that function's
  *                  documentation for details.
  *                  If you want to request a specific signature algorithm, see
  *                  \c mbedtls_pk_verify_ext().
+ *
+ * \param ctx       The PK context to use. It must have been set up.
+ * \param md_alg    Hash algorithm used.
+ * \param hash      Hash of the message to sign
+ * \param hash_len  Hash length
+ * \param sig       Signature to verify
+ * \param sig_len   Signature length
  *
  * \return          0 on success (signature is valid),
  *                  #PSA_ERROR_INVALID_SIGNATURE if the signature is invalid,
@@ -598,6 +598,13 @@ int mbedtls_pk_verify_ext(mbedtls_pk_sigalg_t type,
 /**
  * \brief           Make signature (including padding if relevant).
  *
+ * \note            The signature algorithm used will be the one that would be
+ *                  selected by \c mbedtls_pk_get_psa_attributes() called with a
+ *                  usage of #PSA_KEY_USAGE_SIGN_HASH - see that function's
+ *                  documentation for details.
+ *                  If you want to request a specific signature algorithm, see
+ *                  \c mbedtls_pk_sign_ext().
+ *
  * \param ctx       The PK context to use. It must have been set up
  *                  with a private key.
  * \param md_alg    Hash algorithm used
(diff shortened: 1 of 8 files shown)


include/mbedtls/pk.h
```diff
@@ -283,32 +256,21 @@ void mbedtls_pk_restart_free(mbedtls_pk_restart_ctx *ctx);
...
- *
- * In order for the above operations to succeed, the policy of the wrapped PSA
- * key must allow the specified algorithm.
- *
- * PK contexts wrapping an EC keys also support \c mbedtls_pk_check_pair(),
- * whereas RSA ones do not.
+ * The resulting context can only perform operations that are allowed by the
+ * key's policy. Additionally, it currently has the following limitations:
+ * - restartable operations can't be used;
+ * - for RSA keys, signature verification is not supported, and neither is use
```
gilles-peskine-arm, line 266: Is verification supported with ECDSA? I can't find that in the test code.

include/mbedtls/pk.h
```diff
@@ -447,30 +409,22 @@ int mbedtls_pk_can_do_psa(const mbedtls_pk_context *pk, psa_algorithm_t alg,
  *                    \p usage, exporting and copying the key, and
  *                    possibly other permissions as documented for the
  *                    \p usage parameter.
- *                    The permitted algorithm policy is determined as follows
- *                    based on the #mbedtls_pk_sigalg_t type of \p pk,
- *                    the chosen \p usage and other factors:
- *                      - #MBEDTLS_PK_SIGALG_RSA_PKCS1V15 whose underlying
- *                        context uses the PKCS#1 v1.5 padding mode:
+ *                    For keys created with \c mbedtls_pk_wrap_psa(), the
+ *                    permitted algorithm is the same as the original PSA key.
```
gilles-peskine-arm, line 413: > the permitted algorithm is the same as the original PSA key.

The thing is, a PSA key does not in general have “the permitted algorithm”. It could be the same permitted algorithm _policy_ (including the enrollment algorithm), or it could be the PSA key's primary algorithm (which is what it was in 3.6, and AFAIK this hasn't changed).

include/mbedtls/pk.h
```diff
@@ -522,24 +476,17 @@ int mbedtls_pk_import_into_psa(const mbedtls_pk_context *pk,
                                mbedtls_svc_key_id_t *key_id);
 
 /**
- * \brief           Create a PK context starting from a key stored in PSA.
+ * \brief           Set up a PK context with the key material from a PSA key
+ *                  pair.
```
gilles-peskine-arm, line 480: Not necessarily a key pair, it can be a public key.

include/mbedtls/pk.h
```diff
@@ -522,24 +476,17 @@ int mbedtls_pk_import_into_psa(const mbedtls_pk_context *pk,
...
  *
- *                  Once this functions returns the PK object will be completely
+ *                  Once this function returns the PK object will be completely
  *                  independent from the original PSA key that it was generated
  *                  from.
- *                  Calling mbedtls_pk_sign() or mbedtls_pk_verify(), on the
- *                  resulting PK context will perform the corresponding
- *                  algorithm for that PK context type.
- *                  * For ECDSA, the choice of deterministic vs randomized will
- *                    be based on #MBEDTLS_PK_ALG_ECDSA.
```
gilles-peskine-arm, line 537: This information is still relevant. And for RSA keys, we should state that PKCS#1v1.5 signature (`PSA_ALG_RSA_PKCS1V15_SIGN(hash_alg)` or `PSA_ALG_RSA_PKCS1V15_SIGN_RAW`) will be used.

I do think this should be mentioned explicitly, because it can be somewhat surprising: you don't necessarily end up with the same algorithm with `mbedtls_pk_wrap_psa`.

Also applies to `mbedtls_pk_copy_public_from_psa`.

include/mbedtls/pk.h
```diff
@@ -611,20 +550,27 @@ int mbedtls_pk_verify(mbedtls_pk_context *ctx, mbedtls_md_type_t md_alg,
...
- *                  \c mbedtls_ecp_set_max_ops() to reduce blocking for ECC
+ *                  \c psa_interruptible_set_max_ops() to reduce blocking for ECC
  *                  operations. For RSA, same as \c mbedtls_pk_verify().
  *
+ * \note            See the note on \c mbedtls_pk_verify() for which algorithm is
+ *                  used. If you want to select a specific signature algorithm
+ *                  since the is no restartable equivalent of \c
+ *                  mbedtls_pk_verify_ext(), you'll want to import the key into
+ *                  PSA using \c mbedtls_pk_import_into_psa() and then call PSA
+ *                  interruptible functions directly.
```
gilles-peskine-arm, line 561: Given that this function is only useful for ECDSA, and there is only one ECDSA verification algorithm, I think this whole note is useless. For RSA, there's already a sentence above that says it's the same as `mbedtls_pk_verify`.

include/mbedtls/pk.h
```diff
@@ -633,42 +579,42 @@ int mbedtls_pk_verify_restartable(mbedtls_pk_context *ctx,
                                   mbedtls_pk_restart_ctx *rs_ctx);
 
 /**
- * \brief           Verify signature, with explicit selection of the signature algorithm.
- *                  (Includes verification of the padding depending on type.)
+ * \brief           Verify signature, selecting a specific algorithm.
+ *                  (Includes verification of the padding if applicable.)
```
gilles-peskine-arm, line 583: “includes verification of the padding” is an obsolete concept from back when textbook RSA was considered a cryptographic algorithm rather than a low-level building block. Please remove this clause while you're at it.

include/mbedtls/pk.h
```diff
@@ -696,12 +634,23 @@ int mbedtls_pk_sign(mbedtls_pk_context *ctx, mbedtls_md_type_t md_alg,
...
+ *                  return early and restart according to the limit set with \c
+ *                  psa_interruptible_set_max_ops() to reduce blocking for ECC
+ *                  operations. For RSA, same as \c mbedtls_pk_sign().
+ *
+ * \note            See the note on \c mbedtls_pk_sign() for which algorithm is
+ *                  used. If you want to select a specific signature algorithm
+ *                  since the is no restartable equivalent of \c
+ *                  mbedtls_pk_sign_ext(), you'll want to import the key into
+ *                  PSA using \c mbedtls_pk_import_into_psa() and then call PSA
+ *                  interruptible functions directly.
```
gilles-peskine-arm, line 649: Given that wrapped PSA keys aren't supported (which should be mentioned in this function's description, btw), this simplifies to “this function always uses `MBEDTLS_PK_ALG_ECDSA`” (plus being equivalent to `mbedtls_pk_sign` for RSA keys as stated just above).

gilles-peskine-arm (changes requested):
I have reviewed for correctness (with the loss of distinction between v1.5 and v2.1 RSA keys that we discussed on Slack and that will be implemented in https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/471).

I haven't reviewed for completeness yet, I'll do another pass for that but it may have to wait until tomorrow.

include/mbedtls/pk.h
valeriosetti: Yes, IMO it's [supported](https://github.com/Mbed-TLS/TF-PSA-Crypto/blob/779296b0fb7165d7f36f436ca34085afff48c02c/drivers/builtin/src/pk_wrap.c#L715)

include/mbedtls/pk.h
```diff
@@ -552,23 +499,16 @@ int mbedtls_pk_import_into_psa(const mbedtls_pk_context *pk,
...
- *                  Calling mbedtls_pk_verify() on the resulting
- *                  PK context will perform the corresponding algorithm for that
- *                  PK context type.
- *
- *                  For an RSA key,
- *                  the original key's policy determines the output key's padding
- *                  mode: PCKS1 v2.1 is set if the PSA key policy is OAEP or PSS,
- *                  otherwise PKCS1 v1.5 is set.
- *
+
```
valeriosetti, line 511:
```suggestion
 *
```

include/mbedtls/pk.h
```diff
@@ -583,24 +523,23 @@ int mbedtls_pk_copy_public_from_psa(mbedtls_svc_key_id_t key_id, mbedtls_pk_cont
 /**
  * \brief           Verify signature (including padding if relevant).
  *
+ * \note            The signature algorithm used will be the one that would be
+ *                  selected by \c mbedtls_pk_get_psa_attributes() called with a
+ *                  usage of #PSA_KEY_USAGE_VERIFY_HASH - see that function's
+ *                  documentation for details.
```
valeriosetti, line 529: I'm not sure this is consistent with what will be `mbedtls_pk_get_psa_attributes` after https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/471. For example for an RSA key `mbedtls_pk_get_psa_attributes` will always return PKCS v1.5 algorithm, whereas `rsa_verify_wrap` picks the proper padding mode depending on `pk->rsa_padding`.
Will this be updated after https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/471?

valeriosetti (changes requested):

include/mbedtls/pk.h
mpg: I had not checked tests and has just looked at the info structure fro which pointers are not NULL. Thank you Valerio for locating tests.

include/mbedtls/pk.h
mpg: > whereas `rsa_verify_wrap` picks the proper padding mode depending on `pk->rsa_padding`.

Unless I'm missing something, the value of `pk->rsa_padding` will always be v1.5 at this point because:
- that's the default set by `init()`
- after 471 none of the 5 public function for setting up a key modify it
- there is no public method for modifying it.

Specifically, in 471, `mbedtls_pk_set_rsa_(padding|hash)()` are used only in tests, not in the library (and not in mbedtls).

In fact, I think we should remove that field (and `rsa_hash`) entirely. I didn't initially suggest doing that in 471 in order to keep it minimal, but that's something I'd like to do right after it. But perhaps it makes more sense to do it in 471 in order to be more confident about the fact that they're not used? (The same reason why I removed `pk_context::pk_ctx` here.) Wdyt?

As Gilles said previously, a guiding principle of PK for 1.0 is that it doesn't do policy - that's PSA's job. According to this principle, we shouldn't have those fields in `pk_context`. (Again, it makes sense that you introduced them in the initial refactoring of how RSA keys are held, because it's better to not change existing behaviour when refactoring, but I think there should have been a 2nd step where we change the behaviour to simplify and align with the new principles.)

valeriosetti: > Unless I'm missing something, the value of pk->rsa_padding will always be v1.5 at this point because:

Oh, you are perfectly right, sorry.

> In fact, I think we should remove that field (and rsa_hash) entirely. I didn't initially suggest doing that in 471 in order to keep it minimal, but that's something I'd like to do right after it. But perhaps it makes more sense to do it in 471 in order to be more confident about the fact that they're not used?

Well, I can open a follow-up PR so that we can both have 471 merged soon (if it's OK) and then keep other changes in the follow-up

include/mbedtls/pk.h
gilles-peskine-arm: > there is not public concept of an algorithm associated with a key

We know that, but users don't necessarily know that. An interface like pk could reasonably have a specific algorithm associated with each key, and could reasonably treat keys as pure mathematical objects with no policy. It's not obvious that we went for the latter. It's even less obvious for users who have used previous versions of Mbed TLS where a key did have partial policy information.

This is why I think this should be mentioned explicitly in the documentation of `copy_from_psa`, not just by reference to the documentation of other functions. We don't necessarily have to have the specifics here, but at least we should state the general principle that this is functionally equivalent to export+import and that no policy information carries over.

include/mbedtls/pk.h
mpg: Duh, it says right below "key pair or public key". What was I thinking.

gilles-peskine-arm (changes requested):
Mostly LGTM now, just a couple of things left.

I also had a high-level look at `pk.h` to look for other things to change, and the only things that stood out were:

* The [“set up” terminology](https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/468#discussion_r2378928212).
* [The salt length for `mbedtls_pk_verify_ext()`](https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/468/files#r2379347862).

include/mbedtls/pk.h
```diff
@@ -628,42 +589,41 @@ int mbedtls_pk_verify_restartable(mbedtls_pk_context *ctx,
...
  *                  or a specific error code.
  *
- * \note            If hash_len is 0, then the length associated with md_alg
- *                  is used instead, or an error returned if it is invalid.
- *
- * \note            \p options parameter is kept for backward compatibility.
- *                  If key type is different from MBEDTLS_PK_SIGALG_RSA_PSS it must
- *                  be NULL, otherwise it's just ignored.
  */
 int mbedtls_pk_verify_ext(mbedtls_pk_sigalg_t type,
```
gilles-peskine-arm, line 609: One more thing about `mbedtls_pk_verify_ext`: with the removal of `options`, we lost the information of what this function does about the salt length. It allows any salt length.

include/mbedtls/pk.h
```diff
@@ -319,7 +281,7 @@ int mbedtls_pk_wrap_psa(mbedtls_pk_context *ctx,
 /**
  * \brief           Get the size in bits of the underlying key
  *
- * \param ctx       The context to query. It must have been initialized.
+ * \param ctx       The context to query. It must have been set up.
```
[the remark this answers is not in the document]
mpg: Yeah, this was on my list at some point, then I removed it as perhaps not needed right now. Well, let's do it after all.

gilles-peskine-arm (approved):
I'm happy with 97a417cc5939f8a057921bf8e2eac45baa9a2185 but the CI isn't.

mpg:
Yes, I'm preparing an mbedtls PR to fix that. Just checking if there were other failures hidden by this one in the same component before I push.

mpg:
https://github.com/Mbed-TLS/mbedtls/pull/10418

PR #474, opened 2025-09-23
Add shadow versions of removed headers

### LINKED ISSUE #473: Keep providing removed headers to hide system headers from an older version
Consider a user who has an application working with Mbed TLS 3.6, and is the process of upgrading their code to TF-PSA-Crypto. Suppose that headers from Mbed TLS 3.6 or older are present in the include path (e.g. because they're in `/usr/include`). Then code that still includes a removed header (e.g. `#include <mbedtls/bignum.h>` may inadvertently pick up the system header, resulting in mysterious-looking build failures.

The goal of this task is to provide dummy versions of the removed headers that just `#error` with a comprehensible message.

Suggested by @gstrauss in https://github.com/Mbed-TLS/mbedtls/issues/10395#issuecomment-3320634565

gilles-peskine-arm:
Resolves #473.

## PR checklist

- [x] **changelog** not required because: only catches coding errors
- [x] **mbedtls not required because: no headers to shadow in Mbed TLS
- **tests**  provided

### DIFF
--- a/drivers/builtin/include/mbedtls/aes.h
+++ b/drivers/builtin/include/mbedtls/aes.h
@@ -3,4 +3,6 @@
  *  SPDX-License-Identifier: Apache-2.0 OR GPL-2.0-or-later
  */
 
+#if !defined(MBEDTLS_PROGRAMS_TEST_QUERY_CONFIG_H)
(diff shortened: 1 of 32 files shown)


gilles-peskine-arm:
This is very messy and would require a lot of work to differentiate the “true” headers from the shadow ones. I can't justify spending any more time on this.

PR #480, opened 2025-09-26
Do not include context types in client-only mode

davidhorstmann-arm:
When building with `MBEDTLS_PSA_CRYPTO_CLIENT` but not `MBEDTLS_PSA_CRYPTO_C`, do not include the definitions of context types specific to the builtin implementation.

These are not needed or used, as the build is 'client-only', meaning that operation structs contain an `mbedtls_psa_client_handle_t` rather than the real operation state.

## PR checklist

- [x] **changelog** provided
- [x] **mbedtls development PR** not required because: Crypto change only
- [ ] **mbedtls 3.6 PR** provided Mbed-TLS/mbedtls# (TODO)
- **tests** not required because: Header include issue

### DIFF
--- a/core/psa_crypto_core.h
+++ b/core/psa_crypto_core.h
@@ -12,6 +12,7 @@
 #include "tf-psa-crypto/build_info.h"
 
 #include "psa/crypto.h"
+#include "psa/crypto_builtin_composites.h"
 #if defined(MBEDTLS_THREADING_C)
 #include "mbedtls/threading.h"
 #endif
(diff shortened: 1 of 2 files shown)


gilles-peskine-arm (dismissed):
LGTM

I'm not sure there can be a concept of completeness here, but it's correct and does make some sense.

davidhorstmann-arm:
Sorry for the force-push, I accidentally updated the framework submodule, there should be no other changes.

mpg (approved):
LGTM if the CI agrees. As Gilles said, I can't vouch for completeness, but this looks reasonable and harmless.

PR #482, opened 2025-09-26
C23 empty initializers

keith-packard:
Add MBEDTLS_EMPTY_INITIALIZER macro for GCC 15 and clang 18 fixes

C23 defines the empty initializer, {}, with different semantics than a partial initializer containing only zero values (like {0}):

> If an object that has automatic storage duration is initialized with an empty initializer, its value is the same as the initialization of a static storage duration object." §6.7.10

Reading further, we find that this will effectively set the entire object to zero, including padding.

GCC 15 (and some builds of clang 18) start leaving portions of structs uninitialized unless we use empty initializers. Fortunately they both support empty initializers even when not compiling in c23 mode.

Because empty initializers are a new feature in C23, we cannot use them with previous versions of language, so this new macro takes a parameter to use as the initializer value for the struct when building with a pre-C23 compiler.

One unfortunate effect of using this macro is that we cannot use explicit initializer values for any members lest we lose the empty initializer semantics for the whole object. This places an implicit dependency on the representation for any symbolic initializers. Capturing that in some fashion (perhaps using a static assertion) might be a nice enhancement to this change.

### DIFF
--- a/include/mbedtls/platform_util.h
+++ b/include/mbedtls/platform_util.h
@@ -190,6 +190,22 @@ struct tm *mbedtls_platform_gmtime_r(const mbedtls_time_t *tt,
                                      struct tm *tm_buf);
 #endif /* MBEDTLS_HAVE_TIME_DATE */
 
+/* C23 adds empty intializers which provide an explicit way of initializing
+ * structs and unions in the same way as objects of static storage duration,
+ * and that ensure that the entire object is initialized, not just some portion.
+ *
+ * GCC 15 (and some builds of clang 18) start leaving portions of
+ * structs uninitialized unless we use empty initializers. Fortunately
+ * they also support empty initializers even when not compiling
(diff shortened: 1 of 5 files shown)


gilles-peskine-arm:
Can you please raise an issue with the bug that you've observed? It would be more productive to discuss the problem and its potential solutions, rather than have you propose solutions that we've already considered but don't work for this project.

mpg:
I'm also closing this PR and encouraging you to first raise an issue so that we can discuss the problem and how it interacts with the specific requirements of TF-PSA-Crypto.

I want to insist that this is not meant to close the discussion, but really to continue it in a different and hopefully more productive way. We really are interested in learning about the specific issues you observed and resolving them. But there's no point keeping a PR open that we know won't be merged with is current approach.

Thanks for your continued (I hope!) interest in TF-PSA-Crypto.

PR #484, opened 2025-09-29
Migration guide: PAKE

### LINKED ISSUE #335: Migration guide for PAKE
Write a [migration guide](https://github.com/Mbed-TLS/TF-PSA-Crypto/tree/development/docs/4.0-migration-guide) section for the PAKE API changes.

mpg:
## Description

Populate the PAKE section of the 1.0 migration guide. Resolves #335

Based on:
- the PSA ChangeLog (link in the document);
- the "Update PAKE API to 1.2" EPIC on [the board](https://github.com/orgs/Mbed-TLS/projects/18/views/1):
  - looking at changes in `crypto_extra.h`,
  - using mbedtls PRs as a basis for the before-after examples.

## PR checklist

- [x] **changelog**  not required because: documentation
- [x] **mbedtls 3.6 PR** not required because: 1.0
- **tests**  not required because: documentation

### DIFF
--- a/docs/4.0-migration-guide/pake.md
+++ b/docs/4.0-migration-guide/pake.md
@@ -1,5 +1,87 @@
 ## Changes to the PAKE interface
 
-The PAKE interface has been updated to match the PSA Crypto 1.2 API. Formerly, the PAKE interface implemented a draft version of that API. Some of the changes are incompatible and will require changes in application code.
+The PAKE interface in TF-PSA-Crypto 1.0 has been updates to match PSA Crypto API [1.2 Final](https://arm-software.github.io/psa-api/crypto/1.2/ext-pake/) PAKE extension, which is the same version that has been integrated into the main specification of PSA Crypto [version 1.3](https://arm-software.github.io/psa-api/crypto/1.3/).
 
-TODO
+In Mbed TLS 3.6, the PAKE interface implemented version [1.1 Beta](https://arm-software.github.io/psa-api/crypto/1.1/ext-pake/) of the PAKE extension. There has been a number of [changes between the beta and the final version](https://arm-software.github.io/psa-api/crypto/1.2/ext-pake/appendix/history.html#changes-between-beta-1-and-final) of the API. The changes that require applications to update their code are detailed in the following subsections.
+
+Note that TF-PSA-Crypto 1.0 still only implements `PSA_ALG_JPAKE` (and only on elliptic curves, specifically only on secp256r1). Support for SPAKE2+ is likely to be added in a future version but is not there yet.
+
+### Combine `psa_pake_set_password_key()` with `psa_pake_setup()`
+
+The function `psa_pake_set_password_key()` has been removed. Its `key` argument is now passed to `psa_pake_setup()` which has gained a new `key` parameter.
+
+Before:
+
+```
+status = psa_pake_setup(&operation, &cipher_suite);
+if (status != PSA_SUCCESS) // error handling ommited for brevity
+status = psa_pake_set_password_key(&operation, key);
+if (status != PSA_SUCCESS) // error handling ommited for brevity
+```
+
+Now:
+
+```
+status = psa_pake_setup(&operation, key, &cipher_suite);
+if (status != PSA_SUCCESS) // error handling ommited for brevity
+```
+
+### Move the hash algorithm parameter into the algorithm identifier
+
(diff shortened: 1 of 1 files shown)


yanesca (commented):
Found a typo and have a non-blocker question, but other than that it looks good to me.

docs/4.0-migration-guide/pake.md
```diff
@@ -1,5 +1,87 @@
...
+
+Now:
+
+```
+psa_key_id_t shared_key_id = (psa_key_id_t) 0;
+
+psa_key_attributes_t shared_key_attributes = PSA_KEY_ATTRIBUTES_INIT;
+psa_set_key_usage_flags(&shared_key_attributes, PSA_KEY_USAGE_DERIVE);
+psa_set_key_algorithm(&shared_key_attributes, alg);
+psa_set_key_type(&shared_key_attributes, PSA_KEY_TYPE_DERIVE);
```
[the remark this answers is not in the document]
mpg: Looking at the code, we don't seem to enforce any specific key type here. We probably should. Then we should have negative tests when the key type is not good, and also a positive test for `KEY_TYPE_HMAC`. I'm creating an issue to track this.

In the meantime, I'll add a sentence here about HMAC.

yanesca (dismissed):
LGTM

docs/4.0-migration-guide/pake.md
```diff
@@ -1,5 +1,111 @@
...
+
+```
+status = psa_pake_get_implicit_key(&pake_op, &derivation_op);
+if (status != PSA_SUCCESS) // error handling ommited for brevity
+```
+
+Now:
+
+```
+psa_algorithm_t alg = PSA_ALG_TLS12_PSK_TO_MS(PSA_ALG_SHA_256);
```
gilles-peskine-arm, line 68: Minor: wouldn't `PSA_ALG_TLS12_ECJPAKE_TO_PMS` be a better example?

docs/4.0-migration-guide/pake.md
```diff
@@ -1,5 +1,111 @@
...
+                      // or other key derivation algorithm
```
gilles-peskine-arm, line 69: It has to be the algorithm that `derivation_op` was set up for.

I think it would be helpful to include the code to set up and use the derivation operation, in both the before and after snippets.

A typical audience for the migration guide is someone who may be upgrading multiple libraries used in a project, and who isn't very familiar with the code that they're upgrading. So we should guide them with respect to understanding their existing application code.

docs/4.0-migration-guide/pake.md
```diff
@@ -1,5 +1,111 @@
...
+                                 &shared_key_attributes,
+                                 &shared_key_id);
+if (status != PSA_SUCCESS) // error handling ommited for brevity
+
+status = psa_key_derivation_input_key(&derivation_op,
+                                      PSA_KEY_DERIVATION_INPUT_SECRET,
+                                      shared_key_id);
+if (status != PSA_SUCCESS) // error handling ommited for brevity
+
+psa_reset_key_attributes(&shared_key_attributes);
```
gilles-peskine-arm, line 87: Minor: this can be done as soon as `psa_pake_get_shared_key` returns.

Also, it doesn't _have_ to be done, but it's arguably good practice, so I agree with having it in, and not making any mention of it being optional (this code is already pretty long).

docs/4.0-migration-guide/pake.md
```diff
@@ -1,5 +1,111 @@
...
+psa_destroy_key(shared_key_id);
```
gilles-peskine-arm, line 88: Can this be done immediately, or do you need to wait until the key derivation operation is finished?

(Answer: according to the PSA specification, you must not destroy a key while it's in use in an operation: that would effectively abort the operation. But in our implementation, destroying a key while an operation is in progress is currently undetected.)

gilles-peskine-arm (changes requested):
How to replace `psa_pake_get_implicit_key()` with `psa_pake_get_shared_key()` is not fully clear. LGTM otherwise.

For completeness, I checked the changes from API 1.1beta to 1.2 that are relevant to J-PAKE, and the changes in `psa/crypto_extra.h`, and didn't find anything else to mention.

docs/4.0-migration-guide/pake.md
mpg: Absolutely, that's what I looking for, I must have navigated to the wrong part of the file by accident.

yanesca (approved):
LGTM

PR #494, opened 2025-10-03
Expand mbedtls_to_psa_error to LMS

amjoul01:
## Description

This supports translation between LMS Mbed TLS errors to PSA errors.

## PR checklist

- [x] **changelog** not required because: changes are used internally by Mbed TLS.
- [x] **framework PR** not required.
- [ ] **mbedtls development PR** provided Mbed-TLS/mbedtls#
- [ ] **mbedtls 3.6 PR** provided Mbed-TLS/mbedtls# 
- **tests** not required because: this does not change functionality of LMS.

### DIFF
--- a/core/psa_crypto.c
+++ b/core/psa_crypto.c
@@ -55,6 +55,7 @@
 #include "mbedtls/private/gcm.h"
 #include "mbedtls/private/md5.h"
 #include "mbedtls/pk.h"
+#include "mbedtls/lms.h"
 #if defined(MBEDTLS_PK_HAVE_PRIVATE_HEADER)
 #include <mbedtls/private/pk_private.h>
 #endif /* MBEDTLS_PK_HAVE_PRIVATE_HEADER */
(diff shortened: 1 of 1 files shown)


core/psa_crypto.c
```diff
@@ -493,6 +494,19 @@ psa_status_t mbedtls_to_psa_error(int ret)
 #endif
 #endif
 
+#if defined(MBEDTLS_LMS_C)
+        case MBEDTLS_ERR_LMS_BAD_INPUT_DATA:
```
gilles-peskine-arm, line 498: `MBEDTLS_ERR_LMS_BAD_INPUT_DATA`, `MBEDTLS_ERR_LMS_BUFFER_TOO_SMALL` and `MBEDTLS_ERR_LMS_ALLOC_FAILED` are now aliases for the corresponding PSA errors, so we shouldn't attempt to translate them here.

core/psa_crypto.c
```diff
@@ -493,6 +494,19 @@ psa_status_t mbedtls_to_psa_error(int ret)
...
+            return PSA_ERROR_INVALID_ARGUMENT;
+        case MBEDTLS_ERR_LMS_OUT_OF_PRIVATE_KEYS:
```
gilles-peskine-arm, line 500: “Out of private keys” is a rather distinctive error. I'd rather not unify it with a PSA error that only vaguely matches it, in case PSA ends up standardizing LMS signatures but picking a different error code. So it should remain untranslated.

In any case, it won't be raised by PSA functions, since PSA would only provide LMS verification, not LMS signing.

core/psa_crypto.c
```diff
@@ -493,6 +494,19 @@ psa_status_t mbedtls_to_psa_error(int ret)
...
+            return PSA_ERROR_INSUFFICIENT_DATA;
+        case MBEDTLS_ERR_LMS_VERIFY_FAILED:
```
gilles-peskine-arm, line 502: It's unfortunate that `MBEDTLS_ERR_LMS_VERIFY_FAILED` missed the boat when we unified legacy error codes with PSA error codes before the 1.0 release. We didn't commit to have `mbedtls_lms_verify()` return `MBEDTLS_ERR_LMS_VERIFY_FAILED` rather than `PSA_ERROR_INVALID_SIGNATURE`, but application code may rely on it, so I do think that `mbedtls_lms_verify()` should keep returning the legacy error code.

However, what I would prefer to do is to have `lms.c` and `lmots.c` fully use PSA error codes (apart from `MBEDTLS_ERR_LMS_OUT_OF_PRIVATE_KEYS`), rename `mbedtls_lms_verify` to `mbedtls_psa_lms_verify`, and define a new deprecated function `mbedtls_lms_verify` that's just
```
int ret = mbedtls_psa_lms_verify(...);
return ret == PSA_ERROR_INVALID_SIGNATURE ? MBEDTLS_ERR_LMS_VERIFY_FAILED : ret;
```

gilles-peskine-arm (changes requested):
This change is going in the opposite direction from the one we're taking. We want to move towards using PSA errors everywhere applicable, and get rid of the translation functions as much as possible.

Having reviewed the changes here, none of them are actually necessary if we move towards PSA errors inside the LMS module.

gilles-peskine-arm:
Thank you for your contribution! This step isn't going in the right direction, considering that we're moving towards more PSA and less translation. So I'm closing this pull request.

If you'd like to add a PSA interface for LMS, please get in touch with us for architecture and prioritization.

PR #497, opened 2025-10-03
Remove the RSA key mutex

### LINKED ISSUE #4124: Remove the RSA key mutex
RSA key objects include a mutex. `mbedtls_rsa_private` locks the mutex because it caches some auxiliary values used for blinding in the key object. (`mbedtls_rsa_public` also locks the mutex but it seems pointless.) This allows applications to create a key (this must be done in a single-threaded way), then use that key concurrently.

Unfortunately, the presence of the mutex complicates the lifecycle of RSA contexts. For example, part of the library assumes that `mbedtls_rsa_free` can be called twice on the same object, but as I write, this is not the case when `MBEDTLS_THREADING_C` is enabled and the platform does not allow calling `mbedtls_mutex_free` on an already-freed mutex. (There's a fix in https://github.com/ARMmbed/mbedtls/pull/4104, but it's a kludge and might not work in all sensible cases.)

From a high-level architectural perspective, the RSA module is a low-level part of the code dedicated to peforming calculations. Managing concurrency is outside its scope. A better place to protect against concurrent access would be in the PSA dispatch code, [where it is mandated by the PSA specification (as is concurrent key management) but not yet implemented in Mbed TLS](https://github.com/ARMmbed/mbedtls/issues/3263).

ECC contexts do not have a mutex, even though they would need one: some values are cached in the `mbedtls_ecp_group` object to accelerate multiplication. So a multithreaded application that works with RSA keys can't easily be changed to ECC keys: without extra precautions, it would be prone to race conditions.

Since mutexes don't really belong, complicate maintenance, and their presence is potentially misleading, I propose to remove mutexes from RSA keys in Mbed TLS 3.0.

Mailing list discussion: TBA

valeriosetti:
## Description

Resolves https://github.com/Mbed-TLS/mbedtls/issues/4124

## PR checklist

- [x] **changelog** not required because: RSA module was internal and already mutexed by the PSA interface so nothing changes from user point of view.
- **tests**  not required

### DIFF
--- a/drivers/builtin/include/mbedtls/private/rsa.h
+++ b/drivers/builtin/include/mbedtls/private/rsa.h
@@ -22,10 +22,6 @@
 #include "mbedtls/private/bignum.h"
 #include "mbedtls/md.h"
 
-#if defined(MBEDTLS_THREADING_C)
-#include "mbedtls/threading.h"
-#endif
-
 /*
  * RSA Error codes
  */
@@ -111,10 +107,6 @@ typedef struct mbedtls_rsa_context {
                                                     as specified in md.h for use in the MGF
                                                     mask generating function used in the
                                                     EME-OAEP and EMSA-PSS encodings. */
-#if defined(MBEDTLS_THREADING_C)
-    /* Invariant: the mutex is initialized iff ver != 0. */
-    mbedtls_threading_mutex_t MBEDTLS_PRIVATE(mutex);    /*!<  Thread-safety mutex. */
-#endif
 }
 mbedtls_rsa_context;
 
--- a/drivers/builtin/include/mbedtls/private/rsa.h
+++ b/drivers/builtin/include/mbedtls/private/rsa.h
@@ -75,10 +75,6 @@ extern "C" {
  * \brief   The RSA context structure.
  */
 typedef struct mbedtls_rsa_context {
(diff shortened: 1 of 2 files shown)


bjwtaylor (commented):
Do we still need the #include "mbedtls/threading.h" declaration in rsa.c. As far as I can tell you have removed everything that uses it, or is there something I've missed? Not sure if you've seen them and I've not investigated them at all, however we still have !defined(MBEDTLS_THREADING_C) in tests/suites/test_suite_rsa.function. If rsa is not handling threading do we still need them?

valeriosetti:
> the declaration I think we may be able to remove is at the top of rsa.h

Oh sorry, that's actually my fault, I don't know why but I didn't thought about checking the `rsa.h` together with `rsa.c`. Anyway it's solved now and since this PR has a single commit and this change was very minimal, I just updated that single commit and force pushed the PR.

bjwtaylor (dismissed):

mpg (changes requested):
Looks good to me but I think we can go one step further and remove the `ver` field from the context that apparently we only (ab)used for the mutex. (At least I tried removing it and the library still built.)

mpg (approved):
LGTM, thanks!

bjwtaylor (approved):

PR #498, opened 2025-10-06
PK: fix config dependencies and simplify guards

### LINKED ISSUE #490: PK: fix config dependencies and simplify guards
The PK module depends on:
- `MBEDTLS_PSA_CRYPTO_CLIENT` enabled -> this is currently not documented nor enforced in `tf_psa_crypto_check_config.h`, it should be
- at least one of RSA or ECC enabled -> this is currently represented as `RSA_C || PSA_WANT_KEY_TYPE_ECC_PUBLIC_KEY` which is wrong, we should use a `PSA_WANT` for RSA too.

Once this is fixed in `core/tf_psa_crypto_check_config.h` (checks) and `include/psa/crypto_config.h` (documentation) we can clean up code in the pk module (`pk*.h` and `pk*.c` files):
- Remove all `#if defined(MBEDTLS_PSA_CRYPTO_CLIENT)` because that's always true.
- Remove all `#if defined(PSA_WANT_KEY_TYPE_RSA_PUBLIC_KEY) || defined(PSA_WANT_KEY_TYPE_ECC_PUBLIC_KEY)` for the same reason.

valeriosetti:
## Description

Resolves #490.

## PR checklist

- [x] **changelog** not required because: I don't think it's required. With this PR we basically making explicit some dependencies that were already in place also before, so nothing new is happening for which the end user should be warned.
- **tests** not required because: no code change to be tested

### DIFF
--- a/include/mbedtls/pk.h
+++ b/include/mbedtls/pk.h
@@ -156,7 +156,6 @@ typedef struct mbedtls_pk_context {
      * Other keys still use the pk_ctx to store their own context. */
     mbedtls_svc_key_id_t MBEDTLS_PRIVATE(priv_id);
 
-#if defined(PSA_WANT_KEY_TYPE_RSA_PUBLIC_KEY) || defined(PSA_WANT_KEY_TYPE_ECC_PUBLIC_KEY)
     /* Public EC or RSA key in raw format, where raw here means the format returned
      * by psa_export_public_key(). */
     uint8_t MBEDTLS_PRIVATE(pub_raw)[MBEDTLS_PK_MAX_PUBKEY_RAW_LEN];
@@ -166,7 +165,6 @@ typedef struct mbedtls_pk_context {
 
     /* Bits of the private/public key. */
     size_t MBEDTLS_PRIVATE(bits);
-#endif /* PSA_WANT_KEY_TYPE_RSA_PUBLIC_KEY || PSA_WANT_KEY_TYPE_ECC_PUBLIC_KEY */
 
 #if defined(PSA_WANT_KEY_TYPE_ECC_PUBLIC_KEY)
     /* EC family. Only applies to EC keys. */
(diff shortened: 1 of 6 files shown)


include/mbedtls/pk.h
bjwtaylor, line 1: Is there a reason the additional MBEDTLS_PSA_CRYPTO_CLIENT cannot be removed as follows:
```
diff --git a/include/mbedtls/pk.h b/include/mbedtls/pk.h
index a7aef3c55..412d11b97 100644
--- a/include/mbedtls/pk.h
+++ b/include/mbedtls/pk.h
@@ -19,9 +19,7 @@
 
 #include "mbedtls/md.h"
 
-#if defined(MBEDTLS_PSA_CRYPTO_CLIENT)
 #include "psa/crypto.h"
-#endif
 
 /** Type mismatch, eg attempt to do ECDSA with an RSA key */
 #define MBEDTLS_ERR_PK_TYPE_MISMATCH       -0x3F00
@@ -340,7 +338,6 @@ size_t mbedtls_pk_get_bitlen(const mbedtls_pk_context *ctx);
 int mbedtls_pk_can_do_psa(const mbedtls_pk_context *pk, psa_algorithm_t alg,
                           psa_key_usage_t usage);
 
-#if defined(MBEDTLS_PSA_CRYPTO_CLIENT)
 /**
  * \brief           Determine valid PSA attributes that can be used to
  *                  import a key into PSA.
@@ -535,7 +532,6 @@ int mbedtls_pk_copy_from_psa(mbedtls_svc_key_id_t key_id, mbedtls_pk_context *pk
  *                  parameters are not correct.
  */
 int mbedtls_pk_copy_public_from_psa(mbedtls_svc_key_id_t key_id, mbedtls_pk_context *pk);
-#endif /* MBEDTLS_PSA_CRYPTO_CLIENT */
 
 /**
  * \brief           Verify signature.
```

bjwtaylor (approved):

mpg (approved):
LGTM, thanks!

PR #531, opened 2025-10-17
cmake: Add TF_PSA_CRYPTO_CONFIG_NAME option

ronald-cron-arm:
## Description
Add TF_PSA_CRYPTO_CONFIG_NAME CMake option to set one of the named configuration defined in config.py.

## PR checklist
- [x] **changelog** not required (config names are only useful to maintainers)
- [ ] **mbedtls development PR** provided Mbed-TLS/mbedtls# TODO but not necessary to merge this PR
- [ ] **mbedtls 3.6 PR** provided Mbed-TLS/mbedtls# | not required, not sure if we should add this functionality in 3.6 and how different to do it would be
- **tests**  provided

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -63,7 +63,6 @@ if(NOT DEFINED TF_PSA_CRYPTO_AS_SUBPROJECT)
 endif()
 
 # Set the project, Mbed TLS and framework root directory.
-set(TF_PSA_CRYPTO_DIR ${CMAKE_CURRENT_SOURCE_DIR})
 set(TF_PSA_CRYPTO_FRAMEWORK_DIR ${CMAKE_CURRENT_SOURCE_DIR}/framework)
 
 # Put the version numbers into relevant files
@@ -72,7 +71,7 @@ set(version_number_files
         doxygen/tfpsacrypto.doxyfile)
 foreach(file ${version_number_files})
     configure_file(${file}.in
-                   ${TF_PSA_CRYPTO_DIR}/${file})
+                   ${PROJECT_SOURCE_DIR}/${file})
 endforeach(file)
 
 ADD_CUSTOM_TARGET(${TF_PSA_CRYPTO_TARGET_PREFIX}tfpsacrypto-apidoc
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -155,6 +155,15 @@ endif()
 # Make TF_PSA_CRYPTO_CONFIG_FILE and TF_PSA_CRYPTO_USER_CONFIG_FILE into PATHs
 set(TF_PSA_CRYPTO_CONFIG_FILE "" CACHE FILEPATH "TF-PSA-Crypto config file (overrides default).")
 set(TF_PSA_CRYPTO_USER_CONFIG_FILE "" CACHE FILEPATH "TF-PSA-Crypto user config file (appended to default).")
+set(TF_PSA_CRYPTO_CONFIG_NAME "" CACHE STRING "Name of the configuration to use (see config.py documentation)")
+
+set(TF_PSA_CRYPTO_crypto_config_path ${PROJECT_SOURCE_DIR}/include/psa/crypto_config.h)
+if(NOT "${TF_PSA_CRYPTO_CONFIG_NAME}" STREQUAL "")
+    if(NOT "${TF_PSA_CRYPTO_CONFIG_FILE}" STREQUAL "")
+        message(FATAL_ERROR "TF_PSA_CRYPTO_CONFIG_NAME and TF_PSA_CRYPTO_CONFIG_FILE are incompatible.")
+    endif()
+    set(TF_PSA_CRYPTO_crypto_config_path ${PROJECT_BINARY_DIR}/include/psa/crypto_config.h)
+endif(NOT "${TF_PSA_CRYPTO_CONFIG_NAME}" STREQUAL "")
 
 # Create a symbolic link from ${base_name} in the binary directory
 # to the corresponding path in the source directory.
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -157,12 +157,10 @@ set(TF_PSA_CRYPTO_CONFIG_FILE "" CACHE FILEPATH "TF-PSA-Crypto config file (over
 set(TF_PSA_CRYPTO_USER_CONFIG_FILE "" CACHE FILEPATH "TF-PSA-Crypto user config file (appended to default).")
 set(TF_PSA_CRYPTO_CONFIG_NAME "" CACHE STRING "Name of the configuration to use (see config.py documentation)")
 
-set(TF_PSA_CRYPTO_crypto_config_path ${PROJECT_SOURCE_DIR}/include/psa/crypto_config.h)
 if(NOT "${TF_PSA_CRYPTO_CONFIG_NAME}" STREQUAL "")
     if(NOT "${TF_PSA_CRYPTO_CONFIG_FILE}" STREQUAL "")
         message(FATAL_ERROR "TF_PSA_CRYPTO_CONFIG_NAME and TF_PSA_CRYPTO_CONFIG_FILE are incompatible.")
     endif()
-    set(TF_PSA_CRYPTO_crypto_config_path ${PROJECT_BINARY_DIR}/include/psa/crypto_config.h)
 endif(NOT "${TF_PSA_CRYPTO_CONFIG_NAME}" STREQUAL "")
 
 # Create a symbolic link from ${base_name} in the binary directory
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -426,6 +426,9 @@ if(ENABLE_TESTING OR ENABLE_PROGRAMS)
                 ${TF_PSA_CRYPTO_FRAMEWORK_DIR}/tests/include
                 tests/include
                 tests/include/test
+                # Add the build-tree include directory before the source-tree one
+                # so that generated headers in the build tree take precedence.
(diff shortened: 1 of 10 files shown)


CMakeLists.txt
```diff
@@ -63,7 +63,6 @@ if(NOT DEFINED TF_PSA_CRYPTO_AS_SUBPROJECT)
 endif()
 
 # Set the project, Mbed TLS and framework root directory.
-set(TF_PSA_CRYPTO_DIR ${CMAKE_CURRENT_SOURCE_DIR})
```
gilles-peskine-arm, line 66: `programs/test/cmake_subproject/CMakeLists.txt` sets `TF_PSA_CRYPTO_DIR`. I don't know how that works: is it an independent variable, or did it override the value in the subproject when it included the subproject?

include/CMakeLists.txt
```diff
@@ -1,10 +1,44 @@
 option(INSTALL_TF_PSA_CRYPTO_HEADERS "Install TF PSA Crypto headers." ON)
 
+# The handling of TF_PSA_CRYPTO_CONFIG_NAME assumes that the default
+# configuration file 'include/psa/crypto_config.h' exists in the source tree.
+# That file is copied into the build tree and then modified with 'config.py'
+# according to the configuration name provided via TF_PSA_CRYPTO_CONFIG_NAME.
+set(TF_PSA_CRYPTO_CONFIG_NAME "" CACHE STRING "Name of the configuration to use (see config.py documentation)")
+
+if(TF_PSA_CRYPTO_CONFIG_NAME)
```
gilles-peskine-arm: Minor: this would take the wrong branch if we had a config name that [cmake](https://cmake.org/cmake/help/v3.5/command/if.html) considers to be a false constant, such as `off` or `file-notfound`. Same with the config file name (fortunately no false constant ends with `.h`). I think this should be

```suggestion
if(NOT "${TF_PSA_CRYPTO_CONFIG_NAME}" STREQUAL "")
```

include/CMakeLists.txt
```diff
@@ -1,10 +1,44 @@
...
+    if(TF_PSA_CRYPTO_CONFIG_FILE)
+        message(FATAL_ERROR "TF_PSA_CRYPTO_CONFIG_NAME and TF_PSA_CRYPTO_CONFIG_FILE are incompatible.")
+    endif()
+    file(MAKE_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/psa)
+    configure_file(${PROJECT_SOURCE_DIR}/include/psa/crypto_config.h
```
gilles-peskine-arm: If I understand correctly, `crypto_config.h` in the build tree is created when you run `cmake`, and won't be updated if the file changes in the source tree, but will be re-created if you re-run `cmake`?

(I think this behavior is fine, I just want to check my understanding.)

gilles-peskine-arm (commented):
Mostly looks good to me. I have a few doubts that need clarification before I can approve.

CMakeLists.txt
ronald-cron-arm: In `programs/test/cmake_subproject/CMakeLists.txt`, `TF_PSA_CRYPTO_DIR` is a custom variable containing the path to the external TF-PSA-Crypto project, not a custom variable containing the path to the root of the current project.

include/CMakeLists.txt
ronald-cron-arm: It was not the case in the version you read. It is the case now. I've replaced `configure_file` with a simple file copy.
The configuration file in the build tree is updated only when your configure or re-configure the build tree, not if crypto_config.h is modified in the source tree.
Then if you have configured using the TF_PSA_CRYPTO_CONFIG_NAME option, if you modify crypto_config.h in the source tree, this has no impact on the build of the libraries/tests/programs. If you modify crypto_config.h in the build tree, this impacts the build/re-build of the libraries/tests/programs as expected.

tests/CMakeLists.txt
```diff
@@ -153,11 +153,10 @@ if(GEN_FILES)
             ${TF_PSA_CRYPTO_FRAMEWORK_DIR}/scripts/mbedtls_framework/psa_test_case.py
             ${TF_PSA_CRYPTO_FRAMEWORK_DIR}/scripts/mbedtls_framework/test_case.py
             ${TF_PSA_CRYPTO_FRAMEWORK_DIR}/scripts/mbedtls_framework/test_data_generation.py
-            ${CMAKE_CURRENT_SOURCE_DIR}/../include/psa/crypto_config.h
-            ${CMAKE_CURRENT_SOURCE_DIR}/../include/psa/crypto_values.h
-            ${CMAKE_CURRENT_SOURCE_DIR}/../include/psa/crypto_extra.h
+            ${TF_PSA_CRYPTO_crypto_config_path}
```
gilles-peskine-arm: The generated file is independent of the configuration, so the python code should read to the source file, not the modified file in the build tree. And as far as I can tell, they do (they look for files in `include` relative to the source tree, without using different ways to look up e.g. `crypto_values.h` vs `crypto_config.h`).

```suggestion
            ${PROJECT_SOURCE_DIR}/include/psa/crypto_config.h
```

tests/scripts/components-build-system.sh
```diff
@@ -137,6 +137,33 @@ component_tf_psa_crypto_build_custom_config_file () {
...
+
+    msg "build: cmake out-of-source with full config"
+    cd "$OUT_OF_SOURCE_DIR"
+    cmake -DCMAKE_INSTALL_PREFIX="$OUT_OF_SOURCE_DIR/install_full" -DTF_PSA_CRYPTO_CONFIG_NAME=full "$TF_PSA_CRYPTO_ROOT_DIR"
+    cd "$TF_PSA_CRYPTO_ROOT_DIR"
+    echo '#error "cmake -DTF_PSA_CRYPTO_CONFIG_NAME=full does not work"' > "$CRYPTO_CONFIG_H"
+    cmake --build "$OUT_OF_SOURCE_DIR" --target tfpsacrypto
+
+    # Restore the default config file in the source tree as we need a sane
+    # one to build the tests.
```
gilles-peskine-arm: Why does building the tests consume the config file in the source? Or do you mean to generate test data (and `psa_constant_names` IIRC)?

gilles-peskine-arm (changes requested):
LGTM except for one dependency.

tests/CMakeLists.txt
ronald-cron-arm: Yes good point, thanks.

tests/scripts/components-build-system.sh
ronald-cron-arm: Yes, I made the comment more specific.

gilles-peskine-arm (dismissed):
LGTM

tests/scripts/components-build-system.sh
minosgalanakis: Is that required because the `generate_psa_constants` is set as a dependency?

minosgalanakis (commented):
Looks good overall but I am not quite clear why resetting the default config is needed before invoking

~~~~~
   cmake --build "$OUT_OF_SOURCE_DIR"
    cmake --build "$OUT_OF_SOURCE_DIR" --target install
~~~~~

tests/scripts/components-build-system.sh
ronald-cron-arm: `generate_psa_constants.py` compiles C code that includes the file `crypto_config.h` of the source tree. Thus, if it contains "#error" the compilation fails.

ronald-cron-arm:
> Looks good overall but I am not quite clear why resetting the default config is needed before invoking
> 
> ```
>    cmake --build "$OUT_OF_SOURCE_DIR"
>     cmake --build "$OUT_OF_SOURCE_DIR" --target install
> ```

see https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/531#discussion_r2455877830

minosgalanakis (dismissed):
LGTM

gilles-peskine-arm:
Unfortunately there's a merge conflict now.

ronald-cron-arm:
@minosgalanakis @gilles-peskine-arm I've rebased on top current development f892976a3df124bca7f32126b049e77ed95f422c. Mostly overlapping changes with https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/587 but for the installation of the build tree (and not source tree) crypto_config.h file. Here the code has completely changed. To be explicit I am referring to the following code in `include/CMakeLists.txt`
```
if(INSTALL_TF_PSA_CRYPTO_HEADERS)
    install(DIRECTORY "."
        DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}
        FILES_MATCHING PATTERN "*.h"
        PATTERN "crypto_config.h" EXCLUDE
    )

    if("${TF_PSA_CRYPTO_CONFIG_NAME}" STREQUAL "")
        install(FILES "${CMAKE_CURRENT_SOURCE_DIR}/psa/crypto_config.h"
            DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}/psa)
    else()
        install(FILES "${CMAKE_CURRENT_BINARY_DIR}/psa/crypto_config.h"
            DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}/psa)
    endif("${TF_PSA_CRYPTO_CONFIG_NAME}" STREQUAL "")
endif(INSTALL_TF_PSA_CRYPTO_HEADERS)
```

gilles-peskine-arm (approved):
Thanks for updating this, LGTM

minosgalanakis (approved):
LGTM

gilles-peskine-arm:
It would be nice to have this in 3.6 as well, but this is a new feature so it doesn't need to be backported. So I'll go ahead and merge now, before there's another conflict.

PR #543, opened 2025-10-22
Integrate ML-DSA-87 from BoringSSL

gilles-peskine-arm:
Proof-of-concept integration of ML-DSA-87 signature and verification. This is the BoringSSL implementation, kindly contributed by @lukaszobernig in https://github.com/Mbed-TLS/mbedtls/pull/10463.

Uses the SHAKE implementation kindly contributed by @polhenarejos in https://github.com/Mbed-TLS/mbedtls/pull/5821.

Status: for internal evaluation and CI. We are evaluating which ML-DSA implementation to integrate, this is one of the candidates. This pull request will not be merged as is: even if we decide to use this ML-DSA implementation, I will break out and rework the SHAKE part, and you can expect significant history rewrites.

## PR checklist

- [x] **changelog** no (no user-visible feature)
- [x] **mbedtls not required because: crypto only
- **tests**  basic


drivers/builtin/src/mldsa87.c
```diff
@@ -0,0 +1,1231 @@
...
+    }
+  }
+
+  mbedtls_sha3_free(&shake128_ctxt);
+}
+
+static int coefficient_from_nibble_2(uint32_t nibble, uint32_t *result)
+{
+  if (nibble < 15) {
+    *result = mod_sub(2, nibble % 5);
```
lukaszobernig, line 677: We found that this should be replaced with
```
// Constant time "nibble % 5".
nibble = nibble - 5 * ((205 * nibble) >> 10);
*result = mod_sub(2, nibble);
```
to better ensure that compilers do not generate potentially non-constant time division operations.

gilles-peskine-arm:
After some more investigation, and given the further work on memory optimization in mldsa-native, we've decided to take ML-DSA from mldsa-native rather than from BoringSSL. (We don't have the capacity to maintain both options.)

Thank you very much for offering!

PR #552, opened 2025-10-28
Fix non-portable initializer for NEON vectors

### LINKED ISSUE #551: chacha20_neon.c compilation fails in VS due to incorrect number of initializers
https://github.com/Mbed-TLS/TF-PSA-Crypto/blob/adf05e8581147c8c6537374f37ac9b952231be7a/drivers/builtin/src/chacha20_neon.c#L69
This and similar line just below,  would not compile neither in VS 2017 nor in VS 2022.
The code should be:
`uint32x4_t counter_increment = { .n128_i32 = {0, 0, 0, 1} };`

Additional question: Is NEON always present for Windows on ARM64 - the same way as there could be no AMD64 without SSE2?

gilles-peskine-arm:
Fixes #551.

## PR checklist

- [x] **changelog** not required because: not released yet (came after 1.0.0)
- [x] **mbedtls 3.6 PR** not required because: new after 1.0
- **tests**  provided (and I tested locally for big-endian aarch64)

### DIFF
--- a/drivers/builtin/src/chacha20_neon.c
+++ b/drivers/builtin/src/chacha20_neon.c
@@ -65,13 +65,9 @@ static inline uint32x4_t chacha20_neon_vrotlq_7_u32(uint32x4_t v)
 // Increment the 32-bit element within v that corresponds to the ChaCha20 counter
 static inline uint32x4_t chacha20_neon_inc_counter(uint32x4_t v)
 {
-    if (MBEDTLS_IS_BIG_ENDIAN) {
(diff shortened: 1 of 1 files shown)


daverodgman (approved):
LGTM. Verified via Godbolt that (at least for clang) this fix appears to result in no change to the  generated code for both big and little endian. So no implications for perf.

valeriosetti (approved):

davidhorstmann-arm (approved):
LGTM, thanks!

PR #554, opened 2025-10-29
Fix duplication of product version in CMakeLists.txt

### LINKED ISSUE #553: Old cmake package version in install content
### Summary
The installed cmake package file (TF-PSA-CryptoConfigVersion.cmake) contains a hardcoded version number (0.1.0) when building v1.0.0 tag. As a result find_package(TF-PSA-Crypto 1.0.0 REQUIRED) fails to find the library.
IMHO the version should match the product version to allow 3rd parties to depend on the correct version. 

(Note: tfpsacrypto.pc has the correct 1.0.0 version.)

### System information

TF-PSA-Crypto version (number or commit id): v1.0.0
Operating system and version: Ubuntu 22.01 (irrelevant)
Configuration (if not default, please attach `crypto_config.h`): irrelevant
Compiler and options (if you used a pre-built binary, please indicate how you obtained it): irrelevant
Additional environment information: none

### Expected behavior

Package version should match product version. find_package(TF-PSA-Crypto 1.0.0) should find the cmake package when building v1.0.0.

### Actual behavior

find_package(TF-PSA-Crypto 1.0.0 REQUIRED) returns arror as the package version is 0.1.0.

### Steps to reproduce

Create test.cmake file with the following content:
```
cmake_minimum_required(VERSION 3.18)
find_package(TF-PSA-Crypto 0.1.0 REQUIRED)
```

Build TF-PSA-Crypto and install it to INSTALL_DIR.

Run the test.cmake:

```
cmake -P test.cmake -DCMAKE_PREFIX_PATH=$(readlink -f $INSTALL_DIR)
```

### Additional information

Please fix the version number here: https://github.com/Mbed-TLS/TF-PSA-Crypto/blob/development/CMakeLists.txt#L467

Updating this line should be part of the release process.

gilles-peskine-arm:
Fix #553

## PR checklist

- [x] **changelog** provided
- [x] **development PR** https://github.com/Mbed-TLS/mbedtls/pull/10481
- [x] **TF-PSA-Crypto PR** here
- [x] **3.6 PR** not required because: in 3.6, `bump_version.sh` catches all occurrences of the version (although we could backport this if we want more consistency, and we should probably backport this if we move 3.6 to the new release script)
- **tests**  not required because: it's now automatic, so I don't think it's worth testing

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -464,7 +464,7 @@ if(NOT DISABLE_PACKAGE_CONFIG_AND_INSTALL)
     write_basic_package_version_file(
         "cmake/TF-PSA-CryptoConfigVersion.cmake"
             COMPATIBILITY SameMajorVersion
-            VERSION 0.1.0)
+            VERSION "${TF_PSA_CRYPTO_VERSION}")
 
     install(
         FILES "${CMAKE_CURRENT_BINARY_DIR}/cmake/TF-PSA-CryptoConfig.cmake"
--- a/ChangeLog.d/cmake-package-version.txt
+++ b/ChangeLog.d/cmake-package-version.txt
@@ -1,0 +1,3 @@
+Bugfix
+   * Fix CMake package version that was inconsistent with the product version.
+     Fixes #553.


CMakeLists.txt
```diff
@@ -464,7 +464,7 @@ if(NOT DISABLE_PACKAGE_CONFIG_AND_INSTALL)
     write_basic_package_version_file(
         "cmake/TF-PSA-CryptoConfigVersion.cmake"
             COMPATIBILITY SameMajorVersion
-            VERSION 0.1.0)
+            VERSION "${TF_PSA_CRYPTO_VERSION}")
```
gyuri-szing, line 467: `PROJECT_VERSION` is set to `TF_PSA_CRYPTO_VERSION` in the top level CMakeLists.txt by the `project(...)` command. IMHO for clarity both `pkgconfig/tfpsacrypto.pc.in` and `write_basic_package_version_file(..)` in this file should use the same variable.

Just as an FYI:
In TS we use a .txt file to hold the product version. This is read by the documentation (Sphinx) and cmake, and is the single source of truth for the version info. See:
- [CMakeLists.txt](https://review.trustedfirmware.org/plugins/gitiles/TS/trusted-services/+/refs/heads/main/deployments/libsp/opteesp/CMakeLists.txt#16)
- [TsGetVersion.cmake](https://review.trustedfirmware.org/plugins/gitiles/TS/trusted-services/+/refs/heads/main/tools/cmake/common/TsGetVersion.cmake)
- [version.txt](https://review.trustedfirmware.org/plugins/gitiles/TS/trusted-services/+/refs/heads/main/version.txt)
- [conf.py for Sphinx](https://review.trustedfirmware.org/plugins/gitiles/TS/trusted-services/+/refs/heads/main/docs/conf.py#23)

Benefits:
- Simple version updates when making releases.
- version.txt survives if the project is mirrored to another VCS or is downloaded as a .tgz. Git tags would not. We use tags too to give easy access to a specific release with git.

gilles-peskine-arm: Historically, Mbed TLS tried to avoid reliance on specific build tools, so the version was in many different places. In TF-PSA-Crypto, CMake is the only supported build system, and we're down to three places: `CMakeLists.txt`, a C header, and a test file for the get-version API.

bjwtaylor (approved):

valeriosetti (approved):

PR #555, opened 2025-10-29
Test release preparation

### LINKED ISSUE #9521: Generated files should not vary based on build path
We have a number of files that are not committed into git (except for releases) but instead generated by `make generated_files`. The intent is for the generation procedure to [be reproducible](https://github.com/Mbed-TLS/mbedtls-restricted/pull/1286#discussion_r1735790816): two people on different machines should generate the same files.

This is currently not the case for at least one reason: sometimes the generated files include the absolute path to some files (that we used for generated them, and are referenced in comments) - this [happens](https://github.com/Mbed-TLS/mbedtls-restricted/pull/1286#discussion_r1735742870) with `test_certs.h` for example.

This task is to make sure that generated files don't include the build path, or otherwise have content that varies depending on the build path:
- [ ] Add a test to `all.sh` about it (most probably, by extending `component_check_generated_files()`. For example, one could make a copy of the source tree in a subdirectory, generate files in that directory, then rename it, then run `check-generated-files.sh`, which should complain if the files have different contents after the renaming. This test should find (at least) an issue with `test_certs.h`.
- [ ] Fix the issue with `test_certs.h`, and any other issue found by the above test.

Note about scope: this task is not about ensuring that generation is fully reproducible, it's only focusing on the specific issue of embedded paths.

gilles-peskine-arm:
Addresses most of https://github.com/Mbed-TLS/mbedtls/issues/3255 . What is left is to arrange for the Groovy code to pick up the release artifacts created by the new `all.sh` component as Jenkins artifacts.

Addresses most of https://github.com/Mbed-TLS/mbedtls-framework/pull/231 by creating `bump_version.sh` in crypto. We might want to put this functionality into `prepare_release.py`, but I'm not fully convinced: there are significant differences between branches with respect to what version bumping does, so I think the code to do that belongs in each branch.

Fixes https://github.com/Mbed-TLS/mbedtls/issues/9521 by testing that builds done in different directories result in identical tarballs.

Needs preceding PR:

* [ ] https://github.com/Mbed-TLS/mbedtls-framework/pull/231
* [x] Not needed to pass the CI, but needed for `bump_version.sh` not to be incomplete: https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/554
* [x] The `bump_version.sh` commits are in https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/676
* [x]: needs Ubuntu 24.04 on the CI https://github.com/Mbed-TLS/mbedtls-test/pull/230

## PR checklist

- [x] **framework PR** provided https://github.com/Mbed-TLS/mbedtls-framework/pull/231
- [ ] **mbedtls development PR** TODO
- [ ] **mbedtls 4.1 PR** TODO
- [x] **TF-PSA-Crypto development PR** https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/555
- [ ] **TF-PSA-Crypto 1.1 PR** TODO
- [x] **mbedtls 3.6 PR** No, we'll keep the old process for the remaining life time of 3.6
- **tests**  provided

### DIFF
--- a/tests/scripts/components-release.sh
+++ b/tests/scripts/components-release.sh
@@ -39,6 +39,12 @@ git_clone_recursively () {
     fi
 }
 
+support_test_prepare_release () {
+    # `prepare_release.py` uses `git archive --mtime` which was added
+    # in Git 2.40.0.
+    git archive --help | grep -q -e --mtime
+}
+
 component_test_prepare_release () {
     msg "Prepare release: Set up testing environment"
 
--- a/tests/scripts/components-release.sh
+++ b/tests/scripts/components-release.sh
@@ -111,6 +111,18 @@ component_test_prepare_release () {
     git fetch --no-write-fetch-head "$preparation_dir" "$preparation_sha"
     echo ">>>> Release candidate sha: $preparation_sha <<<<"
 
+    msg "Prepare release: checking archive reproducibility"
+    mkdir "${preparation_dir}2" "${artifacts_dir}2"
+    cd "${preparation_dir}2"
+    git_clone_recursively "$preparation_dir"
+    # Starting from the release candidate commit, prepare_release.py should
+    # not make any extra commit.
+    # Starting from the same commit, prepare_relase.py should create
+    # identical archives.
+    framework/scripts/prepare_release.py --artifact-directory "${artifacts_dir}2" -r "$new_version"
+    ls -l "${artifacts_dir}2"
+    diff "$shasum_file" "${artifacts_dir}2/${shasum_file##*/}"
+
     msg "Prepare release: Test the release archive"
     shasum_file=$artifacts_dir/tf-psa-crypto-$new_version.txt
     archive_file=$artifacts_dir/tf-psa-crypto-$new_version.tar.bz2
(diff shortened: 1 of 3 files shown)


tests/scripts/components-release.sh
```diff
@@ -0,0 +1,155 @@
...
+# Copyright The Mbed TLS Contributors
+# SPDX-License-Identifier: Apache-2.0 OR GPL-2.0-or-later
+
+# This file contains test components that are executed by all.sh
+
+################################################################
+#### Release preparation
+################################################################
+
+next_product_version () {
```
gilles-peskine-arm, line 12: I'm not sure if the auxiliary functions here should move to `framework/scripts/all-helpers.sh`. They're pretty specific to release preparation.

tests/scripts/components-release.sh
```diff
@@ -0,0 +1,155 @@
...
+        END {
+            if (!$found) {
+                print STDERR "Product version not found in CMakeLists.txt\n";
+                exit 1;
+            }
+        }
+EOF
+}
+
+git_clone_recursively () {
```
gilles-peskine-arm, line 28: I want to tidy up the handling of submodules (both cloning, and the dirty worktree check) before copying this to mbedtls. For crypto with just one submodule, it's ok. But for the multiple submodules of Mbed TLS, I'd like a loop. And if I write that loop I might as well use it here too.

minosgalanakis:
@gilles-peskine-arm would it be possible to split this pr and add the bump-version.sh to TF-PSA-Crypto separately? it can come in as a standalone script?

gilles-peskine-arm:
@minosgalanakis Done: https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/676

PR #572, opened 2025-11-12
Add new test components with different combinations of PKPARSE and PKWRITE

### LINKED ISSUE #525: Add new test components with different combinations of PKPARSE and PKWRITE
### Suggested enhancement

1. Add test components similar to `component_full_no_pkparse_pkwrite` in `components-configuration-crypto.sh` in Mbed TLS repo, but with:
    - pkparse and !pkwrite
    - !pkparse and pkwrite

2. Clone `component_full_no_pkparse_pkwrite` also in tf-psa-crypto, but leave also one in mbedtls (X.509 and TLS use PK, so it might be useful to see what happens if PK support changes).

The case where both pkparse and pkwrite are set is already included in the many `full` configurations being tested in the CI, so nothing needs to be added for it.

See [here](https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/496#issuecomment-3409646133) and [here](https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/496#issuecomment-3400703738) for some background on how we got to this request.

### Justification

TF-PSA-Crypto needs this because it increases test coverage of possible build configurations.

valeriosetti:
## Description

Resolves #525

## PR checklist

- [x] **changelog** not required because: only tests components being added
- **tests**  provided

### DIFF
--- a/tests/scripts/components-configuration.sh
+++ b/tests/scripts/components-configuration.sh
@@ -259,12 +259,17 @@ common_tf_psa_crypto_full_pkparse_pkwrite () {
     cmake -DCMAKE_BUILD_TYPE:String=Asan "$TF_PSA_CRYPTO_ROOT_DIR"
     make
 
-    # Ensure that PK_PARSE_C and/or PK_PARSE_C were not re-enabled accidentally (additive config).
+    # Ensure that PK_PARSE_C and/or PK_PARSE_C are correctly disabled or enabled
+    # depending on the current scenario.
     if [ $PK_PARSE -eq 0 ]; then
         not grep mbedtls_pk_parse_key "core/libtfpsacrypto.a"
+    else
+        grep mbedtls_pk_parse_key "core/libtfpsacrypto.a"
     fi
     if [ $PK_WRITE -eq 0 ]; then
         not grep mbedtls_pk_write_key_der "core/libtfpsacrypto.a"
+    else
+        grep mbedtls_pk_write_key_der "core/libtfpsacrypto.a"
     fi
 
     msg "test: $message"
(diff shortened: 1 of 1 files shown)


bjwtaylor (dismissed):

tests/scripts/components-configuration.sh
```diff
@@ -232,3 +232,53 @@ component_test_psa_assume_exclusive_buffers_valgrind_cf () {
...
+
+    message="full without"
+    if [ $PK_PARSE -eq 0 ]; then
+        message="$message pkparse"
+    fi
+    if [ $PK_WRITE -eq 0 ]; then
+        message="$message pkwrite"
+    fi
+
+    msg "build:$message"
```
mpg: Cosmetic: `build: $message` (with a space). (Same for `test:` below.)

tests/scripts/components-configuration.sh
```diff
@@ -232,3 +232,53 @@ component_test_psa_assume_exclusive_buffers_valgrind_cf () {
...
+    fi
+
+    cd "$OUT_OF_SOURCE_DIR"
+    cmake -DCMAKE_BUILD_TYPE:String=Asan "$TF_PSA_CRYPTO_ROOT_DIR"
+    make
+
+    # Ensure that PK_PARSE_C and/or PK_PARSE_C were not re-enabled accidentally (additive config).
+    if [ $PK_PARSE -eq 0 ]; then
+        not grep mbedtls_pk_parse_key "core/libtfpsacrypto.a"
+    fi
```
mpg, line 268: Robustness: I'd like to have
```
else
    grep mbedtls_pk_parse_key "core/libtfpsacrypto.a"
```
to confirm that the above test (with `not grep`) is not always true in all builds (eg because we removed or renamed this function, or typoed its name here in `all.sh`, or are grepping the wrong `.a`, etc.)

mpg (commented):
Looking pretty good, but I left two suggestions for improvements (one super minor, the other would give me greater confidence in the tests).

mpg (approved):
Thanks for addressing my comments. LGTM assuming the CI agrees.

bjwtaylor (approved):

PR #573, opened 2025-11-13
Use instruction-level parallelism with AES-NI to speed up AES-CTR

mfil:
## Description

One feature of the AES-NI instructions is instruction-level parallelism, which means that, while one round of AES takes multiple clock cycles to complete, another instruction can already start executing on the next clock cycle. Only when the result of the AES round is used do we need to wait for it to finish.

In particular, this makes it possible to encrypt or decrypt multiple AES blocks in parallel which can be used to speed up AES in some modes of operation. In this commit, it is used to speed up AES-CTR. I have chosen to do up to four blocks in parallel. More would be possible, but while testing on my machine, I didn't see meaningful speed-ups when going beyond 4, and I don't want to clutter the source files with even more assembly.

I have removed the `defines` with the `aesenc` etc. opcodes assuming that we don't need to support these older versions of binutils anymore. If that's wrong, I'll put them back (and add my own macros for the registers I need to use).

I would like to go on and also use it in AES-GCM, but there I ran into a problem. GCM runs the underlying block cipher in ECB mode. Due to an inconsistency in the `mbedtls_cipher` API, ECB mode always encrypts exactly one block. As far as I understand, `mbedtls_cipher` is on its way out, so I don't know if you would want to spend time reviewing a fix for this. Please let me know what you think would be the best approach!

An alternative I can see would be to use CTR mode (making sure that the 32-bit counter in GCM doesn't overflow) inside of GCM, if available.

## PR checklist

- [x] **changelog** provided
- [x] **mbedtls development PR** not required because: No changes in mbedtls are needed to make use of this. The speed-up will take effect as soon as the submodule is updated. (Should I make a pull request just to update the submodule?)
- [x] **mbedtls 3.6 PR** not required because: It's a new feature, not a bugfix. It doesn't have to be backported. (As far as I'm concerned.)
- **tests** not required because: No changes in the output are intended, only performance.

### DIFF
--- a/ChangeLog.d/speedup-aesni-ctr.txt
+++ b/ChangeLog.d/speedup-aesni-ctr.txt
@@ -1,0 +1,3 @@
+Features:
+   * Speed up AES-CTR when using AES-NI by using instruction-level
+     parallelism.
--- a/ChangeLog.d/speedup-aesni-ctr.txt
+++ b/ChangeLog.d/speedup-aesni-ctr.txt
@@ -1,3 +1,3 @@
-Features:
+Features
    * Speed up AES-CTR when using AES-NI by using instruction-level
      parallelism.
--- a/drivers/builtin/src/aes.c
+++ b/drivers/builtin/src/aes.c
@@ -873,7 +873,7 @@ int mbedtls_aes_xts_setkey_dec(mbedtls_aes_xts_context *ctx,
 /* In test_aesni, it is checked whether the software AES implementation is built
  * by checking if mbedtls_internal_aes_encrypt exists in aes.o. Therefore, avoid
  * inlining. */
-#define INTERNAL_AES_ENC_ATTRS __attribute__ ((noinline))
+#define INTERNAL_AES_ENC_ATTRS __attribute__((noinline))
 #else
 #define INTERNAL_AES_ENC_ATTRS
 #endif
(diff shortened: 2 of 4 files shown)


mfil:
It says
> [TF OpenCI: Interface stability tests](https://mbedtls.trustedfirmware.org/job/mbed-tls-tf-psa-crypto-multibranch/job/PR-573-merge/5/display/redirect) — Pre-test checks failed.

but when I click the link to view details, I see a green check on archive-timestamps and everything else is blank.

bjwtaylor:
Hi @mfil, thanks for submitting this PR. It looks like that job has failed due to infrastructure issues, so I would ignore them for the time being. I'm trying to get them resolved and will let you know when it's working again.

mfil:
Hi @bjwtaylor ! Thanks, I'll wait.

In the meantime, I found that the functions with the intrinsics may not actually improve performance depending on the compiler. What I *hoped* would happen is that with optimizations turned on, the compiler would unroll the loops and then pipeline the `aesenc` instructions. Clang 18.1.3 does so, GCC 13.3.0 does not.

I tried hand-unrolling the loops, and then I get the speed-ups also with GCC. Should I commit that?

bjwtaylor:
@mfil, the PR will need to go through a full review process. In the meantime though I would add anything you think is useful. I've now resolved some of the CI issues, please ignore the OpenCI issues and any failures in the NewCI should be related to your changes.

mfil:
Ok, I added another commit that unrolls the "4blocks" function.

mfil:
I fixed the error in `tf_psa_crypto_check_changelog`.

I figured out what's going wrong in `test_aesni`, but not what would be a good way to fix it. It checks that the binary `aes.o` contains the string `mbedtls_internal_aes_encrypt`. That function is declared static, and my changes in `aes.c` seem to cause the compiler in your test container to inline it.

For the failure on aarch64, it's not obvious to me what's going wrong. I guess I have to go play with QEMU.

mfil:
After looking at it for the 10th time, I think I spotted the bug with aarch64.

PR #578, opened 2025-11-20
Simplify interruptible ECDH dispatch

### LINKED ISSUE #9719: Optimize interruptible key agreement dispatch through ECDH
PSA interruptible key agreement dispatches has a long dispatch chain:

* interface in `psa_crypto.c`
* driver dispatch
* built-in implementation in `psa_crypto_ecp.c`
* Legacy ECDH in `ecdh.c`
* ECC arithmetic in `ecp.c`

The dispatch through ECDH has [known inefficiencies](https://github.com/Mbed-TLS/mbedtls/pull/9490#discussion_r1759245539) where we're copying data around, but don't need to. The goal of this task is, at a minimum, to optimize this. Furthermore, look for other inefficiencies in the ECDH indirection and analyze how we might optimize those. The answer might be to eliminate ECDH altogether, or to change the ECDH interface. Note that starting in TF-PSA-Crypto 1.0, `ecdh.h` is an unstable or internal interface and we can change it at any time.

mpg:
## Description

Resolves https://github.com/Mbed-TLS/mbedtls/issues/9719

Note: a similar simplification can be applied to non-interruptible ECDH, but this will require some care to handle Everest dispatch (which is currently done in `ecdh.c`), so it's left for a future PR in order to keep this one short and sweet. (Edit: see https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/582)

Note: the tests enhancement could be useful for other operations than ECDH (key generation, import, signature). I'd rather handle that as a separate follow-up issue, again to keep this PR short and sweet.

## PR checklist

- [x] **changelog** provided
- [x] **mbedtls 3.6 PR** not applicable: interruptible ECDH was not present
- **tests**  provided


tests/suites/test_suite_psa_crypto.function
```diff
@@ -9975,9 +9975,14 @@ void key_agreement_interruptible(int alg_arg,
     TEST_EQUAL(num_ops_prior, MBEDTLS_ECP_OPS_CHK);
 
     min_num_ops = interruptible_operations_get_min_num_ops(key_bits);
+    uint32_t exp_max_ops = max_ops <= min_num_ops ? min_num_ops : max_ops;
 
-    do {
+    psa_ecc_family_t fam = PSA_KEY_TYPE_ECC_GET_FAMILY(our_key);
+    int interruptible = PSA_ECC_FAMILY_IS_WEIERSTRASS(fam);
 
+    size_t nb_calls;
```
davidhorstmann-arm:
```suggestion
    size_t nb_calls = 0;
```
As the CI mentions, below an uninitialized value is incremented.

davidhorstmann-arm (commented):
The CI is unhappy, for now

tests/suites/test_suite_psa_crypto.function
```diff
@@ -9975,9 +9975,14 @@ void key_agreement_interruptible(int alg_arg,
     TEST_EQUAL(num_ops_prior, MBEDTLS_ECP_OPS_CHK);
 
     min_num_ops = interruptible_operations_get_min_num_ops(key_bits);
+    uint32_t exp_max_ops = max_ops <= min_num_ops ? min_num_ops : max_ops;
```
davidhorstmann-arm, line 9981: Since you've hoisted the calculation of which maximum should apply out of the loop, it'd be good to also hoist the explanatory comment (and reword it as needed). Maybe something like:

> Each call should complete in up to the maximum number of operations specified, unless that is less than the minimum number of ops possible in one go for the specified curve size.

tests/suites/test_suite_psa_crypto.function
```diff
@@ -9986,10 +9991,10 @@ void key_agreement_interruptible(int alg_arg,
                number of operations was less than the minium possible for
                the curve size. */
             num_ops = psa_key_agreement_iop_get_num_ops(&operation);
-            if (max_ops <= min_num_ops) {
-                TEST_LE_U(num_ops - num_ops_prior, min_num_ops);
-            } else {
-                TEST_LE_U(num_ops - num_ops_prior, max_ops);
+            uint32_t new_ops = num_ops - num_ops_prior;
+            if (interruptible) {
+                TEST_LE_U(2, new_ops);
```
davidhorstmann-arm: Why do we require that at least 2 ops have happened each call? Surely we could do a call that just perfoms one op?

Unless I've missed something obvious, could you explain why in a comment?

core/psa_crypto.c
```diff
@@ -3768,7 +3768,12 @@ psa_status_t psa_verify_hash_abort(
 void mbedtls_psa_interruptible_set_max_ops(uint32_t max_ops)
 {
 
-#if (defined(MBEDTLS_PSA_BUILTIN_ALG_ECDSA) || \
+#if (defined(MBEDTLS_PSA_BUILTIN_ALG_ECDH) || \
+    defined(MBEDTLS_PSA_BUILTIN_KEY_TYPE_ECC_KEY_PAIR_GENERATE) || \
+    defined(MBEDTLS_PSA_BUILTIN_KEY_TYPE_ECC_KEY_PAIR_IMPORT) || \
+    defined(MBEDTLS_PSA_BUILTIN_KEY_TYPE_ECC_KEY_PAIR_EXPORT) || \
+    defined(MBEDTLS_PSA_BUILTIN_KEY_TYPE_ECC_PUBLIC_KEY) || \
```
valeriosetti, line 3775: A couple of things on this:
1. please update al the comment on the `#endif` line;
2. ideally I agree with the dependency on `MBEDTLS_PSA_BUILTIN_KEY_TYPE_ECC_KEY_PAIR_xxx`, but it wasn't there before and, according to similar scenarios in this file, `_KEY_PAIR_xxx` is used in conjunction with `_PUBLIC_KEY` when there is no algorithm to be also checked. Perhaps we can skip this check? Uniformize it across the entire file is out of scope for this PR.

valeriosetti (changes requested):

davidhorstmann-arm (changes requested):
A couple of small things but LGTM otherwise

core/psa_crypto.c
mpg: Regarding the second point, I'm sorry, but I'm not sure I understand what you suggest, can you clarify?

tests/suites/test_suite_psa_crypto.function
mpg: Hmm, good question, I'm not sure I remember exactly how I ended up with this value of 2. I'll check and come back to you.

core/psa_crypto.c
valeriosetti: A part from import/export functions in this file, other functions where a `MBEDTLS_PSA_BUILTIN_ALG_xxx` is set usually do not have `MBEDTLS_PSA_BUILTIN_KEY_TYPE_ECC_KEY_PAIR_(IMPORT|EXPORT|GENERATE|DERIVE)`. We can keep them here as I think this makes it more complete, but it would be nice to update other functions as well similarly

mpg: I think this function is different, because it's used by all interruptible operations. It's possible in principle to enable say restartable ECC key generation without enabling ECDH nor ECDSA. Then this function needs to be defined in order for restartable ECC key generation to be actually restartable.

Before this PR we had a bug where restartable ECDH wasn't actually restartable when ECDSA was not defined, which was caught by the improved restartable ECDH tests (and `depends.py`). Then I reasoned we have a similar but if key support is enabled but none of ECDH or ECDSA - except it isn't caught by the tests yet, so I tried fixing it by listing all the key operations that have an interruptible version here.

davidhorstmann-arm:
For interest, I measured the code size difference when compiling for M55 with GCC in the baremetal config:

| file | development | this PR | delta |
|---|---|---|---|
| psa_crypto.c.obj | 24008B | 23996B | -12B |
| psa_crypto_ecp.c.obj | 2244B | 2228B | -16B |
| **total:** |  |  | -28B |

Small wins perhaps, but I'm sure we can get more if we eliminate `ecdh.c`.

mpg:
Ah, I'm rebasing on top of development too, because I think the CI won't pass otherwise (framework vs X.509 thing). The commit your reviewed was f67bfa5dadea0fddadc6447641ae922fc4b3454d

valeriosetti (dismissed):

davidhorstmann-arm (dismissed):
LGTM, thanks!

mpg:
When trying to backport the test improvements and bug fix to 3.6, I realise those are not applicable: 3.6 doesn't have interruptible key agreement. This was introduced by https://github.com/Mbed-TLS/mbedtls/pull/9410 and following PRs starting in August 2024 while 3.6.0 was released in March that year.

However this grew a conflict while waiting for me to do the (unnecessary) backport, so I'm going to rebase.

valeriosetti (approved):
I `range-diff` checked the last force push and it still seems OK to me

davidhorstmann-arm (approved):
Similarly, I've range-diffed it and it looks sensible - the only difference being introduced by the move of `crypto_builtin_composites.h` in #587.

PR #581, opened 2025-11-24
psa: Reorganise values of PSA_ERROR_

nicola-mazzucato-arm:
Move standard error codes into a dedicated header that crypto_values.h will consume.
Remove duplication of status codes in crypto_values.h

This is to align with PSA implementation through a common header.

## Description

Please write a few sentences describing the overall goals of the pull request's commits.

## PR checklist

## Notes for the submitter

Help make review efficient:
* Multiple simple commits
* Avoid force-push

### DIFF
--- a/include/psa/crypto_values.h
+++ b/include/psa/crypto_values.h
@@ -27,25 +27,14 @@
 #ifndef PSA_CRYPTO_VALUES_H
 #define PSA_CRYPTO_VALUES_H
 #include "mbedtls/private_access.h"
+#include <psa/error.h>
 #include <psa/crypto_types.h>
 
 /** \defgroup error Error codes
  * @{
  */
 
-/* PSA error codes */
-
-/* Error codes are standardized across PSA domains (framework, crypto, storage,
- * etc.). Do not change the values in this section or even the expansions
- * of each macro: it must be possible to `#include` both this header
- * and some other PSA component's headers in the same C source,
- * which will lead to duplicate definitions of the `PSA_SUCCESS` and
- * `PSA_ERROR_xxx` macros, which is ok if and only if the macros expand
- * to the same sequence of tokens.
- *
- * If you must add a new
- * value, check with the Arm PSA framework group to pick one that other
- * domains aren't already using. */
+/* Extend Standard PSA error codes */
 
 /* Tell uncrustify not to touch the constant definitions, otherwise
  * it might change the spacing to something that is not PSA-compliant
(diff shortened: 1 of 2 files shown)


gilles-peskine-arm (changes requested):
The header name `<psa/error.h>`  is reserved for PSA firmware framework implementations. If TF-PSA-Crypto is used together with a firmware framework implementation, the two projects may define different sets of error codes. If both have a header called `<psa/error.h>`, then whichever project comes first on the include path will hide the other project, which may lead to missing definitions.

This is why TF-PSA-Crypto must not ship `<psa/error.h>`.

adeaarm:
This is the issue that we have in TF-M: We ship `error.h` because we have the SPM. Currently TF-PSA-Crypto `crypto_values.h` is redefining all the error codes and not just the ones which are reserved for the Crypto API, i.e. if TF-PSA-Crypto can't assume to have `psa/error.h` available itself, it then needs to protect those redefinitions with some ifndef define endif pattern and give the option to plugin an external PSA_ERROR_HEADER_FILE, in my opinion.

gilles-peskine-arm:
The [PSA status codes specification](https://arm-software.github.io/psa-api/status-code/1.0/api/status-codes.html#api-reference) was carefully designed so that it is allowed and non-problematic for multiple headers to define an overlapping set of status codes.

TF-PSA-Crypto cannot assume that `<psa/error.h>` is present because it must be able to run on platforms that don't have it. TF-PSA-Crypto cannot provide `<psa/error.h>` because it must not shadow the platform's header on platforms that do have it.

It's ok to include both `<psa/error.h>` and `<psa/crypto.h>` in the same C source because if they define the same macro, then for compliance with the PSA status codes specification, those macros must have the same expansion (not just the same numerical values when evaluated). The only thing that may not be duplicated is the definition of the type `psa_status_t`, and there, again for compliance with the PSA status codes specification, the type definition is done only by whoever is the first to define `PSA_SUCCESS`.

adeaarm:
> It's ok to include both `<psa/error.h>` and `<psa/crypto.h>` in the same C source because if they define the same macro, then for compliance with the PSA status codes specification, those macros must have the same expansion (not just the same numerical values when evaluated). 

This is interesting, because even if they have the same expansion, recent versions of Armclang (but not gcc) flag this as warning thus leading to error if `-Werror` is set:

```
In file included from /data_nvme1n1/jaccoo01/tf-m/trusted-firmware-m/secure_fw/partitions/crypto/../../../interface/include/psa/client.h:15,
                 from /data_nvme1n1/jaccoo01/tf-m/trusted-firmware-m/secure_fw/partitions/crypto/tfm_crypto_api.h:19,
                 from /data_nvme1n1/jaccoo01/tf-m/trusted-firmware-m/secure_fw/partitions/crypto/crypto_alloc.c:15:
/data_nvme1n1/jaccoo01/tf-m/trusted-firmware-m/secure_fw/partitions/crypto/../../../interface/include/psa/error.h:59:9: warning: "PSA_OPERATION_INCOMPLETE" redefined
   59 | #define PSA_OPERATION_INCOMPLETE ((psa_status_t) - 248)
      |         ^~~~~~~~~~~~~~~~~~~~~~~~
In file included from /data_nvme1n1/jaccoo01/tf-m/trusted-firmware-m/secure_fw/partitions/crypto/../../../interface/include/psa/crypto.h:60,
                 from /data_nvme1n1/jaccoo01/tf-m/trusted-firmware-m/secure_fw/partitions/crypto/tfm_mbedcrypto_include.h:18,
                 from /data_nvme1n1/jaccoo01/tf-m/trusted-firmware-m/secure_fw/partitions/crypto/crypto_alloc.c:13:
/data_nvme1n1/jaccoo01/tf-m/trusted-firmware-m/secure_fw/partitions/crypto/../../../interface/include/psa/crypto_values.h:336:9: note: this is the location of the previous definition
  336 | #define PSA_OPERATION_INCOMPLETE           ((psa_status_t)-248)
```

but I have noticed that our `error.h` had spaces in the expansion, not sure if that is being considered as a different expansion in Armclang.

gilles-peskine-arm:
The spaces are the problem. The expansion of the macros has to be the same as a sequence of tokens *and presence-or-absence-of-whitespace*. (Blamce the C standard committee, not me.) So `((psa_status_t) - 248)` does not comply with the PSA status code specification. (Because of this, we have a bypass for our coding style enforcement around our definitions of `PSA_ERROR_xxx` macros.)

nicola-mazzucato-arm:
Thanks both Gilles and Antonio. I drop this PR as we have fixed the redefinitions errors/warnings in the TF-M side.
Thanks again!

PR #582, opened 2025-11-25
Remove ecdh.c

### LINKED ISSUE #1962: mbedtls_ecdh_context fields are not initialized with the proper xxx_init function
This is a minor bug, which can only hit alternative implementations that do something weird.

`mbedtls_ecdh_init` initializes an `mbedtls_ecdh_context` to all-bits-zero. `mbedtls_ecdh_free` calls `mbedtls_ecp_group_free`, `mbedtls_ecp_point_free` and `mbedtls_mpi_free` as applicable on each field. This is not guaranteed to work.

`mbedtls_mpi_free` is safe if its argument is all-bits-zero. It interprets a field as a pointer, but does nothing if the pointer is null, and we require the platform to interpret all-bits-zero as a null pointer (checked in `selftest.c`).

Our implementation of `mbedtls_ecp_group_free` and `mbedtls_ecp_point_free` is safe if its argument is all-bits-zero. More generally, our code using `mbedtls_ecp_group` and `mbedtls_ecp_point` treats a data structure initialized as all-bits-zero identically to a data structure initialized with the corresponding `xxx_init` function. However this may not be the case for alternative implementations.

We should initialize those fields properly by calling `_init`.

The fields `Vi`, `Vf` and `_d` are not used at all, so we could refrain from calling `xxx_free` instead for those fields.

I haven't checked whether other library modules do something similar.

### LINKED ISSUE #4440: Make restartable ECC use the new ECDH context format
In https://github.com/ARMmbed/mbedtls/issues/4404, we removed `MBEDTLS_ECDH_LEGACY_CONTEXT`, so ECDH code uses the new context format. Restartable ECC still uses the old context format. The goal of this task is to switch restartable ECC to using the new ECDH context format. That is, change the code gated by `MBEDTLS_ECP_RESTARTABLE` from forcing `MBEDTLS_ECDH_LEGACY_CONTEXT` on to forcing it off.

### LINKED ISSUE #6048: Remove ECDH legacy context
Remove ECDH legacy context

### LINKED ISSUE #6049: Add low level internal ECDH functions
**Prerequisites:** #6048 

Add low level internal ECDH functions.

### LINKED ISSUE #6050: Switch the ECDH context to the new structures
**Prerequisites:** #6049 

Switch the ECDH context to the new structures.

### LINKED ISSUE #10346: Non-ASCII code character is found in Ecdh.h
Summary: 
Non-ASCII code character is found in Ecdh.h file and causing build failure... 

Details:
Please check line 9. There are two dash line with different length. The 2nd one is absolutely not an ASCII code.

mpg:
## Description

~Based on https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/578~ (merged). For a green CI, also needs:
- ~https://github.com/Mbed-TLS/mbedtls/pull/10523~ (merged) 
- ~https://github.com/Mbed-TLS/mbedtls/pull/10528~ (merged)
- ~https://github.com/Mbed-TLS/mbedtls-framework/pull/254~ (merged)

**Depends on:**
- ~https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/624 to start from a clean state.~ (merged) ~TODO: revisit [Ben's comment about length checks](https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/582#discussion_r2643177933)~.
- ~https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/619 to avoid introducing a hack to pacify check-names~

Resolves https://github.com/Mbed-TLS/mbedtls/issues/1962
Resolves https://github.com/Mbed-TLS/mbedtls/issues/4440
Resolves https://github.com/Mbed-TLS/mbedtls/issues/6048
Resolves https://github.com/Mbed-TLS/mbedtls/issues/6049
Resolves https://github.com/Mbed-TLS/mbedtls/issues/6050
Resolves https://github.com/Mbed-TLS/mbedtls/issues/10346 (for development - still want https://github.com/Mbed-TLS/mbedtls/pull/10347 for 3.6).

Follow-up (or prerequisite?): add (back) FF/ECDH benchmarking based on PSA Crypto - TODO: create issue!

## PR checklist

- [x] **changelog** not required because: internal
- [x] **mbedtls development PR** provided Mbed-TLS/mbedtls#10523
- [x] **mbedtls 3.6 PR** not required because: still public there
- **tests**  not required because: covered by the existing


davidhorstmann-arm:
Measurements for M55 in `baremetal`:

| Module            |                           Base   |  This PR |  Delta | % Delta |
|---|---|---|---|---|
| ecdh                |                         932     |  0  |  -932 | -100.00% |
| TOTAL            |                         206686 | 205754 |   -932  |  -0.45% |

Nearly 1K of savings, nice!

drivers/builtin/src/psa_crypto_ecp.c
```diff
@@ -515,33 +518,85 @@ psa_status_t mbedtls_psa_ecdsa_verify_hash(
...
+    }
+
+    return mbedtls_to_psa_error(
+        mbedtls_ecp_get_type(grp) == MBEDTLS_ECP_TYPE_MONTGOMERY ?
+        mbedtls_mpi_write_binary_le(&secret->X, shared_secret, *shared_secret_length) :
+        mbedtls_mpi_write_binary(&secret->X, shared_secret, *shared_secret_length));
+}
+
+#if defined(MBEDTLS_ECDH_VARIANT_EVEREST_ENABLED)
+static psa_status_t ecdh_everest_shared_secret(
```
bjwtaylor, line 538: Should we still be checking for NULL buffers? I seem to remember we agreed we didn't need to have test coverage of them, however I can't remember if we agreed to allow the potential segfault or check it prior to writing?

drivers/builtin/src/psa_crypto_ecp.c
```diff
@@ -515,33 +518,85 @@ psa_status_t mbedtls_psa_ecdsa_verify_hash(
...
+    const uint8_t *key_buffer, size_t key_buffer_size,
+    const uint8_t *peer_key, size_t peer_key_length,
+    uint8_t *shared_secret, size_t shared_secret_size,
+    size_t *shared_secret_length)
+{
+    if (key_buffer_size < MBEDTLS_X25519_KEY_SIZE_BYTES ||
+        peer_key_length < MBEDTLS_X25519_KEY_SIZE_BYTES) {
+        return PSA_ERROR_INVALID_ARGUMENT;
```
bjwtaylor, line 554: Do we need to test these cases, in the standard test suite lcov indicates they are not covered with MBEDTLS_ECDH_VARIANT_EVEREST_ENABLED.

bjwtaylor (commented):
Just one question on the buffers, as I can't remember what we agreed

drivers/builtin/src/psa_crypto_ecp.c
mpg: As a general rule, we don't check for NULL pointers passed as arguments, and are OK with segfault or whatever the platform is doing. (There are exceptions with legacy code, for example in PK, and we're not actively removing those checks, but new code doesn't have them.)

drivers/builtin/src/psa_crypto_ecp.c
mpg: Ah good point, and thanks for running `lcov` as part of your review!

Taking a closer look:
- I don't think the first one is something that can happen: this function is only called in one place, behind a test that the curve is 25519. Also, the key buffer comes from the keystore. So, I think the only way `key_buffer_size < MBEDTLS_X25519_KEY_SIZE_BYTES` can be true here is if the keystore is inconsistent (attributes say the curve is 25519 but key buffer doesn't match). I don't think that's something that can happen, so we don't need a test case for it. However I'm going to remove the check and replace it with a comment explaining why we don't need it.
- For the second one though, `peer_key_length` comes quite directly from user input, so it definitely seems possible and highly desirable to test with it being invalid. I'll add a test case for that.

mpg: > For the second one though, `peer_key_length` comes quite directly from user input, so it definitely seems possible and highly desirable to test with it being invalid. I'll add a test case for that.

It was so desirable to test that the test revealed an issue in interruptible ECDH: https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/624

drivers/builtin/src/psa_crypto_ecp.c
```diff
@@ -541,8 +541,13 @@ static psa_status_t ecdh_everest_shared_secret(
     uint8_t *shared_secret, size_t shared_secret_size,
     size_t *shared_secret_length)
 {
-    if (key_buffer_size < MBEDTLS_X25519_KEY_SIZE_BYTES ||
-        peer_key_length < MBEDTLS_X25519_KEY_SIZE_BYTES) {
+    /* This static function is only called when we know the curve is x25519,
+     * so we know key_buffer_size is correct unless the keystore is corrupted.
+     */
+    (void) key_buffer_size;
```
mpg: @gilles-peskine-arm I'd like your opinion on this: is it correct to skip this check? Previously we were checking that `key_buffer_size` was large enough, but [Ben pointed out](https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/582/commits/95331f1c0671501bd92be8ae7327a6152c404130#r2643177933) that was not covered by tests. I reasoned that we can't hit this case unless the keystore has inconsistent information (a key whose attributes say it's x25519 but the key buffer is of the wrong size), which I think cannot happen, so went on to remove the check. But I'd like an extra pair or eyes on this change just to be sure.

gilles-peskine-arm: Check the key size in a driver against something that should have been enforced during key creation is a second line of defense. It's not absolutely mandatory, but it's better to have it. It's mostly due to persistent keys.

Suppose an older version of Mbed TLS of TF-PSA-Crypto had a bug that allowed some unexpected keys to be created. Then we fix that bug and we're happy. But users may have such unexpected in the key store. So we'd need to keep track of the former existence of that bug and make sure that we keep this particular check somewhere.

Another thing that can happen with persistent keys is if the storage is not cryptographically protected, and gets corrupted. It's not our job to defend against corrupted storage, but it's better if that can merely result in bad operation and possibly security holes (e.g. if a MAC key used for verification reads out as 0000) but not all the way to a buffer overwrite (though for keys I guess it would only be a buffer overread which is no worse than other things you could easily do with a corrupted key store).

mpg: Ok, so I'll add the check back, with a comment explaining why we're not expecting it to be covered by tests, but we want to have it anyway.

drivers/builtin/src/psa_crypto_ecp.c
mpg: After the latest rebase and changes, the check on `peer_key_length` should now be covered, and the other one has a comment explaining why it's not covered but still present.

bjwtaylor (approved):

valeriosetti (approved):

Issue #542, opened 2025-10-22
Builds with accelerated SHA3 still include the built-in implementation


gilles-peskine-arm:
The code in `sha3.c` and `test_suite_shax.*` uses `PSA_WANT_ALG_SHA3_xxx` as compilation guards. As a result, both will be included in the build even if SHA3 is fully accelerated.

Applies to TF-PSA-Crypto 1.0. It's ok in 3.6.

The goal of this issue is to fix this: if SHA3 is accelerated, `sha3.o` must be empty. Assert this [in `component_test_accel_hash` in `components-configuration-crypto.sh`](https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/586#discussion_r2615123148).

PR #586, opened 2025-12-01
Add support for driver dispatch testing 2/2

### LINKED ISSUE #61: Add support for libtestdriver1
Adapt to CMake build system the libtestdriver1 helper shell functions:
- helper_libtestdriver1_adjust_config
- helper_libtestdriver1_make_drivers: need to add support for the build of the test library in tests/CMakeLists.txt
- helper_libtestdriver1_make_main

As of mbedtls:dev(bae154d76cfe0ee844bf88efc5bbe75e2088d4b0), these helper shell functions are located in all.sh that is yet to be moved to the framework. If they stay in all.sh(preferred), the 3.6 and mbedtls:dev components may need some adaptation. If the helper shell functions needs to be different for TF-PSA-Crypto and mbedtls then they need to be moved into
the branch dependent components-configuration-crypto.sh file.

Demonstrate testing with `libtestdriver1` by adding the equivalent of `test_psa_crypto_config_accel_ecdsa`.

Revisit the compilation guards in `tests/src/drivers/*.c` like:
```C
if (MBEDTLS_VERSION_MAJOR == 3) && (MBEDTLS_VERSION_MINOR == 6)
#include "libtestdriver1/library/psa_crypto_hash.h"
#else
#include "libtestdriver1/tf-psa-crypto/core/psa_crypto_hash.h"
#endif
#endif
```
Can we do better?

ronald-cron-arm:
## Description
Fixes #61
Based on #587

## PR checklist
- [x] **framework PR** provided Mbed-TLS/mbedtls-framework#243
- [x] **mbedtls 3.6 PR**  not required because: TF-PSA-Crypto only work


gilles-peskine-arm:
The commits are nicely split, but even so, 17 commits is on the large side. Would it make sense to split this PR just before "Add generate_test_driver.py"?

ronald-cron-arm:
> The commits are nicely split, but even so, 17 commits is on the large side. Would it make sense to split this PR just before "Add generate_test_driver.py"?

I think it is better to have the whole stack to better understand the why of the first commits but I can split the PR into two as you suggested. I will do that tomorrow.

ronald-cron-arm:
I've created #587 now with the first 10 commits. #587 should be reviewed first.

tests/scripts/generate_test_driver.py
```diff
@@ -0,0 +1,181 @@
...
+import itertools
+import sys
+
+from fnmatch import fnmatch
+from pathlib import Path
+from typing import Iterable, List, Set
+
+import scripts_path # pylint: disable=unused-import
+from mbedtls_framework import build_tree
+from mbedtls_framework import test_driver
```
gilles-peskine-arm, line 19: There's more code in the crypto repository than I would expect, i.e. more code that won't be shared between branches of TF-PSA-Crypto. It makes sense to me that the list of files is here, but at first glance, I'd say all of the code belongs in the framework. Can you please explain how you decided the framework/consumer split?

tests/scripts/components-configuration-crypto.sh
```diff
@@ -0,0 +1,45 @@
...
+
+################################################################
+#### Configuration Testing - Crypto
+################################################################
+
+BUILTIN_BUILD_DIR="drivers/builtin/CMakeFiles/builtin.dir/src"
+support_universal_ctags_5_9_0() {
+    version="$(ctags --version 2>/dev/null)" || return 1
+    [[ "$version" == *"Universal Ctags 5.9.0"* ]]
+}
```
gilles-peskine-arm: That's a very specific requirement: it has to be Universal Ctags (not Exuberant Ctags), and it can't be a newer or older version. Does it really need to be that specific?

I tried running `all.sh test_accel_hash` on Ubuntu 16.04 (with Exuberant Ctags: Universal wasn't in Ubuntu yet) plus Python 3.6, and it passed. It also passes with Exuberant Ctags 5.8 (the last official release from 2009).

How about just
```
support_ctags () {
  [[ "$(ctags --version)" = *Exuberant* ]] 2>/dev/null
}
```

That would match Exuberant Ctags since it froze in 2011, and every release of Universal Ctags so far, while excluding Emacs and BSD ctags. (The command line in `test_driver.py` doesn't work with those; it might be possible to make it work, but I'd say not worth it.)

gilles-peskine-arm (commented):
Just some preliminary observations. I'm not doing an actual code review yet.

tests/scripts/generate_test_driver.py
```diff
@@ -0,0 +1,181 @@
...
+IDENTIFIER_PREFIXES = {
+    "MBEDTLS_",
+    "PSA_",
+    "TF_PSA_CRYPTO_",
+    "mbedtls_",
+    "psa_",
+    "tf_psa_crypto_",
+}
+
+def iter_code_files(root: Path) -> Iterable[Path]:
```
gilles-peskine-arm: Duplicated from `test_driver.py`.

tests/scripts/generate_test_driver.py
```diff
@@ -0,0 +1,181 @@
...
+    """
+    for directory in ("include", "src"):
+        directory_path = root / directory
+        for ext in (".c", ".h"):
+            yield from directory_path.rglob(f"*{ext}")
+
+def get_src_relpaths(builtin: Path, exclude_files: Set[str]) -> List[Path]:
+    """
+    Return the relative paths of all *.c and *.h files under `builtin`,
+    excluding those whose names match any of the patterns in `exclude_files`.
```
gilles-peskine-arm: “file name” either refers to the base name or to the full path (not necessarily absolute), depending on who you ask. Please clarify.


```suggestion
    excluding those whose basenames match any of the patterns in `exclude_files`.
```

tests/scripts/generate_test_driver.py
```diff
@@ -0,0 +1,181 @@
...
+
+    The returned paths are relative to `builtin`.
+    """
+
+    out = []
+    for file in iter_code_files(builtin):
+        if not any(fnmatch(file.name, pattern) for pattern in exclude_files):
```
gilles-peskine-arm: The same pattern appears in `test_driver.py`. It should be an auxiliary function:

```
return [path.relative_to(builtin) for path in iter_code_files_with_excludes(builtin, exclude_files)]
```

tests/scripts/generate_test_driver.py
```diff
@@ -0,0 +1,181 @@
...
+            out.append(file.relative_to(builtin))
+    out.sort()
+    return out
+
+def get_dst_relpaths(src_relpaths: List[Path], driver: str) -> List[Path]:
```
gilles-peskine-arm: This duplicates some of the logic in `test_driver.py` and should be unified.

tests/scripts/generate_test_driver.py
```diff
@@ -0,0 +1,181 @@
...
+    """
+    Get from public and core headers the identifiers that the driver built-in
+    code may reference but does not define.
+    """
+    directories = ("core", "include")
+    files = itertools.chain.from_iterable(Path(directory).rglob("*.h") \
+                                          for directory in directories)
+    identifiers = set()
+    for file in files:
+        identifiers.update(test_driver.run_ctags(file))
```
gilles-peskine-arm: Why not run `ctags` on the list of files in a single batch?

tests/scripts/generate_test_driver.py
```diff
@@ -0,0 +1,181 @@
...
+    identifiers |= set(TFPSACryptoConfig().settings)
+
+    return identifiers
+
+def main():
+    """
+    Main function of this program
+    """
+    parser = test_driver.get_parsearg_base()
+    parser.add_argument('--list-vars-for-cmake', nargs="?", const="__AUTO__", metavar="FILE",
```
gilles-peskine-arm: Please document that this is done _instead_ of creating a test driver tree, not in addition.

tests/scripts/generate_test_driver.py
```diff
@@ -0,0 +1,181 @@
...
+        raise RuntimeError("This script must be run from TF-PSA-Crypto root.")
+
+    builtin = Path("drivers/builtin")
+    if not builtin.is_dir():
+        raise RuntimeError("Do not find expected drivers/builtin directory.")
+    src_relpaths = get_src_relpaths(builtin, EXCLUDE_FILES)
+
+    dst_dir = Path(args.dst_dir)
+    if not dst_dir.is_dir():
+        raise RuntimeError(f"Do not find expected {args.dst_dir} directory.")
```
gilles-peskine-arm: Why doesn't the script create the destination directory? That's what I would have expected.
```
git checkout …
tests/scripts/generate_test_driver.py drivers/testdriver1
tests/scripts/generate_test_driver.py drivers/testdriver2
```

tests/scripts/generate_test_driver.py
```diff
@@ -0,0 +1,181 @@
...
+
+    if args.list_vars_for_cmake:
+        fname = f"{args.driver}-list-vars.cmake" \
+                if args.list_vars_for_cmake == "__AUTO__" else args.list_vars_for_cmake
+
+        with open(dst_dir / fname, "w") as f:
+            f.write(f"set({args.driver}_input_files " + \
+                     " ".join(str(path) for path in src_relpaths) + ")\n\n")
```
gilles-peskine-arm: To facilitate debugging, it would be nicer to have one file per line.

tests/scripts/generate_test_driver.py
```diff
@@ -0,0 +1,181 @@
...
+            f.write(f"set({args.driver}_files " + \
+                    " ".join(str(path) \
+                    for path in get_dst_relpaths(src_relpaths, args.driver)) + ")\n\n")
+            f.write(f"set({args.driver}_src_files " + \
+                    " ".join(str(path) \
+                    for path in src_relpaths if path.suffix == ".c") + ")")
```
gilles-peskine-arm: Missing final newline.

tests/scripts/generate_test_driver.py
```diff
@@ -0,0 +1,181 @@
...
+        return
+
+    #Step 1: Build the test driver tree from `drivers/builtin`
```
gilles-peskine-arm: Both here and in `test_driver.py`, “build” is misleading: in this context, it tends to refer to invoking a C compiler. “Build tree” is even worse because as a noun, it means the directories where non-source files are placed, as opposed to the source tree. We tend to use “generate” for this kind of step that happens before compilation, but it is admittedly vague and already overused. I'd still prefer “generate” to “build”, or better, some different verb. Putting files in a directory can be “populating” it, but that strongly presumes a preexisting target directory. We could use “construct”, which is a synonym of “build” in everyday English but does not have a commonly used meaning in this context.

tests/scripts/generate_test_driver.py
```diff
@@ -0,0 +1,181 @@
...
+    # MBEDTLS_PSA_ACCEL_ are not defined in the code base. They are supposed to
+    # be passed as extra configuration options. We want to prefix them,
+    # especially in 'crypto_adjust_config_enable_builtins.h', thus deduce them
+    # from the PSA_WANT_ ones and add them to the list of identifiers to
+    # prefix in the test driver code.
+    for identifier in external_identifiers:
+        if identifier.startswith("PSA_WANT_"):
+            identifiers.add(identifier.replace("PSA_WANT_", "MBEDTLS_PSA_ACCEL_", 1))
+
+    generator.prefix_identifiers(identifiers)
```
gilles-peskine-arm: Having reviewed `generate_test_driver.py`, I now think that all the code should be in the shared module `test_driver.py`, except:

* Branch-specific data: `EXCLUDE_FILES`, maybe `IDENTIFIER_PREFIXES`.
* The ad hoc fiddling of `identifiers` after adding the external identifiers and before calling `prefix_identifers`.
* Calling `main`.

I think this would be best handled by putting all the logic in `TestDriverGenerator`, and having a method to adjust the identifiers which does nothing by default and is overridden in a derived class defined in the branch-specific script.

gilles-peskine-arm (changes requested):
I have reviewed `tests/scripts/generate_test_driver.py`, but not the rest of this PR, at c1bad1c6232855d4b37884d40b3db727321059df. This is a companion to my [review of the framework part](https://github.com/Mbed-TLS/mbedtls-framework/pull/243#pullrequestreview-3536066960).

tests/scripts/generate_test_driver.py
ronald-cron-arm: This is more convenient for the integration with CMake. At least with the way I've done it.
During the configuration phase, CMake creates the libtestdriver1 directory in the build tree, adds a CMakeLists.txt into it and then adds it as a subdirectory of drivers: essentially add_subdirectory(libtestdriver1), like we have add_subdirectory(builtin).
The CMakeLists.txt in `libtestdriver1` is roughly just the `builtin/CMakeLists.txt` one (as it is inside `libtestdriver1/`) plus the code to generate the test driver code during the build  phase.
Thus `generate_test_driver.py`  is called when `libtestdriver1` has already been created and contains its CMakeLists.txt file.

tests/scripts/generate_test_driver.py
ronald-cron-arm: My guidelines for the split between TF-PSA-Crypto and the framework:
- no knowledge of the repo tree in the framework but that a driver directory contains an `include` and `src` directory ...
- I was not really sure about the need for the "--list-vars-for-cmake" option thus I kept it and its associated code in TF-PSA-Crypto.

tests/scripts/components-configuration-crypto.sh
ronald-cron-arm: That's the only way I found so far to get ctags running successfully on the CI. The ctags command was falling on Ubuntu 16.04 CI even if the command is supposed to be compatible with the Exuberant version you mention above. But I can dig a bit more.

tests/scripts/generate_test_driver.py
ronald-cron-arm: We could use "populate" then as we have a preexisting target directory.

tests/scripts/generate_test_driver.py
ronald-cron-arm: Because the command would contain a long list of files.

tests/scripts/generate_test_driver.py
ronald-cron-arm: I've changed to `create_test_driver_tree`.

tests/scripts/generate_test_driver.py
ronald-cron-arm: I have something along these lines now.

ronald-cron-arm:
@minosgalanakis @gilles-peskine-arm , as discussed offline, currently this PR is mostly part of the validation of the framework PR. Compared to the previous version I've rebased on top of the current head of https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/587 (https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/587/commits/a548d3fadcba59f35ba4e4ed068874346c207349) and then added commits at the top to adjust to the changes in test_driver.py in the framework PR:
```
2b70a5e35 Adapt to systematic file basename prefixing
902bfbfbc generate_test_driver.py: Adjust to build_tree renaming
ae7c2c4f5 generate_test_driver.py: Define subclass of TestDriverGenerator
1f62f9b30 generate_test_driver.py: Remove support for list-vars-for-cmake
```
If you checkout one of these commits and update the associated framework commit, the CI component `test_accel_hash` passes successfully.

tests/scripts/generate_test_driver.py
gilles-peskine-arm: As of 2b70a5e35e5b5b92aa24295cfbaf166db61160e8, the script works fine whether the destination directory exists or not, except for this explicit check. So please remove it.

CMakeLists.txt
```diff
@@ -391,6 +406,9 @@ if(ENABLE_TESTING OR ENABLE_PROGRAMS)
         add_custom_target(tf_psa_crypto_test_keys_header
             DEPENDS ${CMAKE_CURRENT_BINARY_DIR}/tests/include/test/test_keys.h)
         add_dependencies(tf_psa_crypto_test tf_psa_crypto_test_keys_header)
+        if(TF_PSA_CRYPTO_TEST_DRIVER)
```
minosgalanakis, line 409: Design question. Is there any feasible use case to disable testing and programs, and still need the TF_PSA_CRYPTO_TEST_DRIVER? if so that logic should be revisitted.

tests/scripts/libtestdriver1.cmake
```diff
@@ -0,0 +1,28 @@
...
+
+include(${CMAKE_CURRENT_BINARY_DIR}/libtestdriver1-list-vars.cmake)
+
+add_custom_command(
+    OUTPUT ${libtestdriver1_files}
+    WORKING_DIRECTORY ${PROJECT_SOURCE_DIR}
+    COMMAND
+        ${TF_PSA_CRYPTO_PYTHON_EXECUTABLE}
+            ./tests/scripts/generate_test_driver.py ${CMAKE_CURRENT_BINARY_DIR}
+    DEPENDS ${PROJECT_SOURCE_DIR}/tests/scripts/generate_test_driver.py
```
minosgalanakis, line 34: Why this depedency on the script itself? I understand on why that is needed for development, but should we also regenerate the files when the input header have been touched?

Something like:

~~~~~~
add_custom_command(
    OUTPUT ${libtestdriver1_files}
    WORKING_DIRECTORY ${PROJECT_SOURCE_DIR}
    COMMAND
        ${TF_PSA_CRYPTO_PYTHON_EXECUTABLE}
            ./tests/scripts/generate_test_driver.py ${CMAKE_CURRENT_BINARY_DIR}
    DEPENDS
        ${PROJECT_SOURCE_DIR}/tests/scripts/generate_test_driver.py
       ${libtestdriver1_input_files}
~~~~~~

I think we can access the output of 'generate_test_driver.py --list-vars-for-cmake' at this stage?

minosgalanakis (commented):
Looks good overall, and no big objections to the design. Since it is in the bottom of the PR depedency chain, I am only commenting at this stage

CMakeLists.txt
ronald-cron-arm: Programs yes but not testing. TF_PSA_CRYPTO_TEST_DRIVER is part of the testing.

ronald-cron-arm:
@gilles-peskine-arm @minosgalanakis I've rebased on top of development after the merge #587. I've kept the history of the work on generate_test_driver.py alongside the work on test_driver.py in the framework t the bottom of the commit stack. I believe as part of the review of https://github.com/Mbed-TLS/mbedtls-framework/pull/243 you have already reviewed up to the commit "Update framework - latest version of test_driver.py". I also believe I have addressed all the comments I received so far. The CI is green this the PR is ready for a full review now.

core/CMakeLists.txt
```diff
@@ -171,6 +171,11 @@ foreach(target IN LISTS target_libraries)
             target_include_directories(${target} PUBLIC $<BUILD_INTERFACE:${dir}>)
         endforeach()
     endforeach()
+    if(TF_PSA_CRYPTO_TEST_DRIVER)
+        target_include_directories(${target}
+            PRIVATE ${TF_PSA_CRYPTO_FRAMEWORK_DIR}/tests/include
+                    ${TF_PSA_CRYPTO_DIR}/tests/include/test)
```
gilles-peskine-arm: Why are we putting `tests/include/test` and not `tests/include` on the include path?

drivers/CMakeLists.txt
```diff
@@ -26,6 +25,22 @@ if(CMAKE_COMPILER_IS_MSVC)
     endif()
 endif()
 
+if(TF_PSA_CRYPTO_TEST_DRIVER)
+    # Use `libtestdriver1` as the driver name to align with the name used in
+    # Mbed TLS. This allows to reuse most of the test driver code as it is.
```
gilles-peskine-arm: I think that's in reference to Mbed TLS 3.6, right? In context I would understand it as referring to the version of Mbed TLS that matches the version of TF-PSA-Crypto that I'm reading, but I don't see why the content of that version of Mbed TLS would matter here.

drivers/CMakeLists.txt
```diff
@@ -26,6 +25,22 @@ if(CMAKE_COMPILER_IS_MSVC)
...
+    list(APPEND TF_PSA_CRYPTO_DRIVERS_INCLUDE_DIRS ${CMAKE_CURRENT_BINARY_DIR}/libtestdriver1/include)
+
+    file(MAKE_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/libtestdriver1)
+    configure_file(
+        ${PROJECT_SOURCE_DIR}/tests/scripts/libtestdriver1.cmake
+        ${CMAKE_CURRENT_BINARY_DIR}/libtestdriver1/CMakeLists.txt
+        COPYONLY
+    )
+    add_subdirectory(${CMAKE_CURRENT_BINARY_DIR}/libtestdriver1
+                     ${CMAKE_CURRENT_BINARY_DIR}/libtestdriver1)
```
gilles-peskine-arm, line 59: Please explain what's happening here, and how `tests/scripts/libtestdriver1.cmake` gets used. Even the fact that these are the same question isn't fully obvious.

As a CMake newbie, I think what's going on is:

1. We copy `tests/scripts/libtestdriver1.cmake` somewhere in the build tree, and we choose to give it the basename `CMakeLists.txt`.
2. We tell CMake to treat the location of this newly created `CMakeLists.txt` file as a source directory, so it reads and executes that `CMakeLists.txt`.
3. The fact that `${CMAKE_CURRENT_BINARY_DIR}/libtestdriver1` is both the source tree and the build tree for that part of the build is nothing special: it's what happens in any in-tree build, and CMake lets you choose the source/build tree correspondence independently in each subdirectory.

Please let me know if that's wrong or incomplete.

What's the benefit of this approach, compared to having `drivers/libtestdriver1/CMakeLists.txt` in the source tree in version control?

tests/scripts/libtestdriver1.cmake
```diff
@@ -0,0 +1,34 @@
...
+            ${input_files}
+    COMMENT "Generating test driver libtestdriver1 tree"
+)
+
+add_custom_target(tf_psa_crypto_test_driver_src
```
gilles-peskine-arm: So it's `make tf_psa_crypto_test_driver_src` to generate the sources but `make libtestdriver1` to compile them? Can we make that consistent?

tests/configs/user-config-test-driver-extension.h
```diff
@@ -0,0 +1,426 @@
...
+#endif
+
+#if defined(PSA_WANT_ECC_SECP_R1_192) && \
+    !defined(MBEDTLS_PSA_ACCEL_ECC_SECP_R1_192)
+#define LIBTESTDRIVER1_MBEDTLS_PSA_ACCEL_ECC_SECP_R1_192 1
+#endif
+
+#if defined(PSA_WANT_DH_RFC7919_2048) && \
+    !defined(MBEDTLS_PSA_ACCEL_DH_RFC7919_2048)
+#define LIBTESTDRIVER1_MBEDTLS_PSA_ACCEL_DH_RFC7919_2048
```
gilles-peskine-arm: For some reason the `DH_xxx` definitions have an empty expansion instead of 1 like all the others.

tests/configs/user-config-test-driver-extension.h
```diff
@@ -0,0 +1,426 @@
+/**
```
gilles-peskine-arm, line 1: This file would be trivial to generate in terms of its content, and it's repetitive and error prone. With the overhaul of the libtestdriver1 generation, how hard would it be to fit its generation into the build scripts?

tests/scripts/components-configuration-crypto.sh
```diff
@@ -0,0 +1,37 @@
...
+    cmake -DTF_PSA_CRYPTO_TEST_DRIVER=On \
+          -DTF_PSA_CRYPTO_USER_CONFIG_FILE="../tests/configs/user-config-accel-hash.h" ..
+    make
+
+    # Make sure built-in hash objects are empty.
+    not grep mbedtls_md5 ${CMAKE_BUILTIN_BUILD_DIR}/md5.c.o
+    not grep mbedtls_sha1 ${CMAKE_BUILTIN_BUILD_DIR}/sha1.c.o
+    not grep mbedtls_sha256 ${CMAKE_BUILTIN_BUILD_DIR}/sha256.c.o
+    not grep mbedtls_sha512 ${CMAKE_BUILTIN_BUILD_DIR}/sha512.c.o
+    not grep mbedtls_ripemd160 ${CMAKE_BUILTIN_BUILD_DIR}/ripemd160.c.o
```
gilles-peskine-arm, line 30: Just noting: This should have SHA3, except that it can't due to https://github.com/Mbed-TLS/TF-PSA-Crypto/issues/542

gilles-peskine-arm (changes requested):
I think the code is correct, but there are a few places that I'm not sure I understand and I want at least one comment.

minosgalanakis (dismissed):
LGTM.

tests/scripts/generate_test_driver.py is now  neat and explicit and easy to understand and maintain.

The build system changes are consistent with the agreed design. Thank you

p.s It still will require a rebase for the framework conflict

core/CMakeLists.txt
ronald-cron-arm: This is the specific path for "mbedtls/build_info.h" that is included by the framework test driver headers, like in `framework/tests/include/test/drivers/aead.h`.

drivers/CMakeLists.txt
ronald-cron-arm: I've tried and improved the comment.

drivers/CMakeLists.txt
ronald-cron-arm: I've added a comment. 
Your three points above are correct.
> What's the benefit of this approach, compared to having drivers/libtestdriver1/CMakeLists.txt in the source tree in version control?

This way we avoid having a test-specific directory in the library tree. We also plan to generate multiple test drivers, and this approach seems more flexible for that.

tests/scripts/libtestdriver1.cmake
ronald-cron-arm: Changed to `libtestdriver1_generation` and added a TF_PSA_CRYPTO_TEST_DRIVER_GENERATION_TARGETS list in the model of TF_PSA_CRYPTO_DRIVER_TARGETS to avoid hard-coding `libtestdriver1` in test and drivers build scripts.

tests/configs/user-config-test-driver-extension.h
ronald-cron-arm: We could generate it in the build tree in `drivers/libtestdriver1/include` or a sub-directory of it and this should make it. This would mean a framework PR to add the generation logic to test_driver.py. Thus I would prefer to do this later.

ronald-cron-arm:
@gilles-peskine-arm @minosgalanakis I believe I've addressed all the comments. I've added five commits and had to rebase because of a framework conflict. The head before the rebase was https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/586/commits/bd41097f535607cefadec77dbca9d557c2240aec. The rebase can be checked with:
```
git range-diff HEAD~19..HEAD bd41097f535607cefadec77dbca9d557c2240aec~19..bd41097f535607cefadec77dbca9d557c2240aec
```
Only difference: the initial framework commit in the first commit that updates the framework pointer.
The CI is green, this is ready for re-review.

tests/scripts/libtestdriver1.cmake
```diff
@@ -0,0 +1,34 @@
...
+    WORKING_DIRECTORY ${PROJECT_SOURCE_DIR}
+    RESULT_VARIABLE result)
+if(result)
+    message(FATAL_ERROR "Generation of list-vars.cmake failed: ${result}")
+endif()
+
+include(${CMAKE_CURRENT_BINARY_DIR}/libtestdriver1-list-vars.cmake)
+
+set(input_files "")
+foreach(relpath IN LISTS libtestdriver1_input_files)
```
gilles-peskine-arm, line 24: I can't figure out where `libtestdriver1_input_files` is defined.

tests/configs/user-config-test-driver-extension.h
gilles-peskine-arm: Sure, I didn't mean to do it now, sorry for being unclear. Filed as https://github.com/Mbed-TLS/mbedtls-framework/issues/255

core/CMakeLists.txt
gilles-peskine-arm: Ah, ok. Please explain this in a comment.

gilles-peskine-arm (changes requested):
LGTM except I'd like a couple more comments.

tests/scripts/libtestdriver1.cmake
ronald-cron-arm: In libtestdriver1-list-vars.cmake, I am adding a comment.

tests/scripts/libtestdriver1.cmake
```diff
@@ -9,6 +9,10 @@ if(result)
     message(FATAL_ERROR "Generation of list-vars.cmake failed: ${result}")
 endif()
 
+# libtestdriver1-list-vars.cmake defines three list variables:
+# - libtestdriver1_input_files: the list of generate_test_driver.py input files
+# - libtestdriver1_files: the list of generate_test_driver.py output files
+# - libtestdriver1_src_files: the list of generate_test_driver.py output C files
 include(${CMAKE_CURRENT_BINARY_DIR}/libtestdriver1-list-vars.cmake)
```
gilles-peskine-arm, line 21: `libtestdriver1-list-vars.cmake` is created by the `execute_process` just above. It's not obvious because the file name doesn't appear in the command. It is a reasonable expectation given the sequence of instructions in `libtestdriver1.cmake`, but still, I would prefer to have a comment to explain this.

Come to think of it, if `generate_test_driver.py` changes, will CMake re-generate `libtestdriver1-list-vars.cmake`?

core/CMakeLists.txt
```diff
@@ -172,6 +172,8 @@ foreach(target IN LISTS target_libraries)
         endforeach()
     endforeach()
     if(TF_PSA_CRYPTO_TEST_DRIVER)
+        # See the note on include directories for tf_psa_crypto_test target
+        # regarding "tests/include/test".
```
gilles-peskine-arm: Where is that note?

Ah, found it in the root `CMakeLists.txt`. Please add this information in the comment. Also, for the CMake non-experts, it would help to be more explicit that “note on include directories” means “note above the `target_include_directories(tf_psa_crypto_test` call” (`tf_psa_crypto_test` occupies about 50 lines).

gilles-peskine-arm (changes requested):
Sorry, still more comments please.

core/CMakeLists.txt
ronald-cron-arm: I've just duplicated the note as the pointer to the note was getting as long as the note itself.

tests/scripts/libtestdriver1.cmake
ronald-cron-arm: > `libtestdriver1-list-vars.cmake` is created by the `execute_process` just above. It's not obvious because the file name doesn't appear in the command. It is a reasonable expectation given the sequence of instructions in `libtestdriver1.cmake`, but still, I would prefer to have a comment to explain this.

I've added the file name in the command.  

> Come to think of it, if `generate_test_driver.py` changes, will CMake re-generate `libtestdriver1-list-vars.cmake`?

Not sure we need to support that but as the libtestdriver1 tree creation depends on `generate_test_driver.py`, I've updated the code to re-trigger the configuration phase (and thus the generation of `libtestdriver1-list-vars.cmake`) if `generate_test_driver.py` is modified.

gilles-peskine-arm (approved):
Thanks, LGTM

ronald-cron-arm:
@minosgalanakis please have a look to the last version, thanks.

minosgalanakis (approved):
LGTM

Cheked the [range after ](https://github.com/Mbed-TLS/TF-PSA-Crypto/compare/128ac9724773122e8002f7d340aa0be5a404cccd..26c5b39f12ba3a6f0c25c54b08c327c09ca7e8cb) and verified the rebase locally using  `git range-diff HEAD~19..HEAD bd41097f535607cefadec77dbca9d557c2240aec~19..bd41097f535607cefadec77dbca9d557c2240aec`

Issue #579, opened 2025-11-20
Reorganize everest and p256-m files


gilles-peskine-arm:
The goal of this task is to reorganize the files under `drivers/everest` and `drivers/p256-m` so that they follow the same general structure as `drivers/builtin`. Possibly also restructure `drivers/builtin` if it looks like we'd like a different structure there.

The motivation is twofold:

* Simplify our build and test scripts so that we don't need a lot of special cases for each driver. For example [fine-tuning of everest header selections in `check_names.py`](https://github.com/Mbed-TLS/mbedtls-framework/pull/236#discussion_r2541352044). See also the comment at the top of `./drivers/builtin.cmake`.
* Have a clear structure that third-party driver writers can follow for easy integration. Note that facilities and documentation for integrating third-party drivers is out of scope here. What is in scope is to keep this in mind as a future goal when doing the design.

I think that having all installed headers under `.../include` and all source files (including internal headers) under `.../src` is a good start. But we may also want to try to define some rules about upstream vs integration files. For example, can we have some systematic rules about which files `check_names.py` should run on?

Definition of done:

* Document the ideas behind the structure somewhere. It can be just a comment on this issue.
* Reorganize the files.
* Adapt our scripts accordingly, looking for simplifications (try to have `drivers/*/<pattern>` rather than one case for each driver).

PR #587, opened 2025-12-02
Add support for driver dispatch testing 1/2

### LINKED ISSUE #61: Add support for libtestdriver1

ronald-cron-arm:
## Description
Fixes #61
Precedes #586

## PR checklist
- [x] **framework PR** provided Mbed-TLS/mbedtls-framework#243
- [x] **mbedtls 3.6 PR**  not required because: TF-PSA-Crypto only work


ronald-cron-arm:
I've moved the framework update to #586 as we do not need it in this PR.

drivers/builtin/include/mbedtls/private/crypto_builtin_composites.h
```diff
@@ -20,8 +20,8 @@
  *  SPDX-License-Identifier: Apache-2.0 OR GPL-2.0-or-later
  */
 
-#ifndef PSA_CRYPTO_BUILTIN_COMPOSITES_H
-#define PSA_CRYPTO_BUILTIN_COMPOSITES_H
+#ifndef TF_PSA_CRYPTO_MBEDTLS_CRYPTO_BUILTIN_COMPOSITES_H
```
gilles-peskine-arm: Technically out of scope, but since these headers are being moved anyway, shouldn't we move them to the `private` subdirectory?

gilles-peskine-arm: Debatable: since these headers are being moved anyway, should they move to `tf-psa-crypto/*` rather than `mbedtls/*`?

include/tf-psa-crypto/private/crypto_adjust_config_support.h
```diff
@@ -208,4 +208,9 @@
 #define MBEDTLS_PK_PARSE_EC_COMPRESSED
 #endif
 
+#if defined(MBEDTLS_PK_PARSE_EC_EXTENDED) || \
+    defined(MBEDTLS_PK_PARSE_EC_COMPRESSED)
+#define MBEDTLS_ECP_LIGHT
```
gilles-peskine-arm: The documentation of `crypto_adjust_config_support.h` states

> Activate parts of support modules

Here it's the other way round: activating some built-in crypto based on the needs of a high-level module.

This doesn't exactly fit the role of `crypto_adjust_config_tweak_builtins.h` either, although I think it's a little closer:

> Automatically enable certain parts of the cryptography implementation that are required by other parts.

It's much closer to `crypto_adjust_config_enable_builtins.h`:

> Activate built-in implementations of cryptographic mechanisms as needed to fulfill the needs of the PSA configuration.

Here it's not based on the PSA configuration, but on an extension that lives above PSA, so I think that's the right place.

There are no practical consequences because none of the config adjust headers queries `MBEDTLS_ECP_LIGHT`, but the ordering is tricky so I'd prefer to stick to general patterns.

scripts/data_files/driver_templates/psa_crypto_driver_wrappers.h.jinja
```diff
@@ -1563,20 +1563,30 @@ static inline psa_status_t psa_driver_wrapper_aead_encrypt(
 #endif /* PSA_CRYPTO_DRIVER_TEST */
 #endif /* PSA_CRYPTO_ACCELERATOR_DRIVER_PRESENT */
 
-            /* Fell through, meaning no accelerator supports this operation */
+            /* If software fallback is compiled in, try fallback */
```
gilles-peskine-arm: Not just software fallback: also if there's no accelerator. The previous comment was correct: either no accelerator was included in the build, or there was an accelerator but it returned `PSA_ERROR_NOT_SUPPORTED`.

core/CMakeLists.txt
```diff
@@ -110,13 +110,13 @@ endif(NOT USE_STATIC_TF_PSA_CRYPTO_LIBRARY AND NOT USE_SHARED_TF_PSA_CRYPTO_LIBR
 set(tfpsacrypto_target "${TF_PSA_CRYPTO_TARGET_PREFIX}tfpsacrypto")
 set(builtin_target "${TF_PSA_CRYPTO_TARGET_PREFIX}builtin")
 set(everest_target "${TF_PSA_CRYPTO_TARGET_PREFIX}everest")
-set(p256m_target "${TF_PSA_CRYPTO_TARGET_PREFIX}p256m")
+set(p256-m_target "${TF_PSA_CRYPTO_TARGET_PREFIX}p256m")
```
gilles-peskine-arm: The commit e1ffd607126ddb59450ffe610bbecb53e7ed1e36 "cmake: p256-m: Align target and directory name" puzzled me for a while, mostly because I don't know CMake well but also because (as I found out) it is incorrect (running `cmake` fails). (The next commit "cmake: Introduce list of drivers" fixes the problem.)

If I understand correctly, the targets are defined by `drivers/*/CMakeLists.txt`, e.g. before your changes `p256m` is a target because `drivers/p256-m/CMakeLists.txt` contains `add_library(${target} …)` where `${target}` iterates over a list that contains `${p256m_target}`. Then e1ffd607126ddb59450ffe610bbecb53e7ed1e36 changes this target name to `p256-m`. But in `core/CMakeLists.txt`, only the variables `p256-m_target` and `p256-m_static_target` are renamed, not the targets.

I'm not sure it's worth rewriting the history for that, but it would be nicer for both history readers and git bisectors to make the commit rename the target consistently.

CMakeLists.txt
```diff
@@ -350,8 +350,8 @@ endif()
 
 add_subdirectory(framework)
 add_subdirectory(include)
-add_subdirectory(core)
 add_subdirectory(drivers)
+add_subdirectory(core)
```
gilles-peskine-arm, line 361: `core` needs to be after `drivers` because `TF_PSA_CRYPTO_DRIVERS` is defined in `drivers/CMakeLists.txt` and used in `drivers/CMakeLists.txt`, right? This could use a comment. Someone might want to do the opposite move because they would like to define a variable in `core` and use it in drivers.

drivers/CMakeLists.txt
```diff
@@ -1,3 +1,13 @@
+set(drivers "builtin;everest;p256-m")
```
gilles-peskine-arm: Maybe a little easier to maintain and patch?

```suggestion
set(drivers
    builtin
    everest
    p256-m
    )
```

gilles-peskine-arm: Why are there separate variables `drivers` and `TF_PSA_CRYPTO_DRIVERS`?

drivers/CMakeLists.txt
```diff
@@ -1,3 +1,13 @@
...
+# Directories containing the headers exposed by drivers
+set(TF_PSA_CRYPTO_DRIVERS_INCLUDE_DIRS
```
gilles-peskine-arm, line 12: It's not clear to me what the exact intent of this variable is: headers exposed to other drivers? exposed to the core? exposed to library users (and thus installed)?

drivers/CMakeLists.txt
```diff
@@ -1,3 +1,13 @@
...
+        ${CMAKE_CURRENT_SOURCE_DIR}/builtin/include
+        ${CMAKE_CURRENT_SOURCE_DIR}/everest/include
+        ${CMAKE_CURRENT_SOURCE_DIR}/p256-m)
```
gilles-peskine-arm: In the long term, I'd expect this to be collected from `*/CMakeLists.txt`, right? Perhaps with `${subdir}/include` as a default setting. Just checking my understanding of the design, it's out of scope here.

drivers/driver.cmake
```diff
@@ -0,0 +1,63 @@
+
```
gilles-peskine-arm, line 16: Please document this file: how it's meant to be used what variables it expects to be set, and what it's expected to do.

If I understand correctly, this file is meant for drivers that follow the same general structure as ours, to be included by `drivers/*/CMakeLists.txt` after setting a few variables?

drivers/driver.cmake
```diff
@@ -0,0 +1,63 @@
...
+if(CMAKE_COMPILER_IS_GNUCC)
+    set(LIBS_C_FLAGS -Wmissing-declarations -Wmissing-prototypes)
+endif(CMAKE_COMPILER_IS_GNUCC)
+
+if(CMAKE_COMPILER_IS_CLANG)
+    set(LIBS_C_FLAGS -Wmissing-declarations -Wmissing-prototypes -Wdocumentation -Wno-documentation-deprecated-sync -Wunreachable-code)
+endif(CMAKE_COMPILER_IS_CLANG)
```
gilles-peskine-arm, line 23: Setting warning flags seems to be about the quality and conventions of a particular part of the code, so it feels out of place in the generic file.

I'm not sure about the MSVC part below.

drivers/driver.cmake
```diff
@@ -0,0 +1,63 @@
...
+endif()
+
+if(CMAKE_C_COMPILER_ID MATCHES "AppleClang")
+    set(CMAKE_C_ARCHIVE_CREATE   "<CMAKE_AR> Scr <TARGET> <LINK_FLAGS> <OBJECTS>")
+    set(CMAKE_C_ARCHIVE_FINISH   "<CMAKE_RANLIB> -no_warning_for_no_symbols -c <TARGET>")
+endif()
+if(CMAKE_CXX_COMPILER_ID MATCHES "AppleClang")
+    set(CMAKE_CXX_ARCHIVE_CREATE "<CMAKE_AR> Scr <TARGET> <LINK_FLAGS> <OBJECTS>")
+    set(CMAKE_CXX_ARCHIVE_FINISH "<CMAKE_RANLIB> -no_warning_for_no_symbols -c <TARGET>")
+endif()
```
gilles-peskine-arm: This looks like a generic consideration about how the project is built, and there's a similar block in `core/CMakeLists.txt`. So shouldn't it be in the root `CMakeLists.txt`?

drivers/driver.cmake
```diff
@@ -0,0 +1,63 @@
...
+              ${TF_PSA_CRYPTO_DRIVERS_INCLUDE_DIRS})
+    tf_psa_crypto_set_config_files_compile_definitions(${target})
+endforeach(target)
+
+if(USE_SHARED_TF_PSA_CRYPTO_LIBRARY)
+    set_property(TARGET ${${tf_psa_crypto_driver}_target} PROPERTY POSITION_INDEPENDENT_CODE ON)
+endif(USE_SHARED_TF_PSA_CRYPTO_LIBRARY)
+
+if(INSTALL_TF_PSA_CRYPTO_HEADERS)
+  install(DIRECTORY "include/${tf_psa_crypto_driver_include_dir}"
```
gilles-peskine-arm: Why does it have to be one specific, explicit subdirectory of `include`? Why not take them all? It would make sense to me that a driver would have both `include/tf-psa-crypto/…` for glue code and `include/acme/…` for generic vendor code.

drivers/driver.cmake
```diff
@@ -0,0 +1,63 @@
...
+if(CMAKE_COMPILER_IS_GNUCC)
+    set(LIBS_C_FLAGS -Wmissing-declarations -Wmissing-prototypes)
+endif(CMAKE_COMPILER_IS_GNUCC)
+
+if(CMAKE_COMPILER_IS_CLANG)
+    set(LIBS_C_FLAGS -Wmissing-declarations -Wmissing-prototypes -Wdocumentation -Wno-documentation-deprecated-sync -Wunreachable-code)
+endif(CMAKE_COMPILER_IS_CLANG)
+
+if(CMAKE_COMPILER_IS_MSVC)
+    option(MSVC_STATIC_RUNTIME "Build the libraries with /MT compiler flag" OFF)
```
gilles-peskine-arm: What's the scope of an `option`? If it's global (which I guess it would be since it can be set on the command line when you run `cmake`, how does it work if `driver.cmake` is included by multiple drivers?

gilles-peskine-arm (changes requested):
Mostly LGTM, but I'm a bit doubtful about a few things in cmake.

drivers/builtin/include/mbedtls/private/crypto_builtin_composites.h
ronald-cron-arm: I am not completely sure about what you mean here. If you mean `include/tf-psa-crypto` I would say no. The headers in question declare the operation structure of the built-in driver thus they should be in `drivers/builtin/include` to me.
Maybe you mean in `drivers/builtin/include/tf-psa-crypto/` a new sub-directory in `drivers/builtin/include` but then this feels quite out of scope to me.

ronald-cron-arm: I moved them in `drivers/builtin/include` and not `drivers/builtin/include/private` because they declare the structures that are exposed by the built-in driver to the core and the other drivers, not to the application though. But I do not have a strong opinion about it. I can move them to `drivers/builtin/include/private`. Let me know if you would prefer it that way, I am not doing to change right now.

core/CMakeLists.txt
ronald-cron-arm: Indeed, I have fixed the commit. The history rewrite can be handily checked with git range-diff.

drivers/CMakeLists.txt
ronald-cron-arm: drivers is just local to `drivers/CMakeLists.txt` thus lowercase.
TF_PSA_CRYPTO_DRIVERS is only defined in the parent scope, the scope of the top-level CMake.
The case of TF_PSA_CRYPTO_DRIVERS_INCLUDE_DIRS is different as we also use it in the drivers CMakeLists.txt, thus the uppercase to emphasize this is not only local.

drivers/CMakeLists.txt
ronald-cron-arm: I've tried to improve the comment.

drivers/CMakeLists.txt
ronald-cron-arm: Yes that would the idea, especially when we have addressed or even when addressing https://github.com/Mbed-TLS/TF-PSA-Crypto/issues/579. I've added a reference to TF_PSA_CRYPTO_DRIVERS and TF_PSA_CRYPTO_DRIVERS_INCLUDE_DIRS in the issue.

drivers/builtin/include/mbedtls/private/crypto_builtin_composites.h
gilles-peskine-arm: The `private` subdirectories are for headers that need to be installed because they declare something that needs to be visible when compiling applications (typically structs and other types), but the headers are not meant to be included directly. I don't think a driver would ever `#include <mbedtls/crypto_builtin_composites.h>`, would it? We don't make any promise about its content, after all.

gilles-peskine-arm: Yes, I meant that the headers would be included with `#include <tf-psa-crypto/…>` (more specifically `#include <tf-psa-crypto/private/…>` per my other comment above). Sure, it's not strictly necessary, but since the header is getting moved anyway, I figure we could put it in the location where it would be if we were creating it today.

drivers/driver.cmake
ronald-cron-arm: I am just removing this as we do not create driver static libraries.

drivers/driver.cmake
ronald-cron-arm: I've reduced the scope of `driver.cmake` to builtin and generated test drivers (renamed builtin.cmake). With that scope reduction it is fine to keep this in this file.

drivers/builtin/include/mbedtls/private/crypto_builtin_composites.h
ronald-cron-arm: Then if I understand correctly we would have two `tf-psa-crypto` "include directories":
- `include/tf-psa-crypto`
- `drivers/builtin/include/tf-psa-crypto`
and when we do `#include <tf-psa-crypto/...>` it is about one of them.
Right?

gilles-peskine-arm: Right. Actually, we already have two, since there's also `drivers/everest/include/tf-psa-crypto`, so that would make it three. And likewise we already have both `include/mbedtls` and `drivers/builtin/include/mbedtls` that feed `#include <mbedtls/…>`.

ronald-cron-arm: I am not sure how desirable this is. I did not notice `drivers/everest/include/tf-psa-crypto` before. Naively I was expecting it to be `drivers/everest/include/everest`. For the time being, I am going to move the files to `include/mbedtls/private`.

drivers/driver.cmake
ronald-cron-arm: I've aligned with how include is installed.

drivers/driver.cmake
ronald-cron-arm: The scope of an option is the whole project. It is not a problem if it is defined several times I think but that's surely quite ugly. I've moved MSVC_STATIC_RUNTIME option to the top CMakeLists.txt as it is relevant to core and drivers. I have move the code activated by the option to `drivers/CMakeLists.txt` as it concerns all drivers.

include/tf-psa-crypto/private/crypto_adjust_config_support.h
ronald-cron-arm: okay.  I am just reverting the commit "Move definition of MBEDTLS_ECP_LIGHT for PK_PARSE" then. I am not really enthusiast to move it to  crypto_adjust_config_enable_builtins.h. It is not a PSA configuration as you mentionned. The only consequence of not having this change is I think that with the default config ecp.c is always there is the test driver. That is probably not a big deal.

ronald-cron-arm:
@gilles-peskine-arm this is eventually ready for re-review, sorry for the delay. Some changes in the CMake code took me quite some times. 

Compared to the reviewed head (https://github.com/Mbed-TLS/TF-PSA-Crypto/commit/6b601f9078a69fb8dff7e1145ade21a490981a73), I have first fixed the commit (cmake: p256-m: Align target and directory name), added new commits. Then the CI was failing for reasons not related to the changes in this PR and I rebased on top of development and eventually updated the framework pointer (see last commit). The rebase was probably not necessary but it can be check easily with:
```
git range-diff 6b601f9078a69fb8dff7e1145ade21a490981a73~9..6b601f9078a69fb8dff7e1145ade21a490981a73 a548d3fadcba59f35ba4e4ed068874346c207349~22..a548d3fadcba59f35ba4e4ed068874346c207349~13
```
https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/587/commits/a548d3fadcba59f35ba4e4ed068874346c207349 being the new head.

drivers/builtin/src/psa_crypto_cipher.c
```diff
@@ -224,6 +230,12 @@ psa_status_t mbedtls_cipher_values_from_psa(
 
     return mbedtls_cipher_validate_values(alg, key_type);
 }
+#endif /* MBEDTLS_PSA_BUILTIN_CIPHER) ||
```
minosgalanakis, line 264: This is wrapping everything in the above 200 lines including the `mbedtls_cipher_values_from_psa`

mbedtls_cipher_values_from_psa is however called by code that will be compiled in when `MBEDTLS_CIPHER_C` is defined like in the `const mbedtls_cipher_info_t *mbedtls_cipher_info_from_psa()`

This feels a bit like undefined behaviour.

minosgalanakis (commented):
Looks good overall, only one minor detail that would like addressed before aproving

drivers/builtin.cmake
```diff
@@ -1,3 +1,11 @@
+# CMake module for building built-in and generated test drivers.
```
gilles-peskine-arm: I'm surprised that this is not meant for anything other than `builtin` and its copies. I thought that this new file was unifying things that are currently duplicated in `everest` and `p256-m`, or have to be different there because we have a different directory structure for historical reasons.

As an example, in https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/593, I've copied parts of `drivers/builtin/CMakeLists.txt` to make a [`CMakeLists.txt` for a new driver](https://github.com/gilles-peskine-arm/TF-PSA-Crypto/blob/6aef476a1197d7d603cf29b652b44a769e5cb23e/drivers/pqcp/CMakeLists.txt). I thought that after your PR (and perhaps a little more work in the same direction), I could simply do something like
```
set(tf_psa_crypto_driver pqcp)
file(GLOB pqcp_src_files RELATIVE ${CURRENT_CMAKE_SOURCE_DIR} src/*.c)
include(${CMAKE_CURRENT_SOURCE_DIR}/../driver.cmake)
```

But apparently that's no longer a possibility? Why not?

CMakeLists.txt
```diff
@@ -405,9 +405,16 @@ if(ENABLE_TESTING OR ENABLE_PROGRAMS)
...
                 include
-                ${TF_PSA_CRYPTO_DRIVERS_INCLUDE_DIRS}
                 core
                 drivers/builtin/src)
+
+    foreach(driver_target ${TF_PSA_CRYPTO_DRIVER_TARGETS})
+        get_target_property(public_includes ${driver_target} INTERFACE_INCLUDE_DIRECTORIES)
+        if(public_includes)
+            target_include_directories(tf_psa_crypto_test PUBLIC ${public_includes})
+        endif()
```
gilles-peskine-arm: Why not just
```
target_include_directories(tf_psa_crypto_test PUBLIC ${public_includes})
```
? I don't see why we would not do it if e.g. the public include list is e.g. `include/foo/HELLO;include/foo/SOMETHING-NOTFOUND`. And `target_include_directories` with an empty list of items works, doesn't it?

Applies in `core/CMakeLists.txt` as well.

drivers/builtin/src/psa_crypto_cipher.c
gilles-peskine-arm: Good point. No undefined behavior, but this would break the build in a configuration where `mbedtls_cipher_info_from_psa` is defined but `mbedtls_cipher_values_from_psa` is not defined. It would be an unusual configuration. I think the only case is if built-in CMAC is enabled (notice that `mbedtls_cipher_info_from_psa()` is called in `psa_crypto_mac.c`) but no built-in cipher or AEAD.

gilles-peskine-arm (changes requested):
Almost LGTM, but I don't understand the restricted scope of `drivers/*.cmake`, plus the edge case that Minos found.

drivers/builtin.cmake
ronald-cron-arm: That's the goal. I have restricted it for the time being to `builtin` and its copies for the time being as the code in `everest` and `p256-m` is not ready for that probably. I am adding a comment and link to it in  https://github.com/Mbed-TLS/TF-PSA-Crypto/issues/579.

gilles-peskine-arm: I know that everest and p256-m are not ready, but if it's the goal, why not call it `driver.cmake`? And once this PR is merged, should I use it for a new driver or not?

ronald-cron-arm: okay then I name it `driver.cmake` and add a comment saying that we are not using it currently for everest and p256-m but that we should use it for new drievrs.

CMakeLists.txt
ronald-cron-arm: Yes it seems safe to simplify here but not in `core/CMakeLists.txt` as there the list is inside a generator expression: "$<BUILD_INTERFACE:..>". I've rewritten that part to be even more robust regarding paths with spaces.

drivers/builtin/src/psa_crypto_cipher.c
ronald-cron-arm: Good catch. I've made a change and written a novel.

ronald-cron-arm:
@gilles-peskine-arm @minosgalanakis this is ready for re-review, thanks.

drivers/builtin/src/psa_crypto_cipher.c
```diff
@@ -230,6 +230,37 @@ psa_status_t mbedtls_cipher_values_from_psa(
...
+    psa_key_type_t key_type,
+    mbedtls_cipher_mode_t *mode,
+    mbedtls_cipher_id_t *cipher_id)
+{
+    (void) alg;
+    (void) key_type;
+    (void) mode;
+    (void) cipher_id;
+
+    return PSA_ERROR_NOT_SUPPORTED;
```
gilles-peskine-arm, line 262: This seems weird to me, but it does match what the code is currently doing. I thought that `mbedtls_cipher_values_from_psa` would need to return actual data when built-in CMAC is enabled, but no built-in cipher, with the expectation that e.g. AES is enabled but only accelerated. I guess we don't support that configuration, or there's a preexisting bug that nobody has complained about. If there's a bug, it means a systematic runtime failure of CMAC in an exotic configuration where someone specifically wants CMAC, so anybody who cares would have noticed. Either way, at this point, I'm satisfied that this patch isn't making things worse.

gilles-peskine-arm (approved):
LGTM

minosgalanakis (approved):
LGTM

Thank you for addressing those comments so promptly

PR #600, opened 2025-12-11
[tf-psa-crypto] tests: pk: add a common function to create a PSA key out of predefined keys

### LINKED ISSUE #598: tests: pk: add a common function to create a PSA key out of predefined keys
### Suggested enhancement

This function already exists in [test_suite_pk.function](https://github.com/Mbed-TLS/TF-PSA-Crypto/blob/348d0eeb40bfc053225d4f1da69b60644c1d3110/tests/suites/test_suite_pk.function#L179) and, after #577, a similar one is going to exist also in `test_suite_pkwrite.function`.

The goal is to create a common helper function to be used in all `test_suite_xxx` files when necessary.

### Justification

TF-PSA-Crypto needs this because it's nice to minimize duplications also in test code.

valeriosetti:
## Description

Resolves #598 
Depends on https://github.com/Mbed-TLS/mbedtls-framework/pull/257

## PR checklist

- [x] **changelog** not required because: only test changes
- [x] **framework PR** provided https://github.com/Mbed-TLS/mbedtls-framework/pull/257
- **tests**  provided


bjwtaylor (dismissed):

tests/suites/test_suite_pk.function
```diff
@@ -239,8 +206,10 @@ static int pk_setup(mbedtls_pk_context *pk, mbedtls_pk_type_t pk_type, int curve
 
     if ((pk_type == MBEDTLS_PK_RSA) || (pk_type == MBEDTLS_PK_RSASSA_PSS)) {
 #if defined(PSA_WANT_KEY_TYPE_RSA_PUBLIC_KEY)
-        TEST_EQUAL(get_predefined_key_data(0, curve_or_keybits, &key_data, &key_data_len,
-                                           &pub_key_data, &pub_key_data_len), 0);
+        mbedtls_pk_helpers_get_predefined_key_data(PSA_KEY_TYPE_RSA_KEY_PAIR, curve_or_keybits,
```
gilles-peskine-arm: Here and elsewhere: if the function fails, the test case is marked as fail, but the execution will continue until something else notices a problem. It would be better to return early (or goto exit if there's something to clean up). That's not critical, since it only matters if the test case has already failed. But it would help the next person who's trying to understand how the test is failing.

gilles-peskine-arm (dismissed):
LGTM, although the behavior on failure is fragile and I'd prefer to properly stop on failure.

tests/suites/test_suite_pk.function
valeriosetti: Yeah, sorry. This dates back to the original implementation of `mbedtls_pk_helpers_get_predefined_key_data` which didn't returned any value. I will update this immediately

tests/suites/test_suite_pk.function
```diff
@@ -377,30 +360,13 @@ static psa_status_t pk_psa_setup(psa_key_type_t type, size_t bits,
...
-#if defined(PSA_WANT_KEY_TYPE_ECC_PUBLIC_KEY)
-        mbedtls_ecp_group_id grp_id;
-        grp_id = mbedtls_ecc_group_from_psa(PSA_KEY_TYPE_ECC_GET_FAMILY(type), bits);
-        TEST_EQUAL(get_predefined_key_data(1, grp_id, &key_data, &key_data_size,
-                                           &pub_key_data, &pub_key_data_size), 0);
-#else /* PSA_WANT_KEY_TYPE_ECC_PUBLIC_KEY */
-        TEST_FAIL("EC keys are not supported");
-#endif /* PSA_WANT_KEY_TYPE_ECC_PUBLIC_KEY */
+    if (PSA_KEY_TYPE_IS_RSA(type) || PSA_KEY_TYPE_IS_ECC(type)) {
+        mbedtls_pk_helpers_get_predefined_key_data(type, bits, &key_data, &key_data_size);
```
gilles-peskine-arm: This is still missing the value return check. Otherwise LGTM

gilles-peskine-arm (approved):
LGTM

bjwtaylor (approved):

Issue #62, opened 2023-09-26
Add acceleration test components and associated reference components


ronald-cron-arm:
Add in TF-PSA-Crypto the equivalent of the test_psa_crypto_config_accel.* and test_psa_crypto_config_reference.* components.

`test_psa_crypto_config_accel_ecdsa` not in scope of this issue (part of #61).

Depends on #61.

PR #612, opened 2025-12-15
Add dispatch testing of ECDSA and ECDH to a driver

ronald-cron-arm:
## Description
This PR progresses https://github.com/Mbed-TLS/TF-PSA-Crypto/issues/62.
~~Depends on: #586~~

## PR checklist
- [x] **mbedtls 3.6 PR** not required because: TF-PSA-Crypto only work


ronald-cron-arm:
I've re-based on top of development following the merge of #586.

core/CMakeLists.txt
gilles-peskine-arm, line 1: Preexisting: why not take `*.c`?

In Mbed TLS, we enumerated the crypto files because we wanted to control which files went into each library (crypto, x509, tls). But here this doesn't apply.

core/constant_time.c
gilles-peskine-arm, line 1: I disagree with the move from `constant_time.c` to `constant_time_internal.c`. The reason most functions in this file are not public is that it's an internal API. However, many of these functions are used in Mbed TLS, and unlike most uses of internal functions in Mbed TLS, this is here to stay, it's not some leftover from having rushed the split.

The constant-time functions may also be used by third-party drivers, although at their own risk since we don't promise API stability for them. When we create the directory for support modules like platform and base64, `constant_time.c` should also move there. In the meantime, `constant_time.c` and `constant_time_internal.h` should move to `core`, with no splitting.

I guess it would be more accurate to move `drivers/builtin/src/constant_time_internal.h` to `include/tf-psa-crypto/private/constant_time_experimental.h`, with documentation stating that these are unstable interfaces that may change at any time. That's not at all urgent. I think it won't matter until we start supporting running Mbed TLS with a newer minor version of TF-PSA_Crypto: at that time we will need to stabilize these interfaces to some extent.

drivers/builtin/src/bignum.c
```diff
@@ -2370,6 +2370,70 @@ int mbedtls_mpi_gen_prime(mbedtls_mpi *X, size_t nbits, int flags,
 
 #endif /* MBEDTLS_GENPRIME */
 
+
+#if defined(MBEDTLS_ASN1_WRITE_C) || defined(MBEDTLS_ASN1_PARSE_C) || \
+    defined(PSA_HAVE_ALG_SOME_ECDSA)
```
gilles-peskine-arm: These are not the same guards as before the move. Before the move, in addition to `MBEDTLS_BIGNUM_C` for both, `mbedtls_asn1_get_mpi` was guarded by `MBEDTLS_ASN1_PARSE_C`, and `mbedtls_asn1_write_mpi` was guarded by `MBEDTLS_ASN1_WRITE_C`.

I think ~neither~ the old guards were correct. Let's check who consume these functions:

* pkwrite/pkparse consumes ASN.1 in the obvious direction (for the mpi function: only pkparse) and `*check_config.h` has the corresponding check.
* X.509 consumes ASN.1 in the obvious direction and `*check_config.h` has the corresponding check. (X.509 also consumes a few functions in the non-obvious direction, which is why parts of `asn1write.c` and `asn1parse.c` are enabled even if only the opposite direction is enabled, but that doesn't apply to the mpi functions.)
* RSA consumes ASN.1 in both directions and `*adjust*` auto-enables them.
* For built-in ECDSA, `*adjust*` auto-enables ASN.1 in both directions.

So there's nobody that's trying to use `mbedtls_asn1_get_mpi` without `MBEDTLS_ASN1_PARSE_C`, or `mbedtls_asn1_write_mpi` without `MBEDTLS_ASN1_WRITE_C`.

tests/scripts/components-configuration-crypto.sh
```diff
@@ -11,6 +11,28 @@
...
+component_test_accel_ecdsa() {
+    msg "build: accelerated ECDSA"
+
+    # Build
+    # -----
+
+    cd $OUT_OF_SOURCE_DIR
+
+    cmake -DTF_PSA_CRYPTO_TEST_DRIVER=On \
+          -DTF_PSA_CRYPTO_USER_CONFIG_FILE="../tests/configs/user-config-accel-ecdsa.h" ..
```
gilles-peskine-arm, line 45: I appreciate the simplicity of having a ready-to-use file instead of a convoluted script. However, we are losing the ability to enumerate all the curves (and all the key pair sub-features). If we add a curve, it won't be enabled automatically unless we remember to add it to the config.

I'm not sure, is that ok? I think that if we forget a curve in the accel list, then the library will see that ECDSA can only be partially accelerated, and so `ecdsa.c.o` won't be empty, and we'll get a CI failure below. However, this is gray-box testing: we're assuming that the library implements this logic to enable built-in ECDSA if some curves are not accelerated. Does that matter? Are we validating that logic elsewhere?

tests/configs/user-config-accel-ecdsa.h
```diff
@@ -0,0 +1,18 @@
...
+#define MBEDTLS_PSA_ACCEL_KEY_TYPE_ECC_KEY_PAIR_GENERATE
+#define MBEDTLS_PSA_ACCEL_ECC_BRAINPOOL_P_R1_256
+#define MBEDTLS_PSA_ACCEL_ECC_BRAINPOOL_P_R1_384
+#define MBEDTLS_PSA_ACCEL_ECC_BRAINPOOL_P_R1_512
+#define MBEDTLS_PSA_ACCEL_ECC_MONTGOMERY_255
+#define MBEDTLS_PSA_ACCEL_ECC_MONTGOMERY_448
+#define MBEDTLS_PSA_ACCEL_ECC_SECP_K1_256
+#define MBEDTLS_PSA_ACCEL_ECC_SECP_R1_256
+#define MBEDTLS_PSA_ACCEL_ECC_SECP_R1_384
+#define MBEDTLS_PSA_ACCEL_ECC_SECP_R1_521
```
gilles-peskine-arm, line 16: To make it a little more visible that we have a file that's supposed to be complete, I'd like to have the key types and curves in a separate file `tests/configs/user-config-accel-ecc-all-types.h`, included here. Or perhaps `user-config-accel-all-ecc.h` and undefine ECDH and ECJPAKE.

Similar structure for `user-config-accel-ecdh.h` of course.

tests/configs/user-config-accel-ecdsa.h
```diff
@@ -0,0 +1,18 @@
...
+#define MBEDTLS_PSA_ACCEL_KEY_TYPE_ECC_PUBLIC_KEY
+#define MBEDTLS_PSA_ACCEL_KEY_TYPE_ECC_KEY_PAIR_BASIC
+#define MBEDTLS_PSA_ACCEL_KEY_TYPE_ECC_KEY_PAIR_IMPORT
+#define MBEDTLS_PSA_ACCEL_KEY_TYPE_ECC_KEY_PAIR_EXPORT
+#define MBEDTLS_PSA_ACCEL_KEY_TYPE_ECC_KEY_PAIR_GENERATE
+#define MBEDTLS_PSA_ACCEL_ECC_BRAINPOOL_P_R1_256
+#define MBEDTLS_PSA_ACCEL_ECC_BRAINPOOL_P_R1_384
+#define MBEDTLS_PSA_ACCEL_ECC_BRAINPOOL_P_R1_512
+#define MBEDTLS_PSA_ACCEL_ECC_MONTGOMERY_255
+#define MBEDTLS_PSA_ACCEL_ECC_MONTGOMERY_448
```
gilles-peskine-arm, line 12: This reproduces the current `component_test_psa_crypto_config_accel_ecdsa` in Mbed TLS.

But ECDSA doesn't do Curve25519 or Curve448, so I wonder if we should keep them unaccelerated here?

gilles-peskine-arm (changes requested):
I disagree with the split of `constant_time.c`. A few minor issues otherwise.

core/constant_time.c
ronald-cron-arm: fine by me, I will move constant_time.c and constant_time_internal.h to core.

tests/configs/user-config-accel-ecdsa.h
ronald-cron-arm: I have gone the `user-config-accel-all-ecc.h` way (just named user-config-accel-ecc.h there though) in the following PR where I felt the need to contain the number of `tests/configs/user-config-accel-xyz` files. `user-config-accel-ecc.h` defines all `MBEDTLS_PSA_ACCEL_` ECC macros and the acceleration components unset some of them depending on their needs.

tests/scripts/components-configuration-crypto.sh
ronald-cron-arm: As explained below, in the next PR I reduce to one the number of user config file needed for all ECC acceleration components: `user-config-accel-ecc.h`. The file defines all the possible  `MBEDTLS_PSA_ACCEL_` ECC macros. Then we could have a check component checking that `user-config-accel-ecc.h` is up to date. Wdyt?

gilles-peskine-arm: Sounds good.

To be clear, I'm not sure that it's a problem if the list goes out of date: I think it will fail safe, in the sense that AFAICT it will cause a CI failure which shouldn't be too hard to understand.

tests/configs/user-config-accel-ecdsa.h
ronald-cron-arm: By not accelerating them we test that this does not re-enable ecdsa.c.
But then we need to disable PK as otherwise, we hit:
#error "Unsupported partial support for ECC curves acceleration, see docs/driver-only-builds.md"

gilles-peskine-arm: Thanks for trying it! PK is probably the most important thing to test for the impact of having drivers for asymmetric crypto. So please keep PK enabled and Montgomery curves accelerated.

gilles-peskine-arm (changes requested):
LGTM at f983e9cd507013e9d24e765feb6ce334d487dbd4, although it would be nice to add a comment explaining why Montgomery curves need to be accelerated in `user-config-accel-ecdsa.h`.

I'm ok to force-push to remove b34dcd90c1b4ce89763853370a007f94ffedf2d5 from the history. Also ok to squash the reverted commits.

There's a [CI failure](https://ci.trustedfirmware.org/job/mbed-tls-tf-psa-crypto-multibranch/job/PR-612-head/4/pipeline-console/log?nodeId=2978) I don't understand in `tf_psa_crypto_build_custom_config_file`. It seems related to 16ed54500500d0fe92a04047d856472230275bbf, which is desirable but out of scope, so feel free to also remove that commit if you can't see an easy fix.

ronald-cron-arm:
> LGTM at [f983e9c](https://github.com/Mbed-TLS/TF-PSA-Crypto/commit/f983e9cd507013e9d24e765feb6ce334d487dbd4), although it would be nice to add a comment explaining why Montgomery curves need to be accelerated in `user-config-accel-ecdsa.h`.
> 
> I'm ok to force-push to remove [b34dcd9](https://github.com/Mbed-TLS/TF-PSA-Crypto/commit/b34dcd90c1b4ce89763853370a007f94ffedf2d5) from the history. Also ok to squash the reverted commits.
> 
> There's a [CI failure](https://ci.trustedfirmware.org/job/mbed-tls-tf-psa-crypto-multibranch/job/PR-612-head/4/pipeline-console/log?nodeId=2978) I don't understand in `tf_psa_crypto_build_custom_config_file`. It seems related to [16ed545](https://github.com/Mbed-TLS/TF-PSA-Crypto/commit/16ed54500500d0fe92a04047d856472230275bbf), which is desirable but out of scope, so feel free to also remove that commit if you can't see an easy fix.

I have removed b34dcd9 and 16ed545 and squashed the revert commits. I need to think more about 16ed545. I will add a comment in the next PR about the Montgomery curves where `user-config-accel-ecdsa.h` is removed.

ronald-cron-arm:
I've rebased from https://github.com/Mbed-TLS/TF-PSA-Crypto/commit/932baaff1c4bf9bcd2ecea1d382bb424b46a71b7 to https://github.com/Mbed-TLS/TF-PSA-Crypto/commit/cc6bd1f68e36a84312d436b22d377f7b67ffb765 to get a more recent framework as needed by Mbed TLS.
```git range-diff 932baaff1c4bf9bcd2ecea1d382bb424b46a71b7~14..932baaff1c4bf9bcd2ecea1d382bb424b46a71b7 cc6bd1f68e36a84312d436b22d377f7b67ffb765~14..cc6bd1f68e36a84312d436b22d377f7b67ffb765```
to check the rebase.

gilles-peskine-arm (approved):
LGTM at cc6bd1f68e36a84312d436b22d377f7b67ffb765

minosgalanakis (approved):
Looks good:

I can confirm that the code does what it is intended and moving things around with few tweaks:
- drivers/builtin/src/psa_util.c was split into core/psa_util.c drivers/builtin/
    src/psa_util_internal.c. The migrated code is unchanged.
- mbedtls_asn1_write_mpi and mbedtls_asn1_get_mpi were migrated from drivers/
    builtin/src/asn1write.c and asn1parse.c into drivers/builtin/src/bignum.c. The only change was the removal of the PSA_HAVE_ALG_SOME_ECDSA from the guards, which was redundant as discussed in https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/612/commits/038be0d6008b5c8f60f1b9995ec0f46d90b464b4#r2632813886
- Added new template configs and new test component to test the accelerated dispatch

PR #615, opened 2025-12-18
PK: remove `pk_can_do()` -- part 1

valeriosetti:
### TL;DR
Addresses the first part of #523: modify `mbedtls_pk_[sign|verify]_ext()` so that they don't use `mbedtls_pk_can_do()`.

### Details

Commits in this PR can be divided into 3 blocks:

1. Extending tests for `mbedtls_pk_[sign|verify]_ext()` in order to cover all possible cases
2. Re-shape `mbedtls_pk_[sign|verify]_ext()` so that they _really_ perform as an `_ext` function over the simpler `mbedtls_pk_[sign|verify]()` counterparts: i.e. try to sign|verify using the specified sigalg and PK context.
   * This removes a lot of "special case handling" that was done previously
   * This simplifies testing: many comments added at point (1) to explain why `[sign|verify]_ext()` was behaving differently than expected are removed.
   * For example, after this reshaping opaque RSA keys can perform signature verification while previously this was not possible. And this didn't require any special treatment for the opaque context: just treating all the PK context in the same way allowed to improve the functionality.
3. Adapt tests that are failing after this change
   * In most of the cases it's just a change to the returned value.
   * As mentioned above, in some other cases there was an improvement in functionality so a test that was failing for lack of support now is passing.
   * There are only changes to PK tests, meaning that "users" of `[sign|verify]_ext()` are not affected by this functions' reshaping! To me this is a sort of proof that these changes might be potentially transparent for the end user.

### Advantage of the reshape

Once this will be merged we can make `mbedtls_pk_[sign|verify]()` calling into their `_ext()` counterpart (instead of the restartable ones) as it it normally happens for other "standard"/"extended" functions in the library.

## PR checklist

- [ ] **changelog** not required because: `mbedtls_pk_can_do` is internal, so nothing to be notified
- [ ] **mbedtls not required because: no change there
- **tests**  provided


drivers/builtin/src/pk.c
```diff
@@ -1148,71 +1160,83 @@ int mbedtls_pk_verify_ext(mbedtls_pk_sigalg_t type,
...
 
-    if (type != MBEDTLS_PK_SIGALG_RSA_PSS) {
-        return mbedtls_pk_verify(ctx, md_alg, hash, hash_len, sig, sig_len);
+    /* First select attributes to associate with the key being imported into PSA. */
+    key_type = PSA_KEY_TYPE_PUBLIC_KEY_OF_KEY_PAIR(mbedtls_pk_get_key_type(ctx));
+    psa_set_key_type(&attributes, key_type);
+    psa_set_key_usage_flags(&attributes, PSA_KEY_USAGE_VERIFY_HASH);
+
+    if (PSA_KEY_TYPE_IS_RSA(key_type)) {
+        psa_set_key_algorithm(&attributes, PSA_ALG_RSA_PKCS1V15_SIGN(PSA_ALG_ANY_HASH));
```
bjwtaylor, line 1175: So if you build the branch with PSA_WANT_ALG_RSA_PKCS1V15_SIGN and MBEDTLS_KEY_EXCHANGE_ECDHE_RSA_ENABLED unset a number of the tests will fail. Which I think is related to this line. Is it expected that it will no longer work with this configuration.

valeriosetti: Ideally if `MBEDTLS_PSA_CRYPTO_C` is defined then RSA-PSS is still used as enrollment algorithm, so verification should still work if the signature to be verified is really RSA-PSS and not PKCS v1.5.
But where did you get that configuration? The CI is green, but now I suspect that I missed something :/

bjwtaylor: So, this is potentially the question that config is completely made up to check if this change causes any issues. I guess it should still build with those configs disabled though? I also built development with the same configs disabled and you don't get the same failures. So I guess it's possible a regression, though I'm not sure if this is part of a wider plan?

valeriosetti: I'm currently investigating this and it's a bit weird. I just built `mbedtls` using
```
scripts/config.py full
scripts/config.py unset PSA_WANT_ALG_RSA_PKCS1V15_SIGN
scripts/config.py unset MBEDTLS_KEY_EXCHANGE_ECDHE_RSA_ENABLED
```
as you suggested and there are indeed tests which are failing. For example
`X509 CRT verification #1 (Revoked Cert, Expired CRL, no CN)`
from `test_suite_x509parse.data`. The problem is that this test has a `depends_on` on `MBEDTLS_PKCS1_V15` which should be disabled based on my configuration, but it's executed anyway (and it fails, of course).
I need to investigate on this a bit further

valeriosetti: Oh wait, that's because you must disable _also_ `PSA_WANT_ALG_RSA_PKCS1V15_CRYPT`. Doing that the test passes. Apparently legacy crypto had a single configuration for PKCS v1.5 (sign and crypt) whereas PSA API differentiate between the two.

valeriosetti: @bjwtaylor I just tried this component on `mbedtls`'s `development` (+ `git submodule update --recursive`):
```
component_tmp() {
    msg "Configure"

    scripts/config.py full
    scripts/config.py unset PSA_WANT_ALG_RSA_PKCS1V15_SIGN
    scripts/config.py unset MBEDTLS_KEY_EXCHANGE_ECDHE_RSA_ENABLED

    rm -rf build
    mkdir build

    msg "Build"
    (
        cd build
        cmake -D CMAKE_BUILD_TYPE=Debug ..
        make
    )

    msg "Test"
    (
        cd build/tests
        ./test_suite_x509parse
    )
}
```
and it fails the same as with my PR checked out in tf-psa-crypto. So it seems my PR is not introducing regressions.
Can you please double check?

bjwtaylor: As discussed, I think this requires further investigation as it worked prior to the addition of these commits. If we need to also disable PSA_WANT_ALG_RSA_PKCS1V15_CRYPT this should probably be reflected in the config dependencies. What do you think? EDIT: hadn't seen this above, let me confirm and come back to you.

bjwtaylor (commented):
as per https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/615#discussion_r2768434259

drivers/builtin/src/pk.c
bjwtaylor: So, It's also failing for me on development now. So it could be something we've introduced recently, though doesn't look like a regression on this PR. So I'll leave it up to you and potentially the second reviewer, as to if this is worth fixing.

PR #618, opened 2025-12-22
SHA3 cleanup before adding SHAKE

### LINKED ISSUE #542: Builds with accelerated SHA3 still include the built-in implementation

gilles-peskine-arm:
* Fix mix-up of compile-time guards: since TF-PSA-Crypto 1.0, we were using `PSA_WANT` instead of `MBEDTLS_PSA_BUILTIN` in `sha3.c`. Fixes https://github.com/Mbed-TLS/TF-PSA-Crypto/issues/542.
* Split `test_suite_shax` into separate suites for each of `sha1.c`, `sha256.c`, `sha512.c` and `sha3.c`. They're unrelated except thematically.
* Split `test_suite_sha3.data` which was really long.
* Simplify test case and test function dependencies.

This is in preparation for adding SHAKE.

Needs preceding PR:

* [x] https://github.com/Mbed-TLS/mbedtls/pull/10548 to fix a CI failure due to bringing back the former option `MBEDTLS_SHA3_C` as an internal macro.
* [x] https://github.com/Mbed-TLS/mbedtls/pull/10554 to prepare for the SHA3 changes (mix-up of compile-time guards in `selftest.c`, move `component_test_sha3_variations`)
* [x] https://github.com/Mbed-TLS/mbedtls/pull/10556 to update references to `test_suite_shax` in TLS's `analyze_outcomes.py`

## PR checklist

- [x] **changelog** provided
- [ ] **mbedtls 3.6 PR** no bug of compile-time guards, but we should probably split the test suites in the same way
- **tests**  provided


drivers/builtin/include/mbedtls/private/crypto_adjust_config_tweak_builtins.h
```diff
@@ -30,6 +30,14 @@
  */
 #define MBEDTLS_USE_PSA_CRYPTO
 
+/* Whether any Keccak variant is enabled, i.e. the bulk of sha3.c. */
```
mpg, line 33: Hmm, I would have put this block in `enable_builtins.h`, because that's what we are doing: enabling a built-in module. Also, that's where all the other `MBEDTLS_SHAxxx_C` are defined, so why handle SHA3 differently?

tests/suites/test_suite_sha1.function
```diff
@@ -2,7 +2,11 @@
 #include "mbedtls/private/sha1.h"
 /* END_HEADER */
 
-/* BEGIN_CASE depends_on:MBEDTLS_SHA1_C */
+/* BEGIN_DEPENDENCIES
+ * depends_on:MBEDTLS_SHA1_C
+ * END_DEPENDENCIES */
+
+/* BEGIN_CASE depends_on */
```
mpg: Note: I didn't know this was valid syntax. I usually write this as `/* BEGIN_CASE */`. (Just thinking out loud, I'm happy if the CI is happy.)

tests/scripts/components-configuration-tuning.sh
```diff
@@ -0,0 +1,47 @@
...
+# Copyright The Mbed TLS Contributors
+# SPDX-License-Identifier: Apache-2.0 OR GPL-2.0-or-later
+
+# This file contains test components that are executed by all.sh
+
+################################################################
+#### Configuration Testing - Optimization tuning
+################################################################
+
+## test_sha3_variation NAME THETA PI CHI RHO
```
mpg: Minor: the comment should probably mention that the caller needs to set `targets` appropriately. Alternatively, just set it in this function in order to make it more self-contained. (Yes, that means globbing will be repeated, but I don't think we care, compared to a full build.)

tests/scripts/components-configuration-tuning.sh
```diff
@@ -0,0 +1,47 @@
...
+#define MBEDTLS_PSA_BUILTIN_ALG_SHA3_256 1
+#define MBEDTLS_PSA_BUILTIN_ALG_SHA3_384 1
+#define MBEDTLS_PSA_BUILTIN_ALG_SHA3_512 1
+#define MBEDTLS_SHA3_THETA_UNROLL $2
+#define MBEDTLS_SHA3_PI_UNROLL $3
+#define MBEDTLS_SHA3_CHI_UNROLL $4
+#define MBEDTLS_SHA3_RHO_UNROLL $5
+EOF
+
+    cmake -D CMAKE_BUILD_TYPE:String=Release -D TF_PSA_CRYPTO_CONFIG_FILE="$MBEDTLS_TEST_CONFIGURATION.h" "$TF_PSA_CRYPTO_ROOT_DIR"
```
mpg, line 35: Readability: are we confident / is it obvious to everyone in the team that we can change cmake variables when re-invoking it in an existing build dir? Otherwise, we may want to remove it first.

mpg (dismissed):
Looks good to me, just a few minor points/questions.

tests/scripts/components-configuration-tuning.sh
mpg: Or perhaps just add a comment.

tests/suites/test_suite_sha1.function
gilles-peskine-arm: I didn't know this was valid either. I just removed `:MBEDTLS_SHA1_C` from individual functions when it became a file dependency, and didn't realize that `depends_on` was left alone!

tests/scripts/components-configuration-tuning.sh
gilles-peskine-arm: I'm not a CMake expert by any means, and it's not a problem for me. Re-running `cmake` is how you reconfigure a build directory. So I don't think this warrants any special attention.

drivers/builtin/include/mbedtls/private/crypto_adjust_config_tweak_builtins.h
gilles-peskine-arm: The split between `enable_builtins.h` and `tweak_builtins.h` is recent and I don't think it has reached its final form. I can see three kinds of things:

* If `PSA_WANT_xxx` and not `MBEDTLS_PSA_ACCEL_xxx` then enable `MBEDTLS_PSA_BUILTIN_xxx`. That's the documented goal of `enable_builtins.h`.
* Given that some built-in feature is requested, which we know because `MBEDTLS_PSA_BUILTIN_xxx` is enabled, activate its constituent parts. This is mostly done in `enable_builtins.h` right now, but arguably it doesn't fit the description of that file (“Activate built-in implementations” is done as soon as `MBEDTLS_PSA_BUILTIN_xxx` is defined), whereas it mostly fits the description of `tweak_builtins.h` (“Also define some internal symbols that are derived from public ones” — `MBEDTLS_PSA_BUILTIN_xxx` isn't public but it's derived from public symbols following a general rule rather than some ad hoc definitions).
* Some built-in feature needs another built-in feature to work, e.g. the various cipher things that require `MBEDTLS_CIPHER_C`. That's the ostensible purpose of `tweak_builtins.h` both by looking at its content and by reading its description.

I think `MBEDTLS_SHA3_C` is closer to `MBEDTLS_CIPHER_C` since it's common not only to multiple SHA3 sizes, but also (in the next PR after this one) to SHAKE. That's why I put it in `tweak_builtins.h`. Arguably we should rename `sha3.c` to `keccak.c`, and it could be in scope for a “SHA3 cleanup”?

But I also see the value of consistency with other hashes, so I could go either way.

Maybe we should have three headers and not two but I don't want to go that far here, I'd prefer to stick to sha3.

mpg: Ok, let's not sweat it, and keep the code as it is, as there are argument both ways.

mpg (approved):
LGTM, thanks!

valeriosetti (approved):

PR #619, opened 2025-12-23
crypto: generate_config_checks.py: use current data

### LINKED ISSUE #249: generate_config_checks.py treats all old internal options as current internal options
`scripts/generate_config_checks.py` in TF-PSA-Crypto and Mbed TLS consider any internal option of the last major release to be an internal option of the current version. This doesn't make sense:

* A former internal option could become a public option. Then users would not be able to enable it (`*_check_config_user.h` would complain).
*  If an internal option of 1.0/4.0 is subsequently removed, it causes `check_names.py` to complain in `*_check_config_user.h` (if that generated file is present, which is not the case when it runs from `all.sh`, hence it doesn't affect the CI).

For example, in https://github.com/Mbed-TLS/TF-PSA-Crypto/commit/348d0eeb40bfc053225d4f1da69b60644c1d3110, `MBEDTLS_ECP_DP_SECP192K1_ENABLED` is no longer used anywhere. But since it was an internal option in TF-PSA-Crypto 1.0, `core/tf_psa_crypto_config_check_user.h` contains

```
#if defined(MBEDTLS_ECP_DP_SECP192K1_ENABLED)
#  error "MBEDTLS_ECP_DP_SECP192K1_ENABLED is not a valid configuration option."
#endif
```

More accurately, this should be considered a removed option (present in 3.6, internal in 1.0). And, also, the logic should allow for a formerly internal option to become public.

At this point it's not clear to me yet whether the bug is solely in the logic in `scripts/generate_config_checks.py`, or also in `config_checks_generator.py`. Also, I haven't checked the [spec](https://github.com/Mbed-TLS/mbedtls-framework/blob/main/docs/architecture/config-check-framework.md) yet.

gilles-peskine-arm:
Use current data rather than data from 1.0 to generate config checks. Fix https://github.com/Mbed-TLS/mbedtls-framework/issues/249

Needs preceding PR: submodule in https://github.com/Mbed-TLS/mbedtls-framework/pull/260

## PR checklist

- [x] **changelog** not required because: not user-visible
- [x] **framework PR** provided https://github.com/Mbed-TLS/mbedtls-framework/pull/260
- [x] **crypto PR** provided https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/619
- [x] **mbedtls development PR** provided https://github.com/Mbed-TLS/mbedtls/pull/10548
- [x] **mbedtls 3.6 PR** not required because: feature added in 1.0/4.0
- **tests**  provided


scripts/data_files/config-options-current.txt
```diff
@@ -0,0 +1,219 @@
+MBEDTLS_AESCE_C
```
gilles-peskine-arm, line 1: To regenerate this file, run
```
tests/scripts/check_option_lists.py -u
```

This command appears in the commit message that adds the file. It doesn't appear in the file itself because I wanted to keep it simple, so there's no room for a comment.

scripts/data_files/config-options-current.txt
```diff
@@ -0,0 +1,219 @@
...
+MBEDTLS_SHA512_SMALLER
+MBEDTLS_SHA512_USE_A64_CRYPTO_IF_PRESENT
+MBEDTLS_SHA512_USE_A64_CRYPTO_ONLY
+MBEDTLS_TEST_CONSTANT_FLOW_MEMSAN
+MBEDTLS_TEST_CONSTANT_FLOW_VALGRIND
+MBEDTLS_TEST_HOOKS
+MBEDTLS_THREADING_ALT
+MBEDTLS_THREADING_C
+MBEDTLS_THREADING_PTHREAD
+PSA_CRYPTO_CONFIG_H
```
gilles-peskine-arm, line 132: The presence of the header guard is preexisting. I didn't consider it a problem in https://github.com/Mbed-TLS/mbedtls-framework/pull/196 and I still don't. It has very little practical consequence: users will be allowed to define `PSA_CRYPTO_CONFIG_H` in their crypto config file, which is exactly what would happen if the macro was omitted when constructing the list of options. In Mbed TLS, users will not be allowed to define `PSA_CRYPTO_CONFIG_H` if the crypto config file doesn't define it, which is actually kind of desirable even though it doesn't really matter.

mpg (dismissed):
LGTM, thanks!

bjwtaylor (dismissed):

tests/scripts/check_option_lists.py
```diff
@@ -35,7 +35,11 @@ def main():
         data.update_shadow_file(args.always_update)
     else:
         up_to_date = True
-        up_to_date &= data.compare_shadow_file_verbosely()
+        if not data.is_shadow_file_up_to_date():
+            print(f'{data.shadow_file_path()} is out of date')
+            print(f'After adding or removing a conifg option, you need to run')
+            print(f'{sys.argv[0]} and commit the result.')
```
mpg: Shouldn't we suggest the `-u` option? Otherwise, the script I'm running (presumably without options) is telling me to run the same script again (apparently without options)...

tests/scripts/check_option_lists.py
```diff
@@ -35,7 +35,11 @@ def main():
         data.update_shadow_file(args.always_update)
     else:
         up_to_date = True
-        up_to_date &= data.compare_shadow_file_verbosely()
+        if not data.is_shadow_file_up_to_date():
+            print(f'{data.shadow_file_path()} is out of date')
+            print(f'After adding or removing a conifg option, you need to run')
```
mpg: typo: conIFg

mpg (changes requested):
LGTM overall, but the output message can be improved.

mpg (dismissed):
LGTM, thanks!

gilles-peskine-arm:
I rebased on top of the target branch to resolve the conflict with the framework submodule, and updated the submodule-updating commit accordingly. The previous history is at https://github.com/gilles-peskine-arm/TF-PSA-Crypto/tree/config-checks-generator-current-crypto-6

mpg (approved):
LGTM, thanks!

bjwtaylor (approved):

PR #624, opened 2025-12-30
Fix "Simplify interruptible ECDH dispatch"

mpg:
## Description

Follow-up to https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/578 - fixing a bug that was introduced there, and closing a pre-existing test gap.

## PR checklist

- [x] **changelog** not required because: bug was never released
- [x] **mbedtls 3.6 PR** not required because: development only (interruptible ECDH not in 3.6)
- **tests**  provided

### DIFF
--- a/tests/suites/test_suite_psa_crypto.function
+++ b/tests/suites/test_suite_psa_crypto.function
@@ -9807,6 +9807,7 @@ void key_agreement(int alg_arg,
     psa_algorithm_t alg = alg_arg;
     psa_key_type_t our_key_type = our_key_type_arg;
     psa_key_attributes_t attributes = PSA_KEY_ATTRIBUTES_INIT;
+    unsigned char *bad_peer_key = NULL;
     unsigned char *output = NULL;
     size_t output_length = ~0;
     size_t key_bits;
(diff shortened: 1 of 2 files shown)


tests/suites/test_suite_psa_crypto.function
```diff
@@ -10038,13 +10062,25 @@ void key_agreement_interruptible(int alg_arg,
 
     PSA_ASSERT(psa_key_agreement_iop_abort(&operation));
 
+    /* Setup with invalid input (peer_key) */
+    size_t bad_peer_key_len = peer_key_data->len - 1;
+    TEST_CALLOC(bad_peer_key, bad_peer_key_len);
+    memcpy(bad_peer_key, peer_key_data->x, bad_peer_key_len);
+    TEST_EQUAL(PSA_ERROR_INVALID_ARGUMENT,
+               psa_key_agreement_iop_setup(&operation, our_key,
```
gilles-peskine-arm, line 10070: Since we're looking at test coverage, I can't find tests of `psa_key_agreement_iop_setup` with certain other bad parameters (bad algorithm, bad attributes).

gilles-peskine-arm (approved):
LGTM for the bug fix and non-regression test. Maybe we're still not quite there for the test coverage, but that can be done separately, without delaying the major bug fix.

tests/suites/test_suite_psa_crypto.function
mpg: Agreed. The documentation is pretty good at listing error cases, it would provide a good reference for more comprehensive testing.

bjwtaylor (approved):

PR #648, opened 2026-01-09
Implement SHAKE in sha3.c

### LINKED ISSUE #621: Implement SHAKE (built-in, no API)
* Implement SHAKE128 and SHAKE256 through a low-level interface, with both one-shot and multipart support..
* Add unit tests.

This is already done in https://github.com/ARMmbed/mbedtls/pull/1549 but it require a major rebase.

Prerequisites:

* SHA-3 implementation: Mbed-TLS/mbedtls#3751
* [SHA-3 cleanup](https://github.com/Mbed-TLS/TF-PSA-Crypto/issues/542): https://github.com/Mbed-TLS/mbedtls/pull/10548

Foreseen extensions:

* Support SHAKE through the PSA key derivation API.

gilles-peskine-arm:
Implement SHAKE in the SHA3 module. Resolves https://github.com/Mbed-TLS/TF-PSA-Crypto/issues/621

An earlier version of this pull request also implemented the SHAKE128/256 and SHAKE256/512 hash algorithms. We don't need those urgently: they'll be needed for HashML-DSA and for Ed448ph respectively. That part is at https://github.com/gilles-peskine-arm/TF-PSA-Crypto/tree/sha3-shake-builtin-2 (with unresolved problems with some test driver builds) and needs preceding PR: https://github.com/Mbed-TLS/mbedtls-framework/pull/267.

## PR checklist

- [x] **changelog** no (nothing user-visible)
- **tests**  provided


drivers/builtin/src/sha3.c
```diff
@@ -297,12 +298,26 @@ int mbedtls_sha3_starts(mbedtls_sha3_context *ctx, mbedtls_sha3_id id)
             ctx->max_block_size = 576 / 8;
             break;
 #endif
+
+#if defined(MBEDTLS_SHA3_WANT_SHAKE128)
```
bjwtaylor, line 305: Should the SHAKE variants be setting the ctx->olen to 0, to clear it to prevent any prior length being retained?

bjwtaylor (commented):
One quick question, though still need to review in more detail when I get a chance.

drivers/builtin/src/sha3.c
gilles-peskine-arm: Yes, that's a good idea. And that shows we're missing a context reuse test case for SHAKE.

drivers/builtin/src/sha3.c
```diff
@@ -377,7 +399,15 @@ int mbedtls_sha3_finish(mbedtls_sha3_context *ctx,
         }
     }
 
-    ret = 0;
+    /* If there is a fixed output length (SHA3), we only accept a single call
+     * to finish(), and have it clean up the context. If the output length is
+     * variable (SHAKE), leave the context ready for further finish() calls. */
+    if (ctx->olen > 0) {
```
valeriosetti, line 404: Adding the comment here, but it refers to all other occurrences. Why not adding a new field in the `mbedtls_sha3_context` to hold the `mbedtls_sha3_id`? This way it wouldn't rely on the fact that `ctx->olen == 0` (which is also the same we have after `mbedtls_sha3_init`)

drivers/builtin/include/mbedtls/private/crypto_adjust_config_tweak_builtins.h
```diff
@@ -30,11 +30,24 @@
...
     defined(MBEDTLS_PSA_BUILTIN_ALG_SHA3_512)
+#define MBEDTLS_PSA_BUILTIN_ALG_SHA3_SOME_HASH
+#endif
+
+#if defined(MBEDTLS_PSA_BUILTIN_ALG_SHAKE128)
+#define MBEDTLS_SHA3_WANT_SHAKE128
+#endif
+#if defined(MBEDTLS_PSA_BUILTIN_ALG_SHAKE256)
+#define MBEDTLS_SHA3_WANT_SHAKE256
+#endif
```
valeriosetti, line 58: IIUC these `MBEDTSL_SHA3_WANT_SHAKExxx` are sort of temporary and due to the missing `PSA_WANT_` counterpart, right? If so, can't we add a comment above to explain this situation. I know you already added a comment in `crypto_config.h`, but since this is adding a different mechanism compared to what we're used to (`PSA_WANT_xxx` <-> `MBEDTLS_PSA_BUILTIN_xxx) I think it's nice to document this so whoever read this has some clue on what's happening.

valeriosetti (changes requested):

drivers/builtin/src/sha3.c
gilles-peskine-arm: What's wrong with relying on `ctx->olen`? It's set in `mbedtls_sha3_starts`. The code also relies on `ctx->max_block_size` which is set at the same time. (Except on context reuse from SHA3 to SHAKE, as Ben noted, which I'll fix and test.)

drivers/builtin/include/mbedtls/private/crypto_adjust_config_tweak_builtins.h
gilles-peskine-arm: No, the SHA3-specific macros are here to stay. It's not fully apparent why yet: the weird thing is that there are hashes that use SHAKE. (SHAKE128/256 used with HashML-DSA, SHAKE256/512 used with Ed448ph, and more used with HashSLH-DSA.) So consider a build which is solely intended for Ed448ph: it will not enable SHAKE256 in `psa_crypto_xof.c` (but it will enable `SHAKE256_512` in `psa_crypto_hash.c`), but it does need SHAKE256 in `sha3.c`. In this build, `MBEDTLS_PSA_BUILTIN_ALG_SHAKE256` is disabled but `MBEDTLS_SHA3_WANT_SHAKE256` is enabled.

I'll improve the comments here.

drivers/builtin/src/sha3.c
valeriosetti: Using a dedicated field to hold the `sha3_id` clarifies the type of sha3 context you're dealing with. For example adding 
a `mbedtls_sha3_id MBEDTLS_PRIVATE(id);` line to `mbedtls_sha3_context` and creating an internal helper like:
```
int mbedtls_sha3_is_shake(mbedtls_sha3_context *ctx)
{
    return (ctx->id == MBEDTLS_SHA3_SHAKE128) || (ctx->id == MBEDTLS_SHA3_SHAKE256));
}
```
would make the above line:
```
if (mbedtls_sha3_is_shake(ctx)) {
```
But I guess it's a matter of preference.

drivers/builtin/src/sha3.c
gilles-peskine-arm: More precisely, the case where it can go wrong is if there's a non-SHAKE operation **without finish**, then a SHAKE operation. The SHAKE operation will be limited to a single finish() call, which will output exactly the length of the hash from the first operation. We aren't currently testing a context reuse without finish (i.e. init, starts, update, starts, update, finish).

drivers/builtin/src/sha3.c
bjwtaylor, line 327: Should this function check for the presence of ctx->finished to prevent an API misuse of using an update after a finish and produce an error if it is set?

drivers/builtin/src/sha3.c
```diff
@@ -711,6 +743,133 @@ static int mbedtls_sha3_long_kat_test(int verbose,
     return result;
 }
 
+#endif /* defined(MBEDTLS_PSA_BUILTIN_ALG_SHA3_SOME_HASH) */
+
+#if defined(MBEDTLS_SHA3_WANT_SHAKE128)
+static const unsigned char shake128_test_input[2][16] =
```
bjwtaylor, line 748: Do we document anywhere where this test data comes from?

drivers/builtin/src/sha3.c
gilles-peskine-arm: Ideally yes, but this is now a private interface.

It would be good to do that in 3.6 though, since `sha3.h` is public there.

drivers/builtin/src/sha3.c
gilles-peskine-arm: _Should_, yes. _Do_, unfortunately, no. But we don't actually rely on it for correctness (we have the unit tests for that), so I can live with it.

bjwtaylor: Yeah, I was more thinking if anybody has to modify it or wants to use it for other tests it's useful to know where it comes from.

drivers/builtin/src/sha3.c
gilles-peskine-arm: As discussed on Thursday, this is an internal interface since 1.0 so we shouldn't spend much energy on making it robust against misuse. `ctx->olen` is only tracked to protect against misuse in `finished()` when the caller requests a length that is different from the length of the selected hash, and it's not relevant at all to SHAKE, so there's no point in making it more robust. So I don't think anything needs to change here.

(We could actually remove `olen` and let the caller beware. I'd probably do that if I was making this module from scratch, but I think simplifying non-SHAKE-related behavior is out of scope here.)

drivers/builtin/src/sha3.c
gilles-peskine-arm: As discussed on Thursday, this is an internal interface, we don't particularly need to protect against misuse.

drivers/builtin/include/mbedtls/private/crypto_adjust_config_tweak_builtins.h
```diff
@@ -38,6 +38,18 @@
...
 #endif
 
+/* If a SHAKE variant is enabled in psa_crypto_xof.c, tell sha3.c that we
+ * want it.
+ *
+ * Note that the PSA API (following NIST standards) defines hash algorithms
+ * that are SHAKE128 or SHAKE256 with a specific output lengths. From the
+ * perspective of sha3.c, these are just users of SHAKE128/SHAKE256, but
+ * from the perspective of psa_crypto_hash.c and psa_crypto_xof.c,
+ * thy are hashes and not XOF. So, for example, if the SHAKE256/512 hash
```
valeriosetti: I think there's a typo at the beginning of the line: *thEy

valeriosetti (dismissed):
I found only a minor typo in a comment, but the rest LGTM.

bjwtaylor (dismissed):

valeriosetti:
@gilles-peskine-arm there seems to be a conflict on this PR. If you can please fix it then we can re-approve and merge ;)

drivers/builtin/include/mbedtls/private/crypto_adjust_config_tweak_builtins.h
gilles-peskine-arm: I amended the commit.

gilles-peskine-arm:
I rewrote the history to remove the framework update commit, since it isn't needed after all. I also fixed the typo that Valerio pointed out while I was at it. The previous version is at https://github.com/gilles-peskine-arm/TF-PSA-Crypto/tree/sha3-shake-builtin-4

valeriosetti (approved):

bjwtaylor (approved):

PR #661, opened 2026-01-15
Use MPI functions to fill RSA contexts in tests for RSA export

### LINKED ISSUE #576: Use MPI functions to fill RSA contexts in tests for RSA export
### Suggested enhancement

In tests like `mbedtls_rsa_export()`, `mbedtls_rsa_export_raw()` and similar (from `test_suite_rsa.function`) `mbedtls_rsa_import()` is used to parse the RSA key. This dates back to when `MBEDTLS_RSA_ALT` was a thing and fields of `rsa_context` structure were ideally unknown. But now that we've removed that option, we know we're using our own definition for the structure, so we can access the fields directly.
The goal of this task is to replace `mbedtls_rsa_import()` with MPI functions to directly fill the fields of the RSA structure.

This issue was suggested [here](https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/559#discussion_r2536859124).

### Justification

TF-PSA-Crypto needs this because these tests are just for the export functionality in RSA so we can (slightly) "simplify" requirements.

valeriosetti:
## Description

Resolves #576
After this PR there is no remaining occurrence of `mbedtls_rsa_import` or `mbedtls_rsa_export`.

## PR checklist

- [x] **changelog** not required because: all removed functions were internal
- **tests**  provided


drivers/builtin/src/rsa.c
```diff
@@ -231,124 +223,113 @@ int mbedtls_rsa_parse_pubkey(mbedtls_rsa_context *rsa, const unsigned char *key,
...
-        (ret = mbedtls_asn1_write_mpi(p, start, &T)) < 0) {
-        goto end_of_export;
+    /* Export QP, DQ, DP */
+#if !defined(MBEDTLS_RSA_NO_CRT)
+    MBEDTLS_RSA_WRITE_MPI(&rsa->QP);
+    MBEDTLS_RSA_WRITE_MPI(&rsa->DQ);
+    MBEDTLS_RSA_WRITE_MPI(&rsa->DP);
+#else /* MBEDTLS_RSA_NO_CRT */
+    if ((ret = mbedtls_rsa_deduce_crt(&rsa->P, &rsa->Q, &rsa->D, &DP, &DQ, &QP)) != 0) {
+        return MBEDTLS_ERROR_ADD(MBEDTLS_ERR_RSA_BAD_INPUT_DATA, ret);
```
bjwtaylor: Do you need to go through the cleanup path to free DP, DQ and DP for this return and others?

drivers/builtin/src/rsa.c
```diff
@@ -231,124 +223,113 @@ int mbedtls_rsa_parse_pubkey(mbedtls_rsa_context *rsa, const unsigned char *key,
...
 
+#define MBEDTLS_RSA_WRITE_MPI(m)                               \
+    do {                                                        \
+        if ((ret = mbedtls_asn1_write_mpi(p, start, m)) < 0) {  \
+            goto end_of_export;                                 \
+        }                                                       \
+        len += ret;                                             \
+    } while (0)
+
 int mbedtls_rsa_write_key(const mbedtls_rsa_context *rsa, unsigned char *start,
```
bjwtaylor, line 291: Does this function need to validate that this is a private key prior to serialising private fields, to prevent a public key being serialised into a invalid private key?

drivers/builtin/src/rsa.c
```diff
@@ -231,124 +223,113 @@ int mbedtls_rsa_parse_pubkey(mbedtls_rsa_context *rsa, const unsigned char *key,
      *      publicExponent    INTEGER   -- e
      *  }
      */
+    mbedtls_mpi_init(&rsa->N);
```
bjwtaylor, line 226: Should this function perform a free prior to init, to prevent leaking of context?

tests/suites/test_suite_rsa.function
```diff
@@ -10,31 +10,55 @@
 
 static int mbedtls_rsa_test_fill_context(mbedtls_rsa_context *rsa,
                                          data_t *P_data, data_t *Q_data,
-                                         data_t *N_data, data_t *E_data)
+                                         data_t *N_data, data_t *E_data,
+                                         data_t *D_data)
 {
-    mbedtls_mpi N; mbedtls_mpi P; mbedtls_mpi Q;
-    mbedtls_mpi E; mbedtls_mpi D;
     int ret;
```
bjwtaylor: Existing issue, but should ret be initialised prior to use?

drivers/builtin/src/rsa.c
valeriosetti: Yes, nice catch, thanks!

drivers/builtin/src/rsa.c
valeriosetti: Adding a free before this initialization means that we allow for this function to be called in whatever scenario and if this were a public function we should take this possibility into account. However nowadays this function is only called from `mbedtls_psa_rsa_load_representation` which [always initialize](https://github.com/Mbed-TLS/TF-PSA-Crypto/blob/81436417ad175af6d4480311ab9e32825739502a/drivers/builtin/src/psa_crypto_rsa.c#L64) the RSA context before calling `mbedtls_rsa_parse_pubkey` so I think there is no problem in taking this shortcut.

drivers/builtin/src/rsa.c
valeriosetti: As for `mbedtls_rsa_parse_pubkey`, this function is private now and only called [here](https://github.com/Mbed-TLS/TF-PSA-Crypto/blob/81436417ad175af6d4480311ab9e32825739502a/drivers/builtin/src/psa_crypto_rsa.c#L164) where a check on the key type (public VS key-pair) is done before the call. I don't think relaxing checks is too bad in this case, wdyt?

tests/suites/test_suite_rsa.function
valeriosetti: I don't think there is any risk to return before `ret` being assigned in this function. `mbedtls_mpi_init` cannot fail and `MBEDTLS_MPI_CHK` sets `ret` internally.
Or better: the only case in which this can happen is if all the `X_data` input parameters are `NULL`. Not a really useful case, but I agree it's better to handle this.

drivers/builtin/src/rsa.c
```diff
@@ -231,124 +223,114 @@ int mbedtls_rsa_parse_pubkey(mbedtls_rsa_context *rsa, const unsigned char *key,
...
-    if ((ret = mbedtls_rsa_export(rsa, NULL, NULL, NULL, NULL, &T)) != 0 ||
-        (ret = mbedtls_asn1_write_mpi(p, start, &T)) < 0) {
+    /* Export QP, DQ, DP */
+#if !defined(MBEDTLS_RSA_NO_CRT)
+    MBEDTLS_RSA_WRITE_MPI(&rsa->QP);
+    MBEDTLS_RSA_WRITE_MPI(&rsa->DQ);
+    MBEDTLS_RSA_WRITE_MPI(&rsa->DP);
+#else /* MBEDTLS_RSA_NO_CRT */
+    if ((ret = mbedtls_rsa_deduce_crt(&rsa->P, &rsa->Q, &rsa->D, &DP, &DQ, &QP)) != 0) {
+        ret = MBEDTLS_ERROR_ADD(MBEDTLS_ERR_RSA_BAD_INPUT_DATA, ret);
```
gilles-peskine-arm: We don't add error codes any longer, so we should avoid `MBEDTLS_ERROR_ADD` in new code. Now it just returns `ret` (it would return `MBEDTLS_ERR_RSA_BAD_INPUT_DATA` if `ret` was 0 but that can't happen here).

```suggestion
```

drivers/builtin/src/rsa.c
gilles-peskine-arm: I'm fine with not checking, but please update the documentation in `rsa_internal.h`: it still documents an error if the key isn't a private key.

drivers/builtin/src/rsa.c
gilles-peskine-arm: Strictly speaking, calling `mbedtls_mpi_init` is not actually necessary, since the only usage site calls `mbedtls_rsa_init` immediately before this function, so `rsa->N` and `rsa-E` are all-bits zero at this point. All-bits-zero is not a valid MPI representation (the sign bit must be 1 or -1, and `mbedtls_mpi_init` sets it to 1), but it is good enough for `mbedtls_mpi_read_binary`.

However, I'm fine with keeping the `mbedtls_mpi_init` calls for robustsness. If we change the way `mbedtls_rsa_parse_pubkey` is used and it receives an uninitialized context, it's better to wipe it than call `free` on something that is not a heap pointer that we intended to free.

tests/suites/test_suite_rsa.function
```diff
@@ -10,31 +10,55 @@
...
+    if (D_data != NULL) {
+        MBEDTLS_MPI_CHK(mbedtls_mpi_read_binary(&rsa->D, D_data->x, D_data->len));
+    }
 
-    MBEDTLS_MPI_CHK(mbedtls_rsa_import(rsa, &N, &P, &Q, &D, &E));
+    if (mbedtls_mpi_size(&rsa->D) == 0) {
+        if ((mbedtls_mpi_size(&rsa->P) != 0) &&
+            (mbedtls_mpi_size(&rsa->Q) != 0) &&
+            (mbedtls_mpi_size(&rsa->E) != 0)) {
+            MBEDTLS_MPI_CHK(mbedtls_rsa_deduce_private_exponent(&rsa->P, &rsa->Q, &rsa->E,
```
gilles-peskine-arm, line 42: Why are we still using `mbedtls_rsa_deduce_private_exponent`? The API since TF-PSA-Crypto 1.0 has no way to construct an RSA key without providing the private exponent.

tests/suites/test_suite_rsa.function
```diff
@@ -10,31 +10,55 @@
...
+                                               &rsa->QP));
+    }
 #endif /* !MBEDTLS_RSA_NO_CRT */
 
 cleanup:
-    mbedtls_mpi_free(&N); mbedtls_mpi_free(&P); mbedtls_mpi_free(&Q);
-    mbedtls_mpi_free(&E); mbedtls_mpi_free(&D);
+    if (ret != 0) {
+        mbedtls_mpi_free(&rsa->N); mbedtls_mpi_free(&rsa->P); mbedtls_mpi_free(&rsa->Q);
+        mbedtls_mpi_free(&rsa->E); mbedtls_mpi_free(&rsa->D);
```
gilles-peskine-arm: Minor: can't we just do `mbedtls_rsa_free(N)`?

tests/suites/test_suite_rsa.function
```diff
@@ -10,31 +10,55 @@
 
 static int mbedtls_rsa_test_fill_context(mbedtls_rsa_context *rsa,
                                          data_t *P_data, data_t *Q_data,
-                                         data_t *N_data, data_t *E_data)
+                                         data_t *N_data, data_t *E_data,
+                                         data_t *D_data)
```
gilles-peskine-arm: Minor: it's weird to have parameters in the order primes, then public, then private exponent). Usually the order of parameters in an RSA key is public, then private exponent, then primes, then CRT. Or occasionally the private exponent is at the end. But the mixture of private and public makes it harder to follow.

gilles-peskine-arm (changes requested):
Mostly LGTM but please update the documentation in `rsa_internal.h`

tests/suites/test_suite_rsa.function
valeriosetti: There are test functions like `mbedtls_rsa_pkcs1_sign` where the D parameter is not provided. IMO it's easier to deduce it rather than update all test data. Wdyt?

gilles-peskine-arm: It's no longer an API function. I thought we could remove it, but it's actually the same function that's used while generating a key (which makes sense), so it won't go away from the library. Still, we should exclude it from builds that have RSA signature/decryption support but not key generation, and that would make it unavailable in tests in such a configuration. And, also, I think it makes sense to have all the expected key data in `.data` files: that makes it easier to cross-check, or to calculate the expected intermediate values externally when debugging a test failure. So I think we should complete the test data. But it makes sense that it's out of scope here. https://github.com/Mbed-TLS/TF-PSA-Crypto/issues/669

gilles-peskine-arm (approved):
LGTM

bjwtaylor (approved):

Issue #655, opened 2026-01-13
Fix mldsa-native integration on MSVC


gilles-peskine-arm:
In https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/652 as of 98a18bf5738a66677670f6528fa24285f3b77c1b (mldsa-native upstream f504c9a2c09842cdc0c03b31b1c90afcd991540f), the MSVC build fails ([example](https://ci.trustedfirmware.org/job/mbed-tls-tf-psa-crypto-multibranch/job/PR-652-head/1/pipeline-console/log?nodeId=4593)). Relevant log snippet:
```
ClCompile:
  wrap_mldsa_native.c
c:\users\administrator\appdata\local\temp\tmp6x8o2e4y\tf-psa-crypto\drivers\pqcp\mldsa-native\mldsa\src\ct.h(150): error C2143: syntax error: missing ')' before '(' [C:\Users\ADMINI~1\AppData\Local\Temp\tmp6x8o2e4y\cmake_solution\tf-psa-crypto\drivers\pqcp\pqcp.vcxproj]
c:\users\administrator\appdata\local\temp\tmp6x8o2e4y\tf-psa-crypto\drivers\pqcp\mldsa-native\mldsa\src\ct.h(150): error C2059: syntax error: ')' [C:\Users\ADMINI~1\AppData\Local\Temp\tmp6x8o2e4y\cmake_solution\tf-psa-crypto\drivers\pqcp\pqcp.vcxproj]
c:\users\administrator\appdata\local\temp\tmp6x8o2e4y\tf-psa-crypto\drivers\pqcp\mldsa-native\mldsa\src\ct.h(150): error C2146: syntax error: missing ')' before identifier 'int32_t' [C:\Users\ADMINI~1\AppData\Local\Temp\tmp6x8o2e4y\cmake_solution\tf-psa-crypto\drivers\pqcp\pqcp.vcxproj]
c:\users\administrator\appdata\local\temp\tmp6x8o2e4y\tf-psa-crypto\drivers\pqcp\mldsa-native\mldsa\src\ct.h(150): error C2091: function returns function [C:\Users\ADMINI~1\AppData\Local\Temp\tmp6x8o2e4y\cmake_solution\tf-psa-crypto\drivers\pqcp\pqcp.vcxproj]
[...]

PR #666, opened 2026-01-19
Stop defining inline in our headers

### LINKED ISSUE #662: Stop defining inline in our headers
We [define `inline` as a macro on ArmCC and MSVC](https://github.com/Mbed-TLS/TF-PSA-Crypto/blob/tf-psa-crypto-1.0.0/include/tf-psa-crypto/build_info.h#L98).

We no longer officially support ArmCC 5, which is a C90 compiler (although it has enough C99 features that it still works for most of our code — I think the main exception is some Arm-specific optimizations which don't compile anymore). We no longer support older versions of MSVC where `inline` wasn't a keyword (we dropped Visual Studio 2010 when we made C99 an official requirement, and in 3.6 we document VS 2017 as the minimum version; my reading of [MSVC documentation](https://learn.microsoft.com/en-us/cpp/cpp/inline-functions-cpp?view=msvc-140) is that `inline` is a keyword at least as far back as VS 2015). So these definitions of `inline` should not be needed any longer.

It's a bit rude to define this in a header that application code will include, even if we only do it on specific compilers where it should be ok. It's true that we've been doing it for a very long time without complaints (in `build_info.h` since Mbed TLS 3.0, and in some public headers since PolarSSL 0.99+). But it has caused a problem recently: [it broke the build of mldsa-native](https://github.com/Mbed-TLS/TF-PSA-Crypto/issues/655#issuecomment-3752488047). So I think we should remove it. If someone really wants to make the library work on a compiler that doesn't have a full C99 including the `inline` keyword, they can define the macro themselves.

This is a trivial change in terms of code difference, but not in terms of implications, so I don't want to do it in an LTS branch.

valeriosetti:
## Description

Resolves #662

## PR checklist

- [ ] **changelog** provided
- **tests**  not required because: no code change

### DIFF
--- a/ChangeLog.d/issue662.txt
+++ b/ChangeLog.d/issue662.txt
@@ -1,0 +1,5 @@
+Removals
+   * Definition of macro 'inline' is removed from 'build_info.h'. This was
+     used on some old non-C99-compliant compilers (ArmCC 5 and old MSVC) for
+     which support for building the library has already been dropped in the
+     past.
--- a/ChangeLog.d/issue662.txt
+++ b/ChangeLog.d/issue662.txt
@@ -1,5 +1,4 @@
 Removals
-   * Definition of macro 'inline' is removed from 'build_info.h'. This was
-     used on some old non-C99-compliant compilers (ArmCC 5 and old MSVC) for
-     which support for building the library has already been dropped in the
-     past.
+   * The headers no longer define 'inline' as a macro. This was done on Arm
+     Compiler 5 and MSVC. The compiler versions that needed this definition are
+     no longer supported since TF-PSA-Crypto 1.0.
--- a/include/tf-psa-crypto/build_info.h
+++ b/include/tf-psa-crypto/build_info.h
@@ -100,12 +100,6 @@
 #define _CRT_SECURE_NO_DEPRECATE 1
 #endif
 
-/* Define `inline` on some non-C99-compliant compilers. */
-#if (defined(__ARMCC_VERSION) || defined(_MSC_VER)) && \
-    !defined(inline) && !defined(__cplusplus)
-#define inline __inline
-#endif
-
 #if defined(TF_PSA_CRYPTO_CONFIG_FILES_READ)
 #error \
     "Something went wrong: TF_PSA_CRYPTO_CONFIG_FILES_READ defined before reading the config files!"


gilles-peskine-arm (dismissed):
LGTM

Although, should we have a changelog entry, since this removes some practical portability? This doesn't remove theoretical portability since it only affects compilers that we officially don't support (ArmCC 5, old MSVC).

valeriosetti:
> Although, should we have a changelog entry, since this removes some practical portability? This doesn't remove theoretical portability since it only affects compilers that we officially don't support (ArmCC 5, old MSVC).

Yeah, indeed this was the question I raised in the description of the PR. TBH I don't know: if you or the 2nd reviewer think it's better to have a ChangeLog, I will add it :)

hoffbrinkle:
I would recommend one. It's a macro in a header file so it could have some "spooky action at distance." If someone is trying to figure out why their build is behaving differently, this could valuable to figure that out.

valeriosetti:
> I would recommend one. It's a macro in a header file so it could have some "spooky action at distance." If someone is trying to figure out why their build is behaving differently, this could valuable to figure that out.

Done ;)

bjwtaylor (dismissed):

ChangeLog.d/issue662.txt
```diff
@@ -0,0 +1,5 @@
+Removals
+   * Definition of macro 'inline' is removed from 'build_info.h'. This was
+     used on some old non-C99-compliant compilers (ArmCC 5 and old MSVC) for
+     which support for building the library has already been dropped in the
+     past.
```
gilles-peskine-arm: Suggested minor English improvements:

> The headers no longer define 'inline' as a macro. This was done on Arm Compiler 5 and MSVC. The compiler versions that needed this definition are no longer supported since TF-PSA-Crypto 1.0.

gilles-peskine-arm:
> It's a macro in a header file so it could have some "spooky action at distance."

Indeed, such spooky action when including both our headers and another library is [why defining the macro was a problem](https://github.com/Mbed-TLS/TF-PSA-Crypto/issues/655#issuecomment-3752488047). I can imagine weirdness in the other direction as well.

gilles-peskine-arm (approved):
LGTM

bjwtaylor (approved):

PR #668, opened 2026-01-20
Software GCM table calculation buggy with gcc -O3

### LINKED ISSUE #665: Software GCM table calculation buggy with gcc -O3
Our pure software implementation of GCM is broken when compiled with `gcc -O3` in Mbed TLS 3.6.5 and TF-PSA-Crypto 1.0.0. All positive unit tests involving GCM with a non-empty message fail.

(This was also mentioned in https://github.com/Mbed-TLS/mbedtls/issues/9743#issuecomment-2449063776, but the compiler warnings in the original issue are unrelated.)

### Reproduction

Steps to reproduce:

* Compile with **`gcc -O3`**, using GCC 10 through 14.2. I haven't been able to reproduce the bug at `-O2`, with older/newer GCC (GCC 9, GCC 14.3, GCC 15.1, GCC 15.2), or with Clang. I don't know exactly which (combination of) optimization(s) trigger the bug, but I'm pretty sure that it only happens with sufficient loop unrolling.

* Arrange for the buggy code path to be invoked. Only **little-endian** architectures are buggy. 32-bit vs 64-bit makes no difference. Any of the following triggers the bug:

    * A little-endian architecture other than x86 and arm, in the default configuration.

    * https://github.com/gilles-peskine-arm/mbedtls/tree/gcm-little-endian-instrument-3.6, with AESNI and AESCE disabled, built with `-O3 -DMBEDTLS_ONLY_GENERIC_CODE`. This forces the same code path as with non-x86/arm architectures.

            mbedtls-prepare-build -p opt -d build-generic-gcc13-opt --config-unset=MBEDTLS_AESCE_C,MBEDTLS_AESNI_C --cc=gcc-13 --cflags='-O3 -DMBEDTLS_ONLY_GENERIC_CODE'

    * https://github.com/gilles-peskine-arm/mbedtls/tree/gcm-little-endian-instrument-3.6, with AESNI and AESCE disabled, built with `-O3 -DGCM_BREAK_PUT`. This pinpoints the code change to just the one buggy calculation.

        Note that I had to patch two places to use the potentially-buggy PUT code. If I only patch one of them, the calculation is correct. If I patch more, the calculation has the same buggy result.

            mbedtls-prepare-build -d build-breakgcm-gcc13-opt -p opt --cc=gcc-13 --cflags='-O3 -DGCM_BREAK_PUT' --config-unset=MBEDTLS_AESNI_C,MBEDTLS_AESCE_C

### Analysis

I haven't traced back when the bug started, but I think it was between 3.5.0 and 3.6.0: the likely culprit is either https://github.com/Mbed-TLS/mbedtls/pull/8719 or https://github.com/Mbed-TLS/mbedtls/pull/8788.

The bug is specifically in `gcm_gen_table()`, when the following code path is used:
* GCM is done in software (no AESNI or AESCE).
* The platform is little-endian.
* The optimization path in [`alignment.h`](https://github.com/Mbed-TLS/mbedtls/blob/mbedtls-3.6.0/library/alignment.h) has `MBEDTLS_EFFICIENT_UNALIGNED_ACCESS` disabled, and `UINT_UNALIGNED_STRUCT` enabled.

In https://github.com/gilles-peskine-arm/mbedtls/tree/gcm-little-endian-instrument-3.6, I added code to print the resulting table. Here's the correct table for the first test case of `test_suite_gcm.aes128_en.run`:

```
0000000000000000 0000000000000000
15febd98009bc186 554551b6d7c95112
2bfd7b300137830c aa8aa36daf92a224
3e03c6a801ac428a ffcff2db785bf336
57faf660026f0619 551546db5f254448
42044bf802f4c79f 0050176d88ec155a
7c078d5003588515 ff9fe5b6f0b7e66c
69f930c803c34493 aadab400277eb77e
aff5ecc004de0c32 aa2a8db6be4a8890
ba0b51580445cdb4 ff6fdc006983d982
840897f005e98f3e 00a02edb11d82ab4
91f62a6805724eb8 55e57f6dc6117ba6
f80f1aa006b10a2b ff3fcb6de16fccd8
edf1a738062acbad aa7a9adb36a69dca
d3f2619007868927 55b568004efd6efc
c60cdc08071d48a1 00f039b699343fee
```

And here's the table in the buggy case (I got the same bad values for all the buggy cases that I've tried):

```
0000000000000000 0000000000000000
1500000000000000 554551b6d7c95112
2bfd7b300137830c aa8aa36daf92a224
3efd7b300137830c ffcff2db785bf336
57faf660026f0619 551546db5f254448
42faf660026f0619 0050176d88ec155a
7c078d5003588515 ff9fe5b6f0b7e66c
69078d5003588515 aadab400277eb77e
aff5ecc004de0c32 aa2a8db6be4a8890
baf5ecc004de0c32 ff6fdc006983d982
840897f005e98f3e 00a02edb11d82ab4
910897f005e98f3e 55e57f6dc6117ba6
f80f1aa006b10a2b ff3fcb6de16fccd8
ed0f1aa006b10a2b aa7a9adb36a69dca
d3f2619007868927 55b568004efd6efc
c6f2619007868927 00f039b699343fee
```

So what happens is that every other row has 9 correct bytes and 7 bytes which duplicate the previous row.

### Whose fault is it?

This code isn't standard C, but it says so on the tin: it uses `__attribute__((packed))` to cheat with alignment constraints. It's not clear to me where, if anywhere, our code misuses GCC. So it could be a compiler bug. But our code is rather twisty and dodgy, so it doesn't inspire a lot of confidence in me.

The same code breaks on Ubuntu 24.04 with GCC 14.2 but works on Ubuntu 25.10 with GCC 14.3 (x86_64 in both cases). So we might have hit one of the [219 bugs fixed in GCC 14.3](https://gcc.gnu.org/bugzilla/buglist.cgi?bug_status=RESOLVED&resolution=FIXED&target_milestone=14.3).

### Goal of this issue

* Create an `all.sh` job that reproduces the problem: build with `gcc -O3` using a problematic version of GCC, with AESNI/AESCE disabled, and arranging to trigger the problematic code path in `alignment.h`. (This may require adding a **private** config macro.)
* Fix the bug in GCM, trying to minimize disruption in `alignment.h`. (Cleaning up `alignment.h` is a bigger task that is out of scope here.) We consider the bug to be fixed if the non-regression test passes. Note that in `gcm.c`, the data is more aligned than the worse case of `alignment.h`, so we can probably get away with using simpler ways to copy/xor that are as efficient.
* We don't have big-endian testing on the CI, so make sure to test locally on a big-endian platform (e.g. PPC).

valeriosetti:
## Description

Resolves #665.

The problem relies in `gcm_gen_table_rightshift` (in `gcm.c`) and it seems to be related to memory aliasing when structure `mbedtls_uint64_unaligned_t` is used and optimization level is `-O3`. 

Without the `may_alias` attribute for `mbedtls_uint64_unaligned_t` structure the compiler will optimize code and interleave `MBEDTLS_PUT_UINT64_BE(MBEDTLS_GET_UINT64_BE(...))` operations with other bytes oriented ones (ex: `u8Dst[8] |= (u8Src[7] & 0x01) << 7;`).

If you are curious here's a comparison between 2 generated `gcm.o` object files with and without `may_alias` added as attribute. 
- without--> [gcm.fail.txt](https://github.com/user-attachments/files/24741072/gcm.fail.txt)
- with --> [gcm.ok.txt](https://github.com/user-attachments/files/24741076/gcm.ok.txt)

Comparing them with Meld (or any other text-diff editor of your choice) you'll see that the only differences between the 2 files are in the order of operations in the `gcm_gen_table_rightshift()` implementation.

For a full explanation of the problem and the relative solution I asked Claude Code to generate it since it looks quite nice I'm posting it here --> [explanation.txt](https://github.com/user-attachments/files/24740968/explanation.txt). I don't think I can write anything better/clearer that that.

## PR checklist

- [x] **changelog** provided
- [x] **mbedtls 3.6 PR** provided: https://github.com/Mbed-TLS/mbedtls/pull/10581
- **tests**  provided


valeriosetti:
@LocutusOfBorg since I don't have `ppc6el` architecture at home, can you please give this patch a try when you have some spare time and let me know if it works fine?

LocutusOfBorg:
Hello, I tried, this is the result:
https://launchpadlibrarian.net/844046519/buildlog_ubuntu-resolute-ppc64el.mbedtls_3.6.5-0.1ubuntu2_BUILDING.txt.gz

cd /<<PKGBUILDDIR>>/obj-powerpc64le-linux-gnu/library && /usr/bin/cc -Dmbedcrypto_EXPORTS -I/<<PKGBUILDDIR>>/include -I/<<PKGBUILDDIR>>/library -I/<<PKGBUILDDIR>>/obj-powerpc64le-linux-gnu/library -g -O3 -Werror=implicit-function-declaration -Werror=array-bounds -Werror=clobbered -Werror=volatile-register-var -D__DEB_CANARY_CFLAGS_8d09e4b85c783cbc30c9b8ae175f2d33__ -fno-omit-frame-pointer -ffile-prefix-map=/<<PKGBUILDDIR>>=. -fstack-protector-strong -Wformat -Werror=format-security -fno-stack-clash-protection -fdebug-prefix-map=/<<PKGBUILDDIR>>=/usr/src/mbedtls-3.6.5-0.1ubuntu2 -D__DEB_CANARY_CPPFLAGS_8d09e4b85c783cbc30c9b8ae175f2d33__ -Wdate-time -D_FORTIFY_SOURCE=3 -Wall -Wextra -Wwrite-strings -Wmissing-prototypes -Wformat=2 -Wno-format-nonliteral -Wvla -Wlogical-op -Wshadow -Wformat-signedness -Wformat-overflow=2 -Wformat-truncation -Werror -Wmissing-declarations -std=c99 -fPIC -MD -MT library/CMakeFiles/mbedcrypto.dir/chacha20.c.o -MF CMakeFiles/mbedcrypto.dir/chacha20.c.o.d -o CMakeFiles/mbedcrypto.dir/chacha20.c.o -c /<<PKGBUILDDIR>>/library/chacha20.c
In file included from /<<PKGBUILDDIR>>/library/cmac.c:29:
In function ‘mbedtls_xor’,
    inlined from ‘mbedtls_cipher_cmac_finish’ at /<<PKGBUILDDIR>>/library/cmac.c:305:5:
/<<PKGBUILDDIR>>/library/common.h:235:14: error: writing 1 byte into a region of size 0 [-Werror=stringop-overflow=]
  235 |         r[i] = a[i] ^ b[i];
      |         ~~~~~^~~~~~~~~~~~~
In file included from /<<PKGBUILDDIR>>/include/mbedtls/cmac.h:17,
                 from /<<PKGBUILDDIR>>/library/cmac.c:33:
/<<PKGBUILDDIR>>/include/mbedtls/cmac.h: In function ‘mbedtls_cipher_cmac_finish’:
/<<PKGBUILDDIR>>/include/mbedtls/cmac.h:61:41: note: at offset 16 into destination object ‘state’ of size 16
   61 |     unsigned char       MBEDTLS_PRIVATE(state)[MBEDTLS_CMAC_MAX_BLOCK_SIZE];
      |                                         ^~~~~
/<<PKGBUILDDIR>>/include/mbedtls/private_access.h:17:33: note: in definition of macro ‘MBEDTLS_PRIVATE’
   17 | #define MBEDTLS_PRIVATE(member) member
      |                                 ^~~~~~
[  9%] Building C object library/CMakeFiles/mbedcrypto.dir/chachapoly.c.o

And this is the applied patch
https://launchpadlibrarian.net/844046227/mbedtls_3.6.5-0.1ubuntu1_3.6.5-0.1ubuntu2.diff.gz

gilles-peskine-arm:
@LocutusOfBorg Can you please try with `-Wno-stringop-overflow`? We know that GCC tends to emit this warning on our code, and that it's spurious. Fixing the actual bug in GCM is much higher priority for us.

LocutusOfBorg:
Build with additional flags:
https://launchpadlibrarian.net/844056957/mbedtls_3.6.5-0.1ubuntu2_3.6.5-0.1ubuntu3.diff.gz

Build log: https://launchpad.net/~costamagnagianfranco/+archive/ubuntu/costamagnagianfranco-ppa/+build/32161736

Build status: SUCCESSFUL!

LocutusOfBorg:
And uploaded to Ubuntu too https://launchpad.net/ubuntu/+source/mbedtls/3.6.5-0.1ubuntu2

tests/scripts/components-compiler.sh
```diff
@@ -129,3 +129,46 @@ component_test_tf_psa_crypto_zeroize () {
...
+    make test
+}
+
+support_test_tf_psa_crypto_optimized_alignment() {
+    ver="$(gcc --version | grep -oP '\d+\.\d+\.\d+' | head -n1)"
+    ver_major="$(echo $ver | cut -d'.' -f1)"
+    ver_minor="$(echo $ver | cut -d'.' -f2)"
+    ver_patch="$(echo $ver | cut -d'.' -f3)"
+
+    [ "$ver_major" -ge 10 ] && [ "$ver_minor" -le 14 ]
```
gilles-peskine-arm: We need 10.0 ≤ version ≤ 14.2 (I couldn't reproduce the problem with 14.3). We could pick 10–13

```suggestion
    [ "$ver_major" -ge 10 ] && [ "$ver_major" -le 13 ]
```
But I think the only suitable version we currently have in our Docker images is GCC 14.2 on Ubuntu 24.04.

```suggestion
    [[ ($ver_major -ge 10 && $ver_major -le 13) || ($ver_major == 14 && $ver_minor -le 2) ]]
```

tests/scripts/components-compiler.sh
```diff
@@ -129,3 +129,46 @@ component_test_tf_psa_crypto_zeroize () {
         done
     done
 }
+
+# This originated from an issue (crypto#665) found in GCM when the library is
```
gilles-peskine-arm: It would be friendlier to give the complete URL https://github.com/Mbed-TLS/TF-PSA-Crypto/issues/665

tests/scripts/components-compiler.sh
```diff
@@ -129,3 +129,46 @@ component_test_tf_psa_crypto_zeroize () {
...
+# built with GCC "10 <= version <= 14" on platforms other than x86 and ARM64.
```
gilles-peskine-arm:
```suggestion
# built with GCC "10 <= version <= 14.2" on platforms other than x86 and ARM64.
```

core/alignment.h
```diff
@@ -15,6 +15,7 @@
 #include <string.h>
 #include <stdlib.h>
 
+#if !defined(MBEDTLS_ALIGNMENT_DISABLE_EFFICENT_UNALIGNED_ACCESS)  //no-check-names
```
gilles-peskine-arm, line 18: Please add a changelog entry.

gilles-peskine-arm (changes requested):
LGTM for the fix. The changelog is missing and the support function needs work.

I've tested locally and confirmed that at 7c4d75d9291863863b53d6fef556340c3da5150a, the code builds and the GCM tests pass with `gcc -O3` with several versions and architectures:

* GCC 10.5, 11.5, 12.4, 13.3, 14.2 on x86_64 with AESNI disabled and `-DMBEDTLS_ALIGNMENT_DISABLE_EFFICENT_UNALIGNED_ACCESS`
* GCC 4.7, GCC 9 on x86_64 with AESNI disabled and `-DMBEDTLS_ALIGNMENT_DISABLE_EFFICENT_UNALIGNED_ACCESS` (plus `-std=c99` for GCC 4.7)
* GCC 14.3, GCC 15.1 on x86_64 with AESNI disabled and `-DMBEDTLS_ALIGNMENT_DISABLE_EFFICENT_UNALIGNED_ACCESS`
* GCC 13.3 on aarch64 (AESCE disabled), arm-gnueabi, arm-gnueabihf, powerpc, powerpc64, riscv64

And I tried a couple of those at the commit before the fix (6e9028cb47e1fa3c615d1ff8856cfc9a464cce33) and can confirm that the bug was present there.

tests/scripts/components-compiler.sh
valeriosetti: I used Ubuntu 22.04 Docker image for testing:
```
$ gcc --version
gcc (Ubuntu 11.4.0-1ubuntu1~22.04.2) 11.4.0
```

gilles-peskine-arm: We recently simplified our CI and now only have 16.04, 18.04 and 24.04. (Which does leave a gap in GCC/Clang versions, but I think we'll gradually move to installing more versions in the same image.)

valeriosetti: Oh OK, I missed this. Thanks for the update ;)

core/alignment.h
valeriosetti: Done. I'll do the backport once this PR is approved so that we can have it ready in 1 shot

ChangeLog.d/issue665.txt
```diff
@@ -0,0 +1,5 @@
+Bugfix
+   * Fix instruction reordering issue in GCM tables generation found with GCC
+     compilers with version between 10.0 and 14.2 (both included) at
+     optimization level "-O3".
+     Fixes #655.
```
bjwtaylor: Shouldn't it be 665?

tests/scripts/components-compiler.sh
```diff
@@ -129,3 +129,47 @@ component_test_tf_psa_crypto_zeroize () {
...
+    cd $OUT_OF_SOURCE_DIR
+    cmake -DCMAKE_C_COMPILER=gcc -DCMAKE_C_FLAGS=" $EXTRA_C_FLAGS " "$TF_PSA_CRYPTO_ROOT_DIR"
+    make
+
+    msg "test: verify alignment with O3 optimizations in GCC"
+    make test
+}
+
+support_test_tf_psa_crypto_optimized_alignment() {
+    ver="$(gcc --version | grep -oP '\d+\.\d+\.\d+' | head -n1)"
```
bjwtaylor: Is grep -oP portable enough, can we use something like "gcc -dumpfullversion -dumpversion"

ChangeLog.d/issue665.txt
valeriosetti: :facepalm: yes, sorry

tests/scripts/components-compiler.sh
valeriosetti: Initially I looked into those parameters, but it turned out that these are not compatible across GCC 7 barrier:
- `-dumpversion` --> < GCC < 7 it returned the full version; GCC >= 7 only returns major.
- `-dumpfullversion` --> GCC < 7 didn´t exist; GCC >= 7 returns the full version.

So IMO my solution is more portable than the one using `dumpfullversion`/`dumpversion`

ChangeLog.d/issue665.txt
```diff
@@ -0,0 +1,5 @@
+Bugfix
+   * Fix instruction reordering issue in GCM tables generation found with GCC
+     compilers with version between 10.0 and 14.2 (both included) at
+     optimization level "-O3".
+     Fixes #665.
```
gilles-peskine-arm: This is user-facing, so the information content should focus on the impact of the bug, not its cause. What matters is that it causes the calculation to fail. If we want to be more specific, we can say that the tag calculation fails (which implies that the encryption is safe, so this doesn't leak data).

Also, there's one more important fact, which is that this only affected some architectures (since AESNI and AESCE bypassed the buggy code). Although, in the other direction, technically, it's possible that more compilers were affected (and even more parts of the library), we just didn't find one.

Proposed wording:

> Fix a bug that caused GCM tag calculations to fail, so that data was correctly encrypted but could not be authenticated. The bug was only observed with GCC 10.0 to 14.2 inclusive, when compiling with -O3, and running without AESNI or AESCE. Fixes #665.

tests/scripts/components-compiler.sh
gilles-peskine-arm: `grep -oP` is portable enough for the CI because we only run on a “full” Windows. But it would be nice if it worked on macOS without GNU tools (so no `grep -P`).

Anyway, since we know that versions of GCC that don't have `-dumpfullversion` are too old, we can use `-dumpfullversion`.

```
    case $(gcc -dumpfullversion 2>/dev/null) in
        ""|?.*) false;; # too old
        10.*|11.*|12.*|13.*) true;;
        14.[012].*) true;;
        *) false;; # too recent
    esac
```

gilles-peskine-arm (changes requested):
I'd like a tweak to the changelog entry. Also, it's not a blocker, but I'd prefer to simplify the support function.

gilles-peskine-arm (approved):
LGTM

tests/scripts/components-compiler.sh
```diff
@@ -129,3 +129,47 @@ component_test_tf_psa_crypto_zeroize () {
...
+
+    msg "test: verify alignment with O3 optimizations in GCC"
+    make test
+}
+
+support_test_tf_psa_crypto_optimized_alignment() {
+    case $(gcc -dumpfullversion 2>/dev/null) in
+        ""|?.*) false;; # too old
+        10.*|11.*|12.*|13.*) true;;
+        14.[012].*) true;;
```
bjwtaylor, line 172: Will this regex allow 14.1x and 14.2x and if so is this an issue? The comment suggests this should stop at 14.2, though this maybe incorrect?

tests/scripts/components-compiler.sh
```diff
@@ -129,3 +129,47 @@ component_test_tf_psa_crypto_zeroize () {
         done
     done
 }
+
+# This originated from an issue (https://github.com/Mbed-TLS/TF-PSA-Crypto/issues/665) found
+# in GCM when the library is built with GCC "10.0 <= version <= 14.2" on platforms other than
+# x86 and ARM64.
+component_test_tf_psa_crypto_optimized_alignment() {
```
bjwtaylor, line 136: It looks like it is intended that this test can run as a independent test? If so, should it run scripts/config.py full prior ro unsetting the configs to ensure a static baseline?

tests/scripts/components-compiler.sh
valeriosetti: Unless I'm mistaken `.[012].` evaluates a single number between two dots. If so this check should be fine. @gilles-peskine-arm wdyt?

gilles-peskine-arm: This is a [shell glob](https://en.wikipedia.org/wiki/Glob_(programming)) syntax, not a Unix regex syntax, so `.` stands for itself (and `?` stands for one arbitrary character).

So `?.*` means any string that's at least two characters long and where the second character is a dot. And `14.[012].*` includes `14.0.0`, `14.1.0`, `14.2.0` but not `14.3.0` or `14.10.0`.

tests/scripts/components-compiler.sh
gilles-peskine-arm: All components start with the default config.

bjwtaylor: Ok, happy to close.

tests/scripts/components-compiler.sh
bjwtaylor: Ok, happy to close.

bjwtaylor (approved):

PR #673, opened 2026-01-22
Only build mbedtls_rsa_deduce_private_exponent when key generation is enabled

### LINKED ISSUE #669: Only build mbedtls_rsa_deduce_private_exponent when key generation is enabled
Since TF-PSA-Crypto 1.0, there is no way to import an RSA private key without the private exponent. We've already done some cleanup (https://github.com/Mbed-TLS/TF-PSA-Crypto/issues/576), but the now private function `mbedtls_rsa_deduce_private_exponent()` is still present in the build and used in tests.

`mbedtls_rsa_deduce_private_exponent()` needs to remain in the code because it's used during key generation. But we don't need it in builds without RSA key generation.

There is test data in `test_suite_rsa.data` which has key pairs given only by (N, E, Q, P) without D. We should rewrite this test data to have complete RSA key data, in particular with D. Strictly speaking, the goal of this issue is only to add the private exponent to the data; however we might want to move directly to the standard ASN.1 representation of private keys with all the fields.

Goals of this issue:

* `test_suite_rsa.data` doesn't have key pairs given only by (N, E, Q, P).
* In `test_suite_rsa.function`, `mbedtls_rsa_test_fill_context()` doesn't try to fill in for a missing D. (Of course, in public keys, only N and E are present.)
* `mbedtls_rsa_deduce_private_exponent` is guarded by `MBEDTLS_PSA_BUILTIN_WANT_RSA_KEY_PAIR_GENERATE` (or `MBEDTLS_GENPRIME`, which is the legacy symbol with the same meaning).

valeriosetti:
## Description

Resolves #669

## PR checklist

- [ ] **changelog** not required because: `mbedtls_rsa_deduce_private_exponent` in an internal function so adding a guard to it is not something visible to the user.
- **tests**  provided: updated as requested in the issue


bjwtaylor (dismissed):

bjwtaylor:
The CI looks broken, is this expected?

valeriosetti:
> The CI looks broken, is this expected?

Nope, there was a problem with the last minute rebase that I did to resolve a conflict on the RSA test data before leaving for vacations. Now it should be fixed. Let's see what the CI thinks. Once it's happy I will ping you again.

bjwtaylor (approved):

PR #683, opened 2026-02-04
PSA API for SHAKE

### LINKED ISSUE #627: XOF PSA driver interface and built-in interface
Implement the PSA XOF interface at the driver wrapper level and in built-ins, with support for SHAKE.

Expected test coverage:

* SHAKE through the built-in interface (good cases only, including multipart testing).
* SHAKE through driver wrappers (good cases only, including multipart testing). Normally we'd rely on testing the API, but we aren't doing the API yet.
* XOF test driver.

Note: we aren't doing the PSA XOF _application_ interface yet due to unresolved technical debt in psasim which makes it hard to add a new API function.

Prototyped in https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/626

Prerequisite: https://github.com/Mbed-TLS/TF-PSA-Crypto/issues/621

gilles-peskine-arm:
Add SHAKE128 and SHAKE256 through the PSA API. Resolves https://github.com/Mbed-TLS/TF-PSA-Crypto/issues/627 and more.

Status: ready.

Needs preceding PR:

1. [x] https://github.com/Mbed-TLS/mbedtls-framework/pull/265
2. [x] https://github.com/Mbed-TLS/mbedtls/pull/10600 — Update the framework to the merge of https://github.com/Mbed-TLS/mbedtls-framework/pull/265 (needed for psasim to pass in the TF-PSA-Crypto CI on branches where #683 is merged).

## PR checklist

- [x] **changelog** provided
- [x] **framework PR** provided https://github.com/Mbed-TLS/mbedtls-framework/pull/265
- [x] **mbedtls development PR** prerequisite to update the framework: https://github.com/Mbed-TLS/mbedtls/pull/10600
- **tests**  provided


drivers/builtin/src/psa_crypto_xof.c
```diff
@@ -0,0 +1,149 @@
...
+    switch (operation->alg) {
+
+#if defined(MBEDTLS_PSA_BUILTIN_ALG_SHAKE128)
+        case PSA_ALG_SHAKE128:
+#endif
+#if defined(MBEDTLS_PSA_BUILTIN_ALG_SHAKE256)
+        case PSA_ALG_SHAKE256:
+#endif
+#if defined(MBEDTLS_PSA_BUILTIN_ALG_SOME_SHAKE)
+    mbedtls_sha3_update(&operation->ctx.shake, input, input_length);
```
gilles-peskine-arm, line 109: The weird indentation is an existing bug (see e.g. `psa_crypto_hash.c`) which I hadn't noticed until now, and I've [reported](https://github.com/uncrustify/uncrustify/issues/4622). To be fair, what we're doing here with guarded case statements is weird. But it makes perfect sense for what we're doing — multiple features that can be activated independently at compile time, but that have very similar implementations.

include/psa/crypto_values.h
```diff
@@ -970,6 +983,36 @@
...
+ *
+ * \return              \c 1 if \p xof_alg supports a context parameter
+ *                      passed with psa_xof_set_context(). This includes
+ *                      XOF algorithms with an optional context.
+ *                      \c 0 if \p xof_alg does not allow a context parameter.
+ *                      Unspecified if \p xof_alg is not a supported
+ *                      XOF algorithm.
+ */
+#define PSA_ALG_XOF_HAS_CONTEXT(xof_alg)        \
+    (((alg) & PSA_ALG_XOF_CONTEXT_FLAG) != 0)
```
bjwtaylor: Should this be xof_alg? Or are you relying on the caller to declare alg prior to calling the macro?

drivers/builtin/src/psa_crypto_xof.c
bjwtaylor, line 109: So, here we are ignoring the return code. I suspect this is because the function currently cannot fail. So in this case it should not return an error, however should we be checking the return code anyway to capture any potential future errors as the code evolves?

drivers/builtin/src/psa_crypto_xof.c
```diff
@@ -0,0 +1,149 @@
...
+{
+    /* A context must be freshly initialized before it can be set up. */
+    if (operation->alg != 0) {
+        return PSA_ERROR_BAD_STATE;
+    }
+
+    switch (alg) {
+#if defined(MBEDTLS_PSA_BUILTIN_ALG_SHAKE128)
+        case PSA_ALG_SHAKE128:
+            mbedtls_sha3_starts(&operation->ctx.shake, MBEDTLS_SHA3_SHAKE128);
```
bjwtaylor, line 62: So, I guess the return code is ignored here because we check MBEDTLS_PSA_BUILTIN_ALG_SHAKE128 is defined and if this defined MBEDTLS_SHA3_WANT_SHAKE128 will also be defined and this will currently always pass. However for future maintainability should the error code be checked to allow for future evolution of the function.

drivers/builtin/src/psa_crypto_xof.c
```diff
@@ -0,0 +1,149 @@
...
+            break;
+#endif
+#if defined(MBEDTLS_PSA_BUILTIN_ALG_SHAKE256)
+        case PSA_ALG_SHAKE256:
+            mbedtls_sha3_starts(&operation->ctx.shake, MBEDTLS_SHA3_SHAKE256);
```
bjwtaylor, line 67: So, I guess the return code is ignored here because we check MBEDTLS_PSA_BUILTIN_ALG_SHAKE256 is defined and if this defined MBEDTLS_SHA3_WANT_SHAKE256 will also be defined and this will currently always pass. However for future maintainability should the error code be checked to allow for future evolution of the function.

drivers/builtin/src/psa_crypto_xof.c
```diff
@@ -0,0 +1,149 @@
...
+            (void) input_length;
+            return PSA_ERROR_BAD_STATE;
+    }
+}
+
+psa_status_t mbedtls_psa_xof_output(
+    mbedtls_psa_xof_operation_t *operation,
+    uint8_t *output, size_t output_size)
+{
+    /* TODO: fill output with something "safe" in case of error.
```
bjwtaylor, line 124: Is this TODO expected to be merged?

bjwtaylor (commented):
Still need to do a full review, though a quick few things I noticed and had questions on.

include/psa/crypto_values.h
gilles-peskine-arm: It should be `xof_alg`.

This didn't break the build because the only place where this macro is used happens to pass `alg` as the argument!

drivers/builtin/src/psa_crypto_xof.c
gilles-peskine-arm: There won't be future errors. The low-level modules don't allocate memory, don't use potentially unavailable hardware, and aren't called with invalid arguments, which are the only reasons they can fail. If we were writing `sha3.c` from scratch, we'd make the functions return `void`, to make the code simpler and smaller.

It was different before TF-PSA-Crypto 1.0 because low-level modules could have alternative implementations (`MBEDTLS_SHA3_ALT`) which could return errors (at least `MBEDTLS_ERR_PLATFORM_HW_ACCEL_FAILED`).

We don't have time to clean up all the low-level modules, or even just `sha3.c`, but there's no point in catching errors that won't happen.

drivers/builtin/src/psa_crypto_xof.c
gilles-peskine-arm: Maybe. We have a best-effort guideline to return “safe” outputs on failure. Some aspects are clear, for example don't leak any part of the plaintext if decryption fails. For something like a hash calculation, we set the output to something that isn't the hash of anything (so all-bytes-zero works). But what is “safe” for a XOF? I'm not sure there's any good answer.

drivers/builtin/src/psa_crypto_xof.c
gilles-peskine-arm: How I think a low-level crypto module interface should look like nowadays: https://github.com/gilles-peskine-arm/TF-PSA-Crypto/blob/8fd3b6d1dfb232a8e6430f4a11b9759bf4e1c982/drivers/builtin/src/ascon_internal.h

drivers/builtin/include/mbedtls/private/crypto_builtin_primitives.h
```diff
@@ -80,6 +80,29 @@ typedef struct {
 
 #define MBEDTLS_PSA_HASH_OPERATION_INIT { 0, { 0 } }
 
+/*
+ * XOF (extendable-output functions) multi-part operation definitions.
+ */
+
+#if defined(MBEDTLS_PSA_BUILTIN_ALG_SHAKE128) || \
+    defined(MBEDTLS_PSA_BUILTIN_ALG_SHAKE256) || \
+    0
```
valeriosetti, line 89:
```suggestion
    defined(MBEDTLS_PSA_BUILTIN_ALG_SHAKE256)
```
?

valeriosetti (changes requested):
Only one very minor detail, but a part from that it looks good to me

drivers/builtin/include/mbedtls/private/crypto_builtin_primitives.h
gilles-peskine-arm: Why? The idea is to use 0 as a terminator so that adding an algorithm doesn't require changing the previous line.

gilles-peskine-arm:
I amended the commit to update the framework submodule to use the head of `main`, now that the framework prerequisite is merged. No change of base was necessary. I fixed https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/683#discussion_r2792741889 in a new commit. I don't think any of the other comments require a change.

drivers/builtin/include/mbedtls/private/crypto_builtin_primitives.h
valeriosetti: Yeah, I guessed that, but it's not a pattern I've seen being used that often in the project. Anyway

valeriosetti (approved):

drivers/builtin/src/psa_crypto_xof.c
bjwtaylor: Ok, happy to close this then.

bjwtaylor (approved):

Issue #476, opened 2025-09-25
Large key buffer allocated on stack in pk.c causes issues on constrained devices


Ashish285:
### Description
In [drivers/builtin/src/pk.c#L727](https://github.com/Mbed-TLS/TF-PSA-Crypto/blob/development/drivers/builtin/src/pk.c#L727), a key buffer is allocated directly on the stack.

```
unsigned char data[PSA_KEY_EXPORT_MAX_SIZE];
```

On edge devices with limited stack memory, this can cause stack overflows or severely limit available stack space for other operations.

### Suggested enhancement

* Consider allocating this buffer from the heap or using a caller-provided buffer.
* If a fixed buffer is required, make it configurable at compile time (e.g., MBEDTLS_PSA_KEY_BUFFER_ALLOC).

mpg:
Hi @Ashish285 and thanks for your report!

You're making a good point. When we added this buffer it was only used for ECC keys, which are always relatively small, so I think a stack buffer was appropriate at that time. Later we started using it for RSA keys as well, making the maximum size much larger, and we didn't revisit the location of this buffer.

> Consider allocating this buffer from the heap or using a caller-provided buffer.

I think that's probably our best option. PK already relies on heap allocations quite heavily - until 3.6 it directly made allocations, and even now it indirectly makes a lot of them via bignum under the hood. So I don't think it makes sense to avoid a heap allocation here.

> If a fixed buffer is required, make it configurable at compile time (e.g., MBEDTLS_PSA_KEY_BUFFER_ALLOC)

I don't think a fixed buffer is required. However, regarding compile-time configuration, note that if your application is only using ECC keys, you'll want to make sure support for RSA keys (`PSA_WANT_KEY_TYPE_RSA_PUBLIC`) is disabled in your config. This should make the size of this buffer more reasonable.

gilles-peskine-arm:
We're overdue for adding PQC support. It's likely that PK will start supporting ML-DSA in 2026, and you really don't want an ML-DSA public key on the stack. Or even just to consume the size of the largest ML-DSA key in all PK objects even when you happen to be doing ECC.

mpg:
Sharing some thoughts I had about this.

[...]

PR #684, opened 2026-02-05
PK write: avoid large stack buffer

mpg:
## Description

Avoid uselessly allocating a large buffer on the stack when writing out RSA keypairs.

A step towards addressing https://github.com/Mbed-TLS/TF-PSA-Crypto/issues/476

## PR checklist

- [x] **changelog** not required because: will be added when we finally fix 476
- [x] **mbedtls 3.6 PR** provided https://github.com/Mbed-TLS/mbedtls/pull/10592
- **tests**  provided

### DIFF
--- a/drivers/builtin/src/pkwrite.c
+++ b/drivers/builtin/src/pkwrite.c
@@ -37,22 +37,22 @@
 static int pk_write_rsa_der(unsigned char **p, unsigned char *buf,
                             const mbedtls_pk_context *pk)
 {
-    uint8_t tmp[PSA_KEY_EXPORT_RSA_KEY_PAIR_MAX_SIZE(PSA_VENDOR_RSA_MAX_KEY_BITS)];
-    size_t tmp_len = 0;
+    psa_status_t status;
+    size_t buf_size = (size_t) (*p - buf);
+    size_t key_len = 0;
 
-    if (psa_export_key(pk->priv_id, tmp, sizeof(tmp), &tmp_len) != PSA_SUCCESS) {
-        return MBEDTLS_ERR_PK_BAD_INPUT_DATA;
+    status = psa_export_key(pk->priv_id, buf, buf_size, &key_len);
+    if (status != PSA_SUCCESS) {
+        return status;
     }
-    /* Ensure there's enough space in the provided buffer before copying data into it. */
-    if (tmp_len > (size_t) (*p - buf)) {
-        mbedtls_platform_zeroize(tmp, sizeof(tmp));
-        return MBEDTLS_ERR_ASN1_BUF_TOO_SMALL;
-    }
-    *p -= tmp_len;
-    memcpy(*p, tmp, tmp_len);
-    mbedtls_platform_zeroize(tmp, sizeof(tmp));
 
-    return (int) tmp_len;
+    /* We wrote to the beginning of the buffer while
+     * we were supposed to write to its end. */
+    *p -= key_len;
+    memmove(*p, buf, key_len);
+    mbedtls_platform_zeroize(buf, *p - buf);
+
+    return (int) key_len;
 }
 
 static int pk_write_rsa_pubkey(unsigned char **p, unsigned char *start,
--- a/tests/suites/test_suite_pkwrite.function
+++ b/tests/suites/test_suite_pkwrite.function
@@ -77,9 +77,7 @@ static void pk_write_check_context(mbedtls_pk_context *key,
                                    unsigned char *check_buf, size_t check_buf_len)
 {
     unsigned char *buf = NULL;
-    int expected_error = is_der ?
-                         MBEDTLS_ERR_ASN1_BUF_TOO_SMALL :
-                         MBEDTLS_ERR_BASE64_BUFFER_TOO_SMALL;
+    int expected_error = PSA_ERROR_BUFFER_TOO_SMALL;
 
     /* Test with:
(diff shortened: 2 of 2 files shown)


tests/suites/test_suite_pkwrite.function
```diff
@@ -104,59 +142,31 @@ static void pk_write_check_common(char *key_file, int is_public_key, int is_der)
...
-     * that the writing functions fails as expected. */
-    for (size_t i = 1; i < buf_len; i++) {
-        TEST_EQUAL(pk_write_any_key(&key, &start_buf, &i, is_public_key,
-                                    is_der), expected_result);
-    }
-    TEST_EQUAL(pk_write_any_key(&key, &start_buf, &buf_len, is_public_key,
-                                is_der), 0);
-
-    TEST_MEMORY_COMPARE(start_buf, buf_len, check_buf, check_buf_len);
+    pk_write_check_context(&key, is_public_key, is_der,
```
valeriosetti: As I was [recently remembered](https://github.com/Mbed-TLS/mbedtls-framework/pull/257#discussion_r2686772354) by @gilles-peskine-arm, using macros like `TEST_EQUAL()` does not cause the test function to terminate immediately. It just marked as failed but then it continues until the next real failure (if any) or to the normal termination.
As a consequence I think that `pk_write_check_context` should return an error on failure and then consuming functions like `pk_write_check_common` should just do:
```
ret = pk_write_check_context(...)
if (ret != 0) {
   goto exit;
}
```

valeriosetti (changes requested):
All good. I only left a comment for improving test code.

tests/suites/test_suite_pkwrite.function
mpg: Ah, thanks for the reminder! I don't think really a problem here: if the caller just continues to execute after a failure, it just possibly test again with a wrapping context, which is admittedly a waste of time, but not causing any real issues.

Anyway, I'm changing as you suggested, for the sake of robustness.

valeriosetti (approved):

tests/suites/test_suite_pkwrite.function
valeriosetti: Yeah, I agree that in this case this wouldn't have been problematic. My intent is mostly to keep some good practices and common patterns so that if in the future someone else would like to implement something and take a look at what's already there, they will find a good reference

valeriosetti: Thanks for addressing this!

drivers/builtin/src/pkwrite.c
```diff
@@ -37,22 +37,22 @@
...
+    size_t key_len = 0;
 
-    if (psa_export_key(pk->priv_id, tmp, sizeof(tmp), &tmp_len) != PSA_SUCCESS) {
-        return MBEDTLS_ERR_PK_BAD_INPUT_DATA;
-    }
-    /* Ensure there's enough space in the provided buffer before copying data into it. */
-    if (tmp_len > (size_t) (*p - buf)) {
-        mbedtls_platform_zeroize(tmp, sizeof(tmp));
-        return MBEDTLS_ERR_ASN1_BUF_TOO_SMALL;
+    status = psa_export_key(pk->priv_id, buf, buf_size, &key_len);
```
bjwtaylor, line 44: I know this is pre-existing. However does the psa_export_key guarantee not to leak during a failure? If not is worth zeroising the buffer? Maybe even worth doing it anyway just as good practice?

bjwtaylor (commented):
Just one small question.

drivers/builtin/src/pkwrite.c
gilles-peskine-arm: All PSA API functions are supposed to leave “safe” outputs on failure. That in particular means no confidential data if the function fails.

mpg: Aha, I asked myself the same question while editing the code, then got focused on something else and forgot about this, so thank you Ben for raising it :)

> All PSA API functions are supposed to leave “safe” outputs on failure. That in particular means no confidential data if the function fails.

Looking at the API reference, [5.3.2](https://arm-software.github.io/psa-api/crypto/1.4/overview/conventions.html#behavior-on-error) says: "the content of output parameters is not defined", but a bit below: "When a function returns an error status, it is recommended that implementations set output parameters to safe defaults to avoid leaking confidential data and limit risk, in case an application does not properly handle all errors."

So I guess users should not rely on the content being safe, but as implementers we are imposing this on ourselves as a matter of implementation quality.

But here in PK we're users of the PSA Crypto API, so perhaps we should not make any assumption beyond what the API reference guarantees? Also, if drivers are in use, I don't think we can assume they're going to follow the recommendation as we do.

mpg: I think the conclusion of yesterday's discussion is that no change is needed here. It is still under discussion, as part of the effort on standardising the driver API, whether the responsibility of ensuring the `psa_export_key()` doesn't leave confidential data around on failure is on the driver or the core, but in any case it is not PK's responsibility.

As a result, I think no change is needed here. @bjwtaylor if that's also your understanding of the discussion, I think this discussion can be marked as resolved.

bjwtaylor: Agreed, I think we can resolve this as per the discussion yesterday.

bjwtaylor (approved):

PR #679, opened 2026-01-30
PK: fix (some) size macros

mpg:
## Description

Resolves https://github.com/Mbed-TLS/TF-PSA-Crypto/issues/671 and addresses the silly mistakes identified in https://github.com/Mbed-TLS/TF-PSA-Crypto/issues/476#issuecomment-3804806635

## PR checklist

- [x] **changelog** not required because: internal fixes
- [x] **framework PR** not required
- [x] **mbedtls development PR** not required because: crypto only
- [x] **mbedtls 3.6 PR**  not required because: 1.x only
- **tests**  provided

PR #687, opened 2026-02-10
PK: avoid large stack buffers in "to/from PSA" functions

### LINKED ISSUE #476: Large key buffer allocated on stack in pk.c causes issues on constrained devices

mpg:
## Description

Fixes https://github.com/Mbed-TLS/TF-PSA-Crypto/issues/476 (together with https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/684 and #679).

### Design notes

The design goals are:
- don't use large stack buffers (currently 2k+ in the default config), see #476;
- at the same time, preserve the ability to use PK without a heap when ECC is the only key type (and drivers are in use).

I took the approach of deciding at compile time whether to use a stack or a heap buffer, based on whether non-ECC key types are enabled in the build.

I hesitated with an alternative approach where we always have a stack buffer, sized for ECC keys, and (if other key types are enabled) we decide at runtime whether we should use it or allocate a larger buffer. Compared to the approach in the PR, the upside is it avoid a heap allocation when using ECC keys in a build with RSA also enabled. The downside is we always have a stack buffer, even when it's useless. (Edit: I think the compile-time approach taken in this PR also has the advantage of being simpler.)

The common point in both approaches is they're based on the set of key types enabled in PK: namely if it's just ECC, we can function without a heap, otherwise (RSA now, ML-DSA later) we use the heap. That's not the only way, we could differentiate based on a new compile-time option, or piggy-back on the existing `MBEDTLS_PSA_STATIC_KEY_SLOTS`. I chose the present approach because it works now without externally-visible change (new option or semantic extension of an existing one) and I don't think it prevents us from adopting another strategy in the future should we choose to do so.

## PR checklist

- [x] **changelog** provided (externally reported)
- [x] **mbedtls 3.6 PR** provided Mbed-TLS/mbedtls#10609
- **tests**  not required because: covered by existing tests (these functions are recent so they should be well-tested already)


drivers/builtin/src/pk.c
```diff
@@ -707,20 +716,69 @@ psa_key_type_t mbedtls_pk_get_key_type(const mbedtls_pk_context *pk)
     return pk->psa_type;
 }
 
+#if defined(PK_HAVE_KEYS_LARGER_THAN_ECC)
+static size_t pk_export_max_size(psa_key_type_t key_type, size_t bits)
```
gilles-peskine-arm: Is this measurably advantageous over `PSA_EXPORT_KEY_OUTPUT_SIZE`?

drivers/builtin/src/pk.c
```diff
@@ -25,6 +25,15 @@
 #include <limits.h>
 #include <stdint.h>
 
+#if defined(PSA_WANT_KEY_TYPE_RSA_PUBLIC_KEY)
+#define PK_HAVE_KEYS_LARGER_THAN_ECC
+#endif
+
+#if defined(PK_HAVE_KEYS_LARGER_THAN_ECC)
```
gilles-peskine-arm: Please add a comment to summarize what this is used for.

I'm not sure “have keys larger than ECC” is the right name. I think it would make more sense to name the macro based on its impact than on its trigger.

----

Having read how it's used, I think it should be named something like `PK_CAN_EXPORT_KEYS_ON_STACK` — with heap being the default case, and small-enough-for-the-stack being an optimization.

drivers/builtin/src/pk.c
```diff
@@ -707,20 +716,69 @@ psa_key_type_t mbedtls_pk_get_key_type(const mbedtls_pk_context *pk)
...
+                                           psa_key_type_t old_type, size_t old_bits,
                                            const psa_key_attributes_t *attributes,
                                            mbedtls_svc_key_id_t *new_key_id)
 {
-    unsigned char key_buffer[PSA_EXPORT_KEY_PAIR_MAX_SIZE];
+#if defined(PK_HAVE_KEYS_LARGER_THAN_ECC)
+    unsigned char *key_buffer = NULL;
+    size_t key_buffer_size = 0;
+#else
+    unsigned char *key_buffer[MBEDTLS_PSA_MAX_EC_KEY_PAIR_LENGTH];
```
gilles-peskine-arm: I'm pretty sure this should be
```suggestion
    unsigned char key_buffer[MBEDTLS_PSA_MAX_EC_KEY_PAIR_LENGTH];
```

Happens again further down.

extras/pk.c
```diff
@@ -707,20 +716,69 @@ psa_key_type_t mbedtls_pk_get_key_type(const mbedtls_pk_context *pk)
...
+        return PSA_KEY_EXPORT_RSA_KEY_PAIR_MAX_SIZE(bits);
+    }
+#endif
+    /* failsafe */
+    return PSA_EXPORT_KEY_PAIR_MAX_SIZE;
+}
+#endif /* PK_HAVE_KEYS_LARGER_THAN_ECC */
+
 static psa_status_t export_import_into_psa(mbedtls_svc_key_id_t old_key_id,
+                                           psa_key_type_t old_type, size_t old_bits,
```
gilles-peskine-arm, line 716: I don't understand why the function takes additional parameters. Under what circumstances could `old_type` and `old_bits` differ from `psa_get_key_type(attributes)` and `psa_get_key_bits(attributes)`?

drivers/builtin/src/pk.c
gilles-peskine-arm: Can this code never be called with public keys?

If it can then the array size is too small. But maybe it can't (since this is a fallback for `psa_copy_key` to bypass policy restrictions — but I don't remember without looking it up if `psa_copy_key` bypasses policy restrictions for public keys the same way export does).

Note that as of 626a230590be34a484b9c47c1423c4c286f1a2cb, `pk_export_max_size` has cases for public keys, but its stack counterpart is always sized for private keys.

drivers/builtin/src/pk.c
```diff
@@ -907,8 +967,13 @@ static int copy_from_psa(mbedtls_svc_key_id_t key_id,
     psa_key_attributes_t key_attr = PSA_KEY_ATTRIBUTES_INIT;
     psa_key_type_t key_type;
     size_t key_bits;
-    /* Use a buffer size large enough to contain either a key pair or public key. */
-    unsigned char exp_key[PSA_EXPORT_KEY_PAIR_OR_PUBLIC_MAX_SIZE];
+#if defined(PK_HAVE_KEYS_LARGER_THAN_ECC)
+    unsigned char *exp_key = NULL;
+    size_t exp_key_size = 0;
+#else
+    unsigned char *exp_key[MBEDTLS_PSA_MAX_EC_KEY_PAIR_LENGTH];
```
gilles-peskine-arm: Two problems: stray `*` on the array declaration, and wrong length because it doesn't accept public keys.

extras/pk.c
```diff
@@ -994,12 +1067,16 @@ static int copy_from_psa(mbedtls_svc_key_id_t key_id,
 #endif /* PSA_WANT_KEY_TYPE_ECC_PUBLIC_KEY */
     {
         (void) key_bits;
-        return MBEDTLS_ERR_PK_BAD_INPUT_DATA;
+        ret = MBEDTLS_ERR_PK_BAD_INPUT_DATA;
+        goto exit;
```
gilles-peskine-arm, line 1072: Is this fixing a resource leak or missing zeroization on failure?

drivers/builtin/src/pk.c
gilles-peskine-arm, line 850: Shouldn't this stack buffer get the same treatment as the other two? (There are three arrays on the stack in `pk.c`, I haven't looked at the other `pk*.c`.)

extras/pk.c
```diff
@@ -25,6 +25,15 @@
 #include <limits.h>
 #include <stdint.h>
```
gilles-peskine-arm, line 27: About the design: I don't like this compile-time selection of stack vs heap much because it's yet another compile-time variation, but of all the solutions I can think of, it's the one I dislike the least.

I prefer this compile-time choice to a runtime choice because the runtime choice means more code size, a compile-time choice anyway so that it's possible to build without malloc, and wasted space on the stack if the heap is going to be used anyway.

I have a slight preference for basing the stack-vs-heap selection on `MBEDTLS_PSA_STATIC_KEY_SLOTS` (or a rebranded option `TF_PSA_CRYPTO_WITHOUT_MALLOC`), so that we're consistent about using vs avoiding heap throughout the library. But I'm perfectly fine with ECC-only-or-not for now, and we can revise it after 1.1.0.

gilles-peskine-arm, line 27: About test coverage: I don't think we're missing testing of code paths. But I'd like some evidence in the outcome analysis that the non-default case is tested. Maybe define the symbol in `pk_internal.h` and add cases for it to `test_suite_config`?

gilles-peskine-arm (changes requested):
Approved for design. The execution needs work.

Note: regarding completeness, I'm assuming that the scope is strictly `pk.c`, excluding auxiliary functions e.g. from other `pk*.c` files.

drivers/builtin/src/pk.c
mpg: Aw, embarrassing.

extras/pk.c
mpg: This function is `static`, called from a single place and the called ensures that `old_type == psa_get_key_type(attributes)` so that's OK. My problem is I couldn't convince myself that there is any guarantee that `bits` is set in the new attributes, ie that `psa_get_key_bits(attributes) != 0`.

So, I mainly wanted to add `old_bits` just to be sure, and added `old_type` for symmetry.

drivers/builtin/src/pk.c
mpg: Ah, I'm afraid I just has the brain bug that I implicitly though key pair was the max, while I know very well for ECC it is not, but apparently I just failed to think about it.

I'll double-check if there's any reason this code path can't be reached with public keys, but if there is that's pure luck, not something I had checked. (And if that happens to be the case, I'll add a comment because that's not obvious).

extras/pk.c
mpg: It's not fixing a pre-existing issue, but since we're now doing an allocation earlier in this function, we would have a resource leak (on the error path) without this change.

drivers/builtin/src/pk.c
mpg: This buffer:
1. Is 528 bytes while the other two are 2364 bytes (in the default config), so I'd say lower priority.
2. Will be entirely removed by https://github.com/Mbed-TLS/mbedtls/issues/10583 anyway (just waiting for https://github.com/Mbed-TLS/mbedtls/issues/10573 first).

mpg: Regarding other buffers in `pk*.c` (excluding `pkcs*.c`) I believe I've got a complete list [there](https://github.com/Mbed-TLS/TF-PSA-Crypto/issues/476#issuecomment-3804650796) with an analysis [below](https://github.com/Mbed-TLS/TF-PSA-Crypto/issues/476#issuecomment-3804806635).

I think all of them are covered by existing PRs or issues, except for the one in `struct mbedtls_pk_context` which is smaller than the buffers the OP was pointing out, and perhaps a bit more complex to fix since it's in a structure, so I [was suggesting](https://github.com/Mbed-TLS/TF-PSA-Crypto/issues/476#issuecomment-3852996548) we consider only the larger buffers in scope for https://github.com/Mbed-TLS/TF-PSA-Crypto/issues/476.

Though I believe for the buffer in the context structure we can use a very similar approach to the one in this PR. And once we have proper testing from https://github.com/Mbed-TLS/mbedtls/issues/10573 (and the experience from the present PR) it shouldn't be hard to do.

drivers/builtin/src/pk.c
mpg: Hmm, tbh the reason I introduced this is I wasn't aware of `PSA_EXPORT_KEY_OUTPUT_SIZE`. I hate to blame my ignorance on other people's naming choices, but the fact that the other macros are called `PSA_KEY_EXPORT_...` but this one `PSA_EXPORT_KEY_...` was probably a factor in me not discovering this macro through completion.

Anyway, now that I know it exists... perhaps there is an advantage in code size, as the compiler can skip the paths for key types other than ECC or RSA. Also I guess having a static function makes it easier for have a single copy of the code.

OTOH, I think the pros of using `PSA_EXPORT_KEY_OUTPUT_SIZE()`  are the usual pros for not re-inventing the wheel: obvious correctness, readability... except code size since it's a macro. (Btw I wonder if we should also have functions for all the non-trivial macros, for use in places that don't require the result to be a compile-time constant. But that's a larger topic and quite out of scope here.)

I should probably make measurements so that we can decide based on facts just not thoughts.

drivers/builtin/src/pk.c
mpg: Agreed with naming it based on impact rather than trigger. I think I'd be inclined to go the other way though: `PK_NEED_HEAP_FOR_KEY_EXPORT`, as the will be more code when using the heap than when using the stack. 

Which one we see as the default vs an optimisation is mainly in our minds but doesn't really affect how much code is guarded by "if using the heap" vs "if using the stack" and I think I prefer the path with the most code to be guarded with a positive condition (ie, `#if defined(FOO)` rather than `if !defined(BAR)`, though I'm not sure I could explain why.

drivers/builtin/src/pk.c
gilles-peskine-arm: > the other macros are called PSA_KEY_EXPORT_... but this one PSA_EXPORT_KEY_...

We settled on [`PSA_EXPORT_KEY_xxx_SIZE`](https://arm-software.github.io/psa-api/crypto/1.4/api/keys/management.html#c.PSA_EXPORT_KEY_OUTPUT_SIZE) for the standard macros, I think to follow the general pattern `upcase(function_name)+thing+`_SIZE`. For consistency, we should rename the semi-private (visible but no doxygen documentation) `PSA_KEY_EXPORT_xxx`.

> I wonder if we should also have functions for all the non-trivial macros, for use in places that don't require the result to be a compile-time constant. But that's a larger topic and quite out of scope here.

It's been on my mental to-do list (not sure if it's on github) for a long time for many reasons: save code size for the complicated ones, provide a way to use the library with a stable ABI, and maybe enable a tiny extra amount of type safety (not sure there's anything concrete though there since all the arguments are unsigned integers). It's not difficult, but it would take time to churn out the code and tests.

extras/pk.c
gilles-peskine-arm: I think `attributes.bits` can be 0 there. I'd prefer to set it from `old_attributes` in `copy_into_psa()` rather than add more parameters to `export_import_into_psa()`.

drivers/builtin/src/pk.c
gilles-peskine-arm: By the way, the macro [`PSA_EXPORT_ASYMMETRIC_KEY_MAX_SIZE`](https://arm-software.github.io/psa-api/crypto/1.4/api/keys/management.html#c.PSA_EXPORT_ASYMMETRIC_KEY_MAX_SIZE) was only added to the API in 1.3, and we don't define it yet. Now would be a good time to implement it.

extras/pk.c
mpg: That was my first thought as well, but we're taking a `const` pointer, and it's `const` all the way to the public function `mbedtls_pk_import_into_psa` so I don't think that's really an option.

I'll note this is a static function with a single call site, so very likely to get inlined even at `-Os` or `-Oz`. I'm not sure why adding a parameter to it is such a problem. (I agree that adding two while only one is need was probably misguided.)

mpg:
@gilles-peskine-arm If you don't mind I'm going to rewrite the history to fix two compilation failures in the commits that introduced them, then I'll address your other feedback with additional commits.

The commit you reviewed was 626a230590be34a484b9c47c1423c4c286f1a2cb

gilles-peskine-arm:
> @gilles-peskine-arm If you don't mind I'm going to rewrite the history to fix two compilation failures in the commits that introduced them, then I'll address your other feedback with additional commits.

Fine by me.

mpg:
Sorry for pushing something that was in such a bad shape (not even compiling in one of the obvious configs of interest) yesterday. What I just pushed at least passes `test_suite_pk` locally in the default config and default minus RSA.

I haven't addressed all of your comments yet, but I'd like to get CI feedback on this version.

extras/pk.c
mpg: > So, I mainly wanted to add `old_bits` just to be sure, and added `old_type` for symmetry.

Ah that's not true, I remember now why I added `old_type`: since `attributes` is for the new key, using it to get the type of the old key seems weird enough that IMO it would require adding a comment along the lines of "we're a static function with a single caller who checked that the type in the new attributes match the type of the old key, so this is fine", at which point adding a new parameter with an explicit name frankly looked like a better option.

Wdyt?

drivers/builtin/src/pk.c
mpg: > I should probably make measurements so that we can decide based on facts just not thoughts.

Building with
```
arm-none-eabi-gcc -mthumb -mcpu=cortex-m55 -Os -I include -I drivers/builtin/include -I core -c drivers/builtin/src/pk.c
```
I'm getting:
```
   text    data     bss     dec     hex filename
   3808       0       0    3808     ee0 pk-func.o
   3996       0       0    3996     f9c pk-macro.o
```
So that's a 188 bytes difference, which IMO is enough to justify the new function. I can add a comment explaining why we're not directly using the macro, though.

drivers/builtin/src/pk.c
mpg: If I understand correctly, `PSA_EXPORT_ASYMMETRIC_KEY_MAX_SIZE` is just the new name for what we currently call `PSA_EXPORT_KEY_PAIR_OR_PUBLIC_MAX_SIZE` in which case it seems very easy to add indeed.

extras/pk.c
gilles-peskine-arm: The function is copying a key, so it feels very natural to me that old attributes = new attributes unless there was a mistake (the “new attributes” come from the application so we can't trust them). There's just the wrinkle that the new attributes might be incomplete: the type needs to be specified explicitly but not the size.

drivers/builtin/src/pk.c
gilles-peskine-arm: Wow, that much?

Hmm, what if we remove support for finite-field DSA? (How did that survive so long‽)

gilles-peskine-arm: > How did that survive so long‽

Well, I tried. https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/321

gilles-peskine-arm: After removing DSA, we're left with
```
#define PSA_EXPORT_KEY_OUTPUT_SIZE(key_type, key_bits)                                              \
    (PSA_KEY_TYPE_IS_UNSTRUCTURED(key_type) ? PSA_BITS_TO_BYTES(key_bits) :                         \
     PSA_KEY_TYPE_IS_DH(key_type) ? PSA_BITS_TO_BYTES(key_bits) :                                   \
     (key_type) == PSA_KEY_TYPE_RSA_KEY_PAIR ? PSA_KEY_EXPORT_RSA_KEY_PAIR_MAX_SIZE(key_bits) :     \
     (key_type) == PSA_KEY_TYPE_RSA_PUBLIC_KEY ? PSA_KEY_EXPORT_RSA_PUBLIC_KEY_MAX_SIZE(key_bits) : \
     PSA_KEY_TYPE_IS_ECC_KEY_PAIR(key_type) ? PSA_KEY_EXPORT_ECC_KEY_PAIR_MAX_SIZE(key_bits) :      \
     PSA_KEY_TYPE_IS_ECC_PUBLIC_KEY(key_type) ? PSA_KEY_EXPORT_ECC_PUBLIC_KEY_MAX_SIZE(key_bits) :  \
     0u)
```

I don't think we rely on 0 as the fallback case anywhere, and I doubt that applications do either, but changing that fallback in an LTS would slightly worry me. But in `development`, how about simplifying this to fall back to the unstructured case?

```
#define PSA_EXPORT_KEY_OUTPUT_SIZE(key_type, key_bits)                                              \
     (key_type) == PSA_KEY_TYPE_RSA_KEY_PAIR ? PSA_KEY_EXPORT_RSA_KEY_PAIR_MAX_SIZE(key_bits) :     \
     (key_type) == PSA_KEY_TYPE_RSA_PUBLIC_KEY ? PSA_KEY_EXPORT_RSA_PUBLIC_KEY_MAX_SIZE(key_bits) : \
     PSA_KEY_TYPE_IS_ECC_KEY_PAIR(key_type) ? PSA_KEY_EXPORT_ECC_KEY_PAIR_MAX_SIZE(key_bits) :      \
     PSA_KEY_TYPE_IS_ECC_PUBLIC_KEY(key_type) ? PSA_KEY_EXPORT_ECC_PUBLIC_KEY_MAX_SIZE(key_bits) :  \
     PSA_BITS_TO_BYTES(key_bits)) /*unstructured; FFDH public or private*/
```

And we could further optimize in configurations without certain key types by making the construction of the definition more complicated.

mpg: With the new definition, the macro has negligible cost over my custom function (like 20 bytes), so let's do that!

mpg:
@gilles-peskine-arm I believe my last push addresses all of your comments. If the CI is happy with it, then it's ready for re-review.

extras/pk_internal.h
```diff
@@ -45,6 +45,28 @@
...
+#define PK_EXPORT_KEYS_ON_THE_STACK
+#endif
+
+#if defined(PK_EXPORT_KEYS_ON_THE_STACK)
+/* We know for ECC, pubkey are longer than privkeys, but double check */
+#define PK_EXPORT_KEY_STACK_BUFFER_SIZE  MBEDTLS_PSA_MAX_EC_PUBKEY_LENGTH
+#if MBEDTLS_PSA_MAX_EC_KEY_PAIR_LENGTH > PK_EXPORT_KEY_STACK_BUFFER_SIZE
+#undef PK_EXPORT_KEY_STACK_BUFFER_SIZE
+#define PK_EXPORT_KEY_STACK_BUFFER_SIZE  MBEDTLS_PSA_MAX_EC_KEY_PAIR_LENGTH
+#endif
```
gilles-peskine-arm, line 67: Why not

```suggestion
#define PK_EXPORT_KEY_STACK_BUFFER_SIZE PSA_EXPORT_ASYMMETRIC_KEY_MAX_SIZE
#endif
```
? It would be equivalent now, but would be a safe default if we start to implement a new algorithm.

include/psa/crypto_sizes.h
```diff
@@ -1027,10 +1023,16 @@
...
-#define PSA_EXPORT_KEY_PAIR_OR_PUBLIC_MAX_SIZE \
+/* This is the name that was standardized in PSA Crypto v1.3 */
+#define PSA_EXPORT_ASYMMETRIC_KEY_MAX_SIZE \
     ((PSA_EXPORT_KEY_PAIR_MAX_SIZE > PSA_EXPORT_PUBLIC_KEY_MAX_SIZE) ? \
      PSA_EXPORT_KEY_PAIR_MAX_SIZE : PSA_EXPORT_PUBLIC_KEY_MAX_SIZE)
 
+/* This is our old custom name from before it was in the spec,
+ * keep it around in case users were relying on it. */
+#define PSA_EXPORT_KEY_PAIR_OR_PUBLIC_MAX_SIZE \
+    PSA_EXPORT_ASYMMETRIC_KEY_MAX_SIZE
```
gilles-peskine-arm: I think we should mark it as deprecated.

```suggestion
#if !defined(MBEDTLS_DEPRECATED_REMOVED)
#define PSA_EXPORT_KEY_PAIR_OR_PUBLIC_MAX_SIZE \
    ((size_t) MBEDTLS_DEPRECATED_NUMERIC_CONSTANT(PSA_EXPORT_ASYMMETRIC_KEY_MAX_SIZE))
#endif
```
and move to `crypto_compat.h`

extras/pk_internal.h
gilles-peskine-arm: And come to think of it, I think it would make sense to base `PK_EXPORT_KEYS_ON_THE_STACK` on the value of `PSA_EXPORT_ASYMMETRIC_KEY_MAX_SIZE`. Like

```
/* Export keys on the stack, effectively, if ECC keys are the largest keys.
 * The threshold is enough for ECC P521, but too small for RSA-2048. */
#if PSA_EXPORT_ASYMMETRIC_KEY_MAX_SIZE < 200
#define PK_EXPORT_KEYS_ON_THE_STACK
#endif
```

or

```
/* Export keys on the stack if ECC keys are the largest keys */
#if PSA_EXPORT_ASYMMETRIC_KEY_MAX_SIZE <= MBEDTLS_PSA_MAX_EC_PUBKEY_LENGTH
#define PK_EXPORT_KEYS_ON_THE_STACK
#endif
```

gilles-peskine-arm (changes requested):
A couple of niggles, and missing changelog (since this was externally reported), otherwise LGTM

extras/pk_internal.h
mpg: >  It would be equivalent now

Not strictly: consider a configuration where ECC and FFDH are enabled, but not RSA. Then from PK's point of view, there are no keys larger than ECC, but from PSA's pov, there are.

> And come to think of it, I think it would make sense to base `PK_EXPORT_KEYS_ON_THE_STACK` on the value of `PSA_EXPORT_ASYMMETRIC_KEY_MAX_SIZE`.

I quite like this idea, as from the beginning of this PR I've felt a bit uncomfortable about the condition "if RSA disabled" which will need changing when we add a new key type (though I've been telling myself that when we add a new key, surely we'll add tests cases and surely they would catch the problem if we forgot to update to "if RSA and ML-DSA both disabled" - surely).

This is a very direct way to express what we want (max export size small enough), it's robust when we add new algorithms, including not accidentally reintroducing very large stack buffers (unlike the [first suggestion](https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/687#discussion_r2799733103) above I think).

The only drawback is that people with FFDH enabled will not get a heapless PK experience. OTOH, I'm not sure there are people who care about getting both a heapless PK and FFDH in their build?

gilles-peskine-arm: > > It would be equivalent now
>
> Not strictly: consider a configuration where ECC and FFDH are enabled, but not RSA. Then from PK's point of view, there are no keys larger than ECC, but from PSA's pov, there are.

Indeed. And in 1.1 or 1.2, we'll have MLDSA in PSA but not yet in PK, and ECC+MLDSA but no RSA (or FFDH) seems like a plausible configuration for systems that might want to avoid a heap.

> > And come to think of it, I think it would make sense to base PK_EXPORT_KEYS_ON_THE_STACK on the value of PSA_EXPORT_ASYMMETRIC_KEY_MAX_SIZE.

I'm actually swinging back of this because of MLDSA. Heapless PK for ECC only, plus MLDSA with no PK support, will make sense in the medium term, and in particular it may be possible in the 1.1 LTS branch. (Maybe: I don't know yet whether we'll reach this milestone before the 1.1 code freeze.)

> The only drawback is that people with FFDH enabled will not get a heapless PK experience. OTOH, I'm not sure there are people who care about getting both a heapless PK and FFDH in their build?

If it wasn't about PK, I wouldn't care. I don't think FFDH-but-no-RSA is important enough to optimize for.

mpg: Unless I'm mistaken there's nothing to do here after all.

gilles-peskine-arm (dismissed):
LGTM

include/psa/crypto_sizes.h
```diff
@@ -901,15 +901,11 @@
  *         If the parameters are not valid, the return value is unspecified.
  */
 #define PSA_EXPORT_KEY_OUTPUT_SIZE(key_type, key_bits)                                              \
-    (PSA_KEY_TYPE_IS_UNSTRUCTURED(key_type) ? PSA_BITS_TO_BYTES(key_bits) :                         \
-     PSA_KEY_TYPE_IS_DH(key_type) ? PSA_BITS_TO_BYTES(key_bits) :                                   \
-     (key_type) == PSA_KEY_TYPE_RSA_KEY_PAIR ? PSA_KEY_EXPORT_RSA_KEY_PAIR_MAX_SIZE(key_bits) :     \
+    ((key_type) == PSA_KEY_TYPE_RSA_KEY_PAIR ? PSA_KEY_EXPORT_RSA_KEY_PAIR_MAX_SIZE(key_bits) :     \
      (key_type) == PSA_KEY_TYPE_RSA_PUBLIC_KEY ? PSA_KEY_EXPORT_RSA_PUBLIC_KEY_MAX_SIZE(key_bits) : \
-     (key_type) == PSA_KEY_TYPE_DSA_KEY_PAIR ? PSA_KEY_EXPORT_DSA_KEY_PAIR_MAX_SIZE(key_bits) :     \
```
mpg, line 908: @gilles-peskine-arm Is this change acceptable in 3.6 or should I do something different for the backport?

extras/pk_internal.h
```diff
@@ -45,6 +45,28 @@
...
+/*
+ * We're trying to statisfy two kinds of users:
+ * - those who don't want to use the heap;
+ * - those who can't afford large stack buffers.
+ *
+ * The current compromise is that if ECC is the only key type supported in PK,
+ * then we export keys on the stack, and otherwise we use the heap.
+ *
+ * Note: add && !ML-DSA when adding support for ML-DSA */
+#if !defined(PSA_WANT_KEY_TYPE_RSA_PUBLIC_KEY)
```
bjwtaylor, line 57: Should this be PSA_WANT_KEY_TYPE_RSA_KEY_PAIR_BASIC?

mpg: I think `RSA_PUBLIC_KEY` is the right one here: it is (in principle*) possible to have builds that only support RSA public keys but not RSA keypairs, while the reverse is prohibited: if RSA keypairs are enabled, then RSA public keys [will be auto-enabled](https://github.com/Mbed-TLS/TF-PSA-Crypto/blob/development/include/tf-psa-crypto/private/crypto_adjust_config_key_pair_types.h#L50). That's not just a temporary thing, it's structural and unlikely to change.

*In practice I don't think we test it, so it probably doesn't actually work. But still, that's something that makes sense and that we'd like to support in the future, so code should be written with that in mind.

mpg (commented):
Thought I was leaving line comments, but apparently I had inadvertently started a review...

include/psa/crypto_sizes.h
gilles-peskine-arm: I think removing DSA is fine since you can't actually use a DSA key anywhere in Mbed TLS 3.6. I'm not sure about the fallback to 0, people could have inadvertently rely on it as a weak indicator of an invalid key type. But it wasn't a reliable indicator of an invalid key type (didn't detect invalid types that happen to have a bit pattern that matches the symmetric or unstructured category, didn't follow the library configuration). So I guess it's ok.

bjwtaylor (dismissed):
If we are all happy on the DSA removal, I am happy for this to be merged.

gilles-peskine-arm:
2f6f182af8872d391adbed1d67fbb7c0be73b2b2 is approved but there's a trivial (add/add) conflict.

mpg:
> there's a trivial (add/add) conflict.

I think it was an add/remove conflict, but anyway, I've resolved it.

valeriosetti (dismissed):

mpg:
There's a conflict again due to the file reorg, so I'll rebase and force-push again.

bjwtaylor (approved):

extras/pk.c
```diff
@@ -899,6 +931,31 @@ int mbedtls_pk_import_into_psa(const mbedtls_pk_context *pk,
     }
 }
 
+static int is_valid_for_pk(psa_key_type_t key_type)
```
valeriosetti, line 934: Question: shouldn't we add usage of this function also in `mbedtls_pk_wrap_psa`?

valeriosetti (approved):
I only have one potential addition proposal, but otherwise LGTM

extras/pk.c
mpg: I think `pk_wrap_psa()` already naturally checks the type while determining which info structure it should use: https://github.com/Mbed-TLS/TF-PSA-Crypto/blob/development/extras/pk.c#L173

But in the future when we change that (we want to get rid of info structures at some point) then yes we will probably want to use this function in `mbedtls_pk_wrap_psa()` as well.

Issue #633, opened 2026-01-05
Write PSA driver for ML-DSA using mldsa-native


gilles-peskine-arm:
Add a PSA driver for ML-DSA. Provide unit tests with data generated using https://github.com/GiacomoPope/dilithium-py.

In this issue, only implement key import/export/export-public (ML-DSA-87 only) and deterministic pure ML-DSA signature and verification, one-shot only, no context.

Write some basic functional testing. We should test every function, but we're only looking for integration mistakes, not for algorithmic mistakes. We assume that mldsa-native is already functionally correct (see https://github.com/Mbed-TLS/TF-PSA-Crypto/issues/641).

Prototype: https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/593

Prerequisite: https://github.com/Mbed-TLS/TF-PSA-Crypto/issues/628

Follow-ups: [key generation](https://github.com/Mbed-TLS/TF-PSA-Crypto/issues/634) (which it may be easier to do here, otherwise we might run into a problem with generated tests), [multipart sign-message and verify-message](https://github.com/Mbed-TLS/TF-PSA-Crypto/issues/635), [hedged signature](https://github.com/Mbed-TLS/TF-PSA-Crypto/issues/636), Hash-ML-DSA, [other parameter sets](https://github.com/Mbed-TLS/TF-PSA-Crypto/issues/632), context

52eason:
Hi Mbed TLS / TF-PSA-Crypto team,

We are evaluating ML-DSA for production firmware signature verification, where Mbed TLS is already used. The roadmap lists “ML-DSA – Initial support” for 2026 CQ2.

Could you share the estimated timeline for the release?

- Is 2026 CQ2 still the target?
- Will it include ML-DSA verification only, or sign + verify?
- Which parameter sets will be supported, especially ML-DSA-87?
- Will the implementation use **mldsa-native**?
- Should this be considered experimental, production-evaluation-ready, or production-ready?

Our use case is verify-only ML-DSA-87 firmware verification. We are deciding whether to wait for CQ2 support or temporarily integrate a separate backend behind our own abstraction.

Thank you.

gilles-peskine-arm:
We do not yet have MLDSA as an application-facing feature, and I don't have a timeline for that. We have started integrating mldsa-native, and a bootloader is also our top priority use case.

[...]

PR #690, opened 2026-02-11
Support committed generated test data and add PQCP test data

gilles-peskine-arm:
1. Prepare to support more committed generated files.
2. Prepare to support additional Python requirements from `framework/util/requirements.txt`.
3. Generate ML-DSA test data from a new script, in partial fulfillment of https://github.com/Mbed-TLS/TF-PSA-Crypto/issues/633. Requires steps 1 and 2.

A step towards https://github.com/Mbed-TLS/TF-PSA-Crypto/issues/633.

Needs preceding PR:

* [x] Small bug fix in mbedtls: https://github.com/Mbed-TLS/mbedtls/pull/10596
* [x] Framework part: https://github.com/Mbed-TLS/mbedtls-framework/pull/278

## PR checklist

- [x] **changelog** provided | not required because: 
- [x] **framework PR** provided https://github.com/Mbed-TLS/mbedtls-framework/pull/278
- [x] **mbedtls development PR** https://github.com/Mbed-TLS/mbedtls/pull/10626 (move from `check_option_lists.py` to `check_committed_generated_files.py`)
- [x] **mbedtls 3.6 PR** not required because: new features
- **tests**  provided


bjwtaylor (dismissed):
The PR looks good, I had one question mainly for interest as to what the plan is for testing code enabled by TF_PSA_CRYPTO_PQCP_MLDSA_87_ENABLED. A quick, but by no means extensive grep for this config didn't produce any results in the tests. Do we have CI tests currently added that will build and test with this config, or is this future planned work? Or have I misunderstood a part of the future architecture?

gilles-peskine-arm:
@bjwtaylor A lot of `all.sh` components, and the `Windows-NNNN` jobs, are based on the `full` config, which enables most commented out boolean options in `*config.h`. So new experimental features such as MLDSA are tested in these components unless explicitly disabled.

You can check which components run a specific test, or with a specific option enabled, in the [outcome file](https://mbed-tls.readthedocs.io/en/latest/kb/testing/testing-ci/#outcome-analysis). The outcome-analysis job on the CI fails if a test case is never executed.

mpg:
> * **mbedtls 3.6 PR** not required because: new features

Well, should we backport the removal of the line length limitation in test data files there? I feel in the future it's going to be annoying if we have some test data that works in crypto/developmnet but then fails when trying to backport to 3.6.

gilles-peskine-arm:
> should we backport the removal of the line length limitation in test data files there? I feel in the future it's going to be annoying if we have some test data that works in crypto/developmnet but then fails when trying to backport to 3.6.

I get your point. But also, in 3.6, currently, you can run some test suites without a heap, if the library is built without a heap. Granted, few test suites don't use malloc themselves anyway, and few users on constrainted platforms actually run the unit tests. But that's still the sort of requirement change that I'm uncomfortable with in an LTS branch.

valeriosetti (dismissed):

tests/suites/host_test.function
```diff
@@ -78,54 +71,81 @@ static int verify_int(char *str, intmax_t *p_value)
...
     do {
-        ret = fgets(buf, len, f);
+        ret = fgets(buffer->content + offset, buffer->size - offset, f);
         if (ret == NULL) {
-            return -1;
+            return (size_t) -1;
+        }
+        length = offset + strlen(buffer->content + offset);
+
+        if (length == buffer->size - 1 && buffer->content[length] != '\n') {
```
mpg: I think at this point we know `buffer->content[length]` is `'\0'` considering the way we computed `length`. You probably mean `buffer->content[length - 1]` here.

tests/suites/host_test.function
```diff
@@ -78,54 +71,81 @@ static int verify_int(char *str, intmax_t *p_value)
...
+            mbedtls_free(buffer->content);
+            buffer->content = new_buffer;
+            buffer->size = new_size;
+            offset = length;
+            continue;
         }
 
-        str_len = strlen(buf);
+        /* Now we have a full line. */
+        offset = 0;
```
mpg, line 128: I can't see this variable being used until the end of this function, so this looks like a dead store to me. Am I missing something?

tests/suites/host_test.function
```diff
@@ -78,54 +71,81 @@ static int verify_int(char *str, intmax_t *p_value)
...
+        }
+
+        if (length == 0) {
+            /* Skip empty line */
             continue;
         }
-        has_string = 0;
-        for (i = 0; i < str_len; i++) {
-            char c = buf[i];
-            if (c != ' ' && c != '\t' && c != '\n' &&
```
mpg, line 111: Ok, so it seems we used to have logic to skip lines consisting of only whitespace. I'm no longer seeing this logic in the new code. I don't think that's wrong (we probably don't have lines that are empty except for whitespace in our data files) but I'd like to confirm this was intentional.

tests/scripts/components-basic-checks.sh
```diff
@@ -31,6 +31,12 @@ component_tf_psa_crypto_check_python_files () {
     $FRAMEWORK/scripts/check-python-files.sh
 }
 
+support_tf_psa_crypto_check_generated_files () {
+    # For generate_mldsa_tests.py, which is run
+    # from check_committed_generated_files.py
+    python3 -c 'import dilithium_py' 2>/dev/null || return 1
+}
+
```
mpg: Minor: should we split `check_committed_generated_files.py` into its own component? It has stronger requirements than the rest of the component, and as discussed elsewhere we might want to run that component on a different schedule (like only in the nightlies in the short term, and in the longer term only in the tf-psa-crypto CI proper, as opposed to the framework CI running (part of) the tf-psa-crypto CI).

tests/suites/test_suite_pqcp_mldsa.function
```diff
@@ -60,3 +67,95 @@ exit:
...
+}
+/* END_CASE */
+
+/* BEGIN_CASE depends_on:TF_PSA_CRYPTO_PQCP_MLDSA_87_ENABLED */
+void sign_deterministic_pure_87(data_t *private_key,
+                                data_t *message,
+                                data_t *expected_signature)
+{
+    uint8_t *signature = NULL;
+    size_t signature_size = MLDSA_SECRETKEYBYTES(87);
```
mpg: Shouldn't this be `MLDSA_BYTES(87)` instead?

(Btw I'm not a big fan of the name of this macro, I'd have preferred something more explicit like `MLDSA_SIGNATURE_BYTES()`, cause here it's like the size of what? Well not secret key or public key cause they have their own macro, so by elimination it must be signatures. And "by elimination" should never be a step in inferring the meaning of something from its name. But anyway, that's vastly out of scope here.)

tests/suites/test_suite_pqcp_mldsa.function
```diff
@@ -60,3 +67,95 @@ exit:
...
+    /* Calculate and check the deterministic signature */
+    size_t signature_length = SIZE_MAX;
+    TEST_EQUAL(tf_psa_crypto_pqcp_mldsa87_signature_internal(signature,
+                                                             &signature_length,
+                                                             message->x, message->len,
+                                                             prefix, prefix_length,
+                                                             rnd,
+                                                             private_key->x,
+                                                             0), 0);
+    TEST_EQUAL(signature_length, MLDSA_BYTES(87));
```
mpg, line 129: That seems a bit redundant with the length checks done by `TEST_MEMORY_COMPARE` below, and the sanity checks done on the test data above.

tests/suites/test_suite_pqcp_mldsa.function
```diff
@@ -60,3 +67,95 @@ exit:
...
+                                data_t *message,
+                                data_t *expected_signature)
+{
+    uint8_t *signature = NULL;
+    size_t signature_size = MLDSA_SECRETKEYBYTES(87);
+    const uint8_t prefix[2] = { 0, 0 }; // pure ML-DSA with empty context
+    const size_t prefix_length = sizeof(prefix);
+    const uint8_t rnd[MLDSA_RNDBYTES] = { 0 };
+
+    TEST_CALLOC(signature, signature_size);
```
mpg, line 114: Actually, not even sure why we have `signature_size` instead of directly doing `TEST_CALLOC(signature, expected_signature->len)`.

tests/suites/test_suite_pqcp_mldsa.function
```diff
@@ -60,3 +67,95 @@ exit:
...
+    TEST_EQUAL(public_key->len, MLDSA_PUBLICKEYBYTES(87));
+    TEST_EQUAL(signature->len, MLDSA_BYTES(87));
+
+    /* Verify the signature */
+    TEST_EQUAL(tf_psa_crypto_pqcp_mldsa87_verify(signature->x, signature->len,
+                                                 message->x, message->len,
+                                                 NULL, 0,
+                                                 public_key->x), 0);
+
+    /* Verify that a one-bit-off signature is wrong */
```
mpg, line 157: Question: are we happy with just modifying the last byte, or should we also try modifying the first byte, or even have a loop where we corrupt each byte in turn?

From a very "black box" perspective, it seems conceivable that the verification function could fail to validate some of the bytes in the input...

tests/scripts/check_committed_generated_files.py
```diff
@@ -13,6 +13,7 @@
 
 GENERATORS = [
     config_macros.Current(shadow_missing_ok=True),
+    generate_files_helper.TestDataGenerator('framework/util/generate_mldsa_tests.py')
```
mpg: Very minor: add a coma at the end?

mpg (commented):
Looking pretty good to me, but a some minor and a few less minor points.

I have been able to reproduce the generated file locally with Python 3.12 (default in Ubuntu 24.04) in a fresh venv after running only `scripts/min_requirements.py`.

tests/suites/host_test.function
gilles-peskine-arm: `offset` will be used in the next loop iteration if the line is empty or a comment.

(This may not be the cleanest structure for a read-line-skipping-comments function, but it works and I think it's good enough for test code.)

tests/suites/host_test.function
gilles-peskine-arm: This code parses `.datax` files, not `.data` files. (It derives from code that parsed `.data` files, a long time ago.) `.data` files probably didn't have non-empty whitespace-only lines, and `.datax` files definitely don't. So I didn't try to preserve the same whitespace behavior.

tests/suites/test_suite_pqcp_mldsa.function
gilles-peskine-arm: I agree about the names. We might later add a similar macro to our public headers (as an unofficial type-specific variants of a standard macro, used to define the standard macro), then we can use it internally as well.

tests/suites/test_suite_pqcp_mldsa.function
gilles-peskine-arm: A habit I have to give a name to output buffer sizes, so that it's clear that what we pass as the output size parameter to a library function is what we've allocated. It doesn't matter here, but I don't think it hurts. It matters more in places like [the driver entry point tests](https://github.com/gilles-peskine-arm/TF-PSA-Crypto/blob/4cb5e123ee7c821c0155c385c1ce851235115ddc/tests/suites/test_suite_psa_crypto_mldsa.function) where the function also tries a larger buffer and a smaller buffer.

tests/suites/test_suite_pqcp_mldsa.function
gilles-peskine-arm: This test suite doesn't aim to validate an ML-DSA implementation, it only aims to validate that we're calling the ML-DSA implementation correctly. So:

* We test both a valid signature and an invalid signature. This validates that we're reading the is-it-valid boolean result correctly.
* We test a signature where only the last byte is invalid. This validates that what we think of as the signature length is probably correct: if we weren't passing enough data then the verification function would do a buffer overread, and if we were passing too much data then the excess data at the end would be ignored.

ML-DSA signatures have some structure, so a decent test coverage would include a lot of negative cases related to that structure. But that's the job of the mldsa-native unit tests, not our job.

tests/suites/host_test.function
mpg: Ah right. I had noted the funny loop structure but failed to connect the dots here. As you say, good enough for test code.

tests/suites/test_suite_pqcp_mldsa.function
mpg: Regardless of the side point about the name, I still think `signature_size = MLDSA_SECRETKEYBYTES(87);` looks wrong here. (Just to made sure the main point isn't lost due to the side point.)

tests/suites/test_suite_pqcp_mldsa.function
mpg: Btw should we also test passing a signature that's one byte too short?

tests/suites/test_suite_pqcp_mldsa.function
gilles-peskine-arm: There's overlap, but I see them as separate contracts: the `signature_length` output is the expected value (e.g. we aren't accidentally calling the 44 implementation when we expected 87), and the `{signature, signature_length}` buffer has the expected content (which does include its length, but that's kind of a side effect).

tests/suites/test_suite_pqcp_mldsa.function
gilles-peskine-arm: > Btw should we also test passing a signature that's one byte too short?

I currently have this check in the driver, so my code doesn't rely on the check done in mldsa-native. But in case we later do rely on the check in mldsa-native, I agree, it would be a good idea to validate that it's present.

valeriosetti (dismissed):

mpg (dismissed):
LGTM, thanks!

valeriosetti (dismissed):

valeriosetti:
Unfortunately there is a conflict on the `framework` repo that needs to be resolved

gilles-peskine-arm:
I've done the obligatory rebase on top of `development` to resolve the framework conflict. The whole framework update is now in the first commit, otherwise the history remains the same.

valeriosetti (approved):

mpg (approved):
Still LGTM after rebase.

PR #694, opened 2026-02-18
Simplify platform requirements before 1.1

### LINKED ISSUE #675: Simplify automatic support of Unix-like platforms across modules
There are several places in TF-PSA-Crypto and Mbed TLS that have an adherence on non-standard C features, either mandatory or optional. We have built-in support for Unix-like platforms and for Windows, and sometimes ways to configure the library on other platforms. “Unix-like” covers a vast range of systems, including:

* Linux kernel with several possible libc.
* Unix implementations from a single vendor such as \*BSD, macOS, QNX, AIX, …
* Various Unix/POSIX/Linux layers on Windows.
* Various embedded systems with partial Unix APIs.

The goal of this issue is to simplify the code that we use to detect, and adapt to, Unix-like platforms, so that we are less afraid of breaking things if we need to fix a bug in the next long-time support branch. It covers both TF-PSA-Crypto and Mbed TLS.

Here are some known problems, as of TF-PSA-Crypto 1.0.0 and Mbed TLS 4.0.0:

* We include `unistd.h` and other Unix headers in many different places: [`platform_util.c`](https://github.com/Mbed-TLS/TF-PSA-Crypto/blob/tf-psa-crypto-1.0.0/drivers/builtin/src/platform_util.c#L149), [`threading.c`](https://github.com/Mbed-TLS/TF-PSA-Crypto/blob/tf-psa-crypto-1.0.0/drivers/builtin/src/threading.c#L26), [`unistd.c`](https://github.com/Mbed-TLS/mbedtls/blob/mbedtls-4.0.0/library/timing.c#L16), [`net_sockets.c`](https://github.com/Mbed-TLS/mbedtls/blob/mbedtls-4.0.0/library/net_sockets.c#L22) (and that's just for the library, the rest matters less). Each file has a different idea of whether to include QNX, Haiku, Midipix…
* Compiler options can break the build as they may influence which symbols are declared in system headers.
* We have hacks to allow a non-C99 `printf` [in `debug.h`](https://github.com/Mbed-TLS/mbedtls/blob/mbedtls-4.0.0/include/mbedtls/debug.h#L90), only for Windows. These hacks may or may not work correctly because the `printf` variant (standard C99 or Windows legacy) is determined when `debug.c` is built, but the format strings is determined when `ssl_*.c` is built. They may end up being different if, for example, a module using debug also defines a macro such as `_POSIX_C_SOURCE` or `_XOPEN_SOURCE` — then Mingw will pull in the GNU printf implementation which is C99-compliant, instead of the default one which isn't. The way Mingw picks the printf implementation seems to have been overhauled in version 13, released in June 2025, and our code may not work with that version (I suspect that's the cause of https://github.com/Mbed-TLS/mbedtls/issues/10566). We've had other problems with these hacks before (https://github.com/Mbed-TLS/mbedtls/issues/10017).

Out of scope:

* Providing additional configurability is out of scope here. It is of course desirable, but it would take a lot of effort. This issue focusing on what works out of the box.
* Only the libraries is in scope. While it would be nice for the sample programs to run on more platforms, it's a lot less important. Test tools are completely out of scope, we only care about them running on our CI environment and on developer machines.

gilles-peskine-arm:
Resolve https://github.com/Mbed-TLS/TF-PSA-Crypto/issues/675

* Rewrite `MBEDTLS_STATIC_ASSERT` and make it actually assert everywhere. This may work on extremely exotic C99-but-not-C11 compilers or compiler warnings.
* Move feature requirement macros that need to be defined before including any system header to a new dedicated header `core/tf_psa_crypto_platform_requirements.h`.
* Redo feature requirement macros from scratch, so that they're simpler and justified. This could cause “exotic” platforms to break.
* Always use C99-compliant printf on MingW, removing the need for `MBEDTLS_PRINTF_SIZET` in `debug.h`.
* Remove the use of `gettimeofday()` (we use `clock_gettime()` for `mbedtls_ms_time()`.

Needs preceding PR: 
* [x] [Stop using `gettimeofday()` in Mbed TLS](https://github.com/Mbed-TLS/mbedtls/pull/10607) (needed to fix the build on FreeBSD)
* [x] framework https://github.com/Mbed-TLS/mbedtls-framework/pull/279
* [x] Prepare mbedtls for the MingW change: https://github.com/Mbed-TLS/mbedtls/pull/10625

## PR checklist

- [x] **changelog** provided
- [x] **development PR** provided https://github.com/Mbed-TLS/mbedtls/pull/10606
- [x] **TF-PSA-Crypto PR** provided https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/694
- [x] **framework PR** provided https://github.com/Mbed-TLS/mbedtls-framework/pull/279
- [x] **3.6 PR** https://github.com/Mbed-TLS/mbedtls/pull/10621 (only a small part that isn't disruptive)
- **tests**  provided + local testing (see https://github.com/Mbed-TLS/mbedtls/pull/10606)

### DIFF
--- a/core/tf_psa_crypto_common.h
+++ b/core/tf_psa_crypto_common.h
@@ -1,7 +1,18 @@
 /**
  * \file tf_psa_crypto_common.h
  *
- * \brief Utility macros for internal use in the library
+ * \brief Utility macros for internal use in the library.
+ *
+ * This file should be included as the first thing in all library C files.
+ * It must not be included by sample programs, since sample programs
+ * illustrate what you can do without the library sources.
+ * It may be included (often indirectly) by test code that isn't purely
+ * black-box testing.
+ *
+ * This file takes care of setting up requirements for platform headers.
+ * It includes the library configuration and derived macros.
+ * It additionally defines various utility macros and other definitions
+ * (but no function declarations).
  */
 /*
  *  Copyright The Mbed TLS Contributors
@@ -15,7 +26,10 @@
  * headers what we expect of them. */
 #include "tf_psa_crypto_platform_requirements.h"
 
+/* From this point onwards, ensure we have the library configuration and
+ * the configuration-derived macros. */
 #include "tf-psa-crypto/build_info.h"
+
 #include "alignment.h"
 
 #include <assert.h>
--- a/core/tf_psa_crypto_common.h
+++ b/core/tf_psa_crypto_common.h
@@ -382,30 +382,121 @@ static inline void mbedtls_xor_no_simd(unsigned char *r,
 #endif
 #endif
 
-/* Always provide a static assert macro, so it can be used unconditionally.
- * It does nothing on systems where we don't know how to define a static assert.
+/** \def MBEDTLS_STATIC_ASSERT
+ *
+ * A static assert macro, equivalent to `static_assert` or `_Static_assert`
+ * in modern C.
+ *
+ * You can use `MBEDTLS_STATIC_ASSERT(expr, msg)` in any position where a
+ * declaration is permitted, both at the toplevel and within a function.
+ * This macro may not be used inside an expression (see #STATIC_ASSERT_EXPR,
+ * available on fewer platforms).
+ *
+ * \param expr  An expression which must be a compile-time constant with
+ *              an integer value. This doesn't have to be a preprocessor
+ *              constant, for example it can use `sizeof`.
+ *              The compilation fails if the value is 0.
+ * \param msg   An error messsage to display if the value of \p expr is 0.
  */
-/* Can't use the C11-style `defined(static_assert)` on FreeBSD, since it
- * defines static_assert even with -std=c99, but then complains about it.
- */
-#if defined(static_assert) && !defined(__FreeBSD__)
+#if __STDC_VERSION__ >= 202311L
+/* static_assert is a keyword since C23 */
 #define MBEDTLS_STATIC_ASSERT(expr, msg)    static_assert(expr, msg)
-/* The GCC compiler supports _Static_assert since version 4.6 even with C99 so checking the
- * compiler version should suffice.
- * For non GCC compilers we check that C>=11 is used since C11 introduced _Static_assert.
- */
-#define MBEDTLS_STATIC_ASSERT_SUPPORT
-#elif !defined(__cplusplus) && \
-    ((defined(__GNUC__) && ((__GNUC__ == 4 && __GNUC_MINOR >= 6) || __GNUC__ > 4)) || \
-    (defined(__STDC_VERSION__) && __STDC_VERSION__ >= 201112L))
+
+#elif __STDC_VERSION__ >= 201112L
+/* _Static_assert is a keyword since C11 */
 #define MBEDTLS_STATIC_ASSERT(expr, msg)    _Static_assert(expr, msg)
-#define MBEDTLS_STATIC_ASSERT_SUPPORT
+
+#elif defined(static_assert) && !defined(__STRICT_ANSI__)
+/* If static_assert is defined as a macro, presumably from <assert.h>
+ * included above, then trust that it is what we expect.
+ * This is a common extension even before C11.
+ * However, don't use it if it looks like a build with `gcc -c99 -pedantic`
+ * or `clang -c99 -pedantic`, because they would complain about the use of
+ * a feature that doesn't exist in C99.
+ */
+#define MBEDTLS_STATIC_ASSERT(expr, msg)    static_assert(expr, msg)
+
+#elif defined(_MSC_VER)
+/* MSVC has `static_assert` as a keyword (not a macro) since
+ * Visual Studio 2010.
+ */
+#define MBEDTLS_STATIC_ASSERT(expr, msg)    static_assert(expr, msg)
+
+#elif defined(__GNUC__) && \
+    ((__GNUC__ == 4 && __GNUC_MINOR >= 6) || __GNUC__ > 4) && \
+    !defined(__STRICT_ANSI__)
+/* _Static_assert is a keyword since GCC 4.6.
+ * However, don't use it if it looks like a build with `gcc -c99 -pedantic`
+ * or `clang -c99 -pedantic`, because they would complain about the use of
+ * a feature that doesn't exist in C99.
+ */
+#define MBEDTLS_STATIC_ASSERT(expr, msg)    _Static_assert(expr, msg)
+
+#elif defined(__COUNTER__)
+/* Fall back to a hack that works in practice with non-ancient GCC-like
+ * compilers and MSVC, and doesn't trigger `-Wredundant-decls`.
+ *
+ * See the `#else` block below for an explanation. Here, we add another
+ * layer to make the declared name unique using the special preprocessor
+ * token `__COUNTER__`.
+ */
+#define MBEDTLS_STATIC_ASSERT_COUNTER(expr, counter) \
+    extern void mbedtls_static_assert_anchor##counter(char[1 - 2 * !(expr)])
+#define MBEDTLS_STATIC_ASSERT_WRAP(expr, counter) \
+    MBEDTLS_STATIC_ASSERT_COUNTER(expr, counter)
+#define MBEDTLS_STATIC_ASSERT(expr, msg) \
+    MBEDTLS_STATIC_ASSERT_WRAP(expr, __COUNTER__)
+
 #else
-/* Make sure `MBEDTLS_STATIC_ASSERT(expr, msg);` is valid both inside and
- * outside a function. We choose a struct declaration, which can be repeated
- * any number of times and does not need a matching definition. */
+/* Fall back to a hack that works in practice with almost all C compilers.
+ *
+ * Constraints:
+ * - Must be valid C99 when `expr` is a constant expression with a nonzero value.
+ * - Must compile without warnings on known compilers when `expr` is a
+ *   constant expression with a nonzero value.
+ * - Must be valid both at file scope and inside a function.
+ * - Must allow multiple static assertions in the same scope.
+ * - Must not rely on `__LINE__` to create unique identifiers, since this
+ *   could lead to collisions, e.g. if `MBEDTLS_STATIC_ASSERT` is used in
+ *   a header, or if a macro expands to multiple uses of
+ *   `MBEDTLS_STATIC_ASSERT`.
+ * - Should result in an error when `expr` evaluates to 0.
+ *
+ * How it works:
+ * - Ostensibly declare a function. This function will never be used, but
+ *   declaring a function that won't be used is routine.
+ * - The function's name is in our namespace, so we just need to avoid that
+ *   name for any other purpose.
+ * - Declaring the same function with the same prototype multiple times is
+ *   also common (it triggers `gcc -Wredundant-decls`, but we handle
+ *   non-ancient GCC separately above).
+ * - The function takes an argument whose type involves an array.
+ * - The array size is 1 (valid) when the expression is true, and -1
+ *   (invalid, triggers an error in almost all compilers) when the expression
+ *   is false.
+ *
+ * Limitations:
+ * - If you have multiple static assertions in the same scope,
+ *   `gcc -Wredundant-decls` complains.
+ * - Technically, an array of length -1 doesn't have to lead to a compilation
+ *   error in C99. In C89, it does: it's a constraint violation. But in C99,
+ *   it could be a variable-length array.
+ * - When the assertion fails, some compilers complain about a negative
+ *   array length without displaying the problematic line, so the message
+ *   is not visible.
+ *
+ * On Godbolt compiler explorer, the only failures I could find are:
+ * - 6502 cc65 complains if there are multiple assertions in the same scope:
+ *   "Multiple definition for `mbedtls_static_assert_anchor'"
+ * - Chibicc 2020-12-07 ignores the assertion.
+ * - LC3 (trunk) ignores the assertion.
+ * - ppci 0.5.5 dies with a NotImplementedError on the `!` operator.
+ * - SDCC 4.0.0 through 4.5.0 complains if there are multiple assertions in
+ *   the same scope:
+ *   "extern definition for 'mbedtls_static_assert_anchor' mismatches with declaration."
+ */
 #define MBEDTLS_STATIC_ASSERT(expr, msg)                                \
-    struct ISO_C_does_not_allow_extra_semicolon_outside_of_a_function
+    extern void mbedtls_static_assert_anchor(char[(expr) ? 1 : -1])
 #endif
 
 /* Define compiler branch hints */
--- a/core/tf_psa_crypto_common.h
+++ b/core/tf_psa_crypto_common.h
@@ -445,7 +445,7 @@ static inline void mbedtls_xor_no_simd(unsigned char *r,
 #define MBEDTLS_STATIC_ASSERT(expr, msg)    static_assert(expr, msg)
 
 #elif defined(__GNUC__) && \
-    ((__GNUC__ == 4 && __GNUC_MINOR >= 6) || __GNUC__ > 4) && \
+    ((__GNUC__ == 4 && __GNUC_MINOR__ >= 6) || __GNUC__ > 4) && \
     !defined(__STRICT_ANSI__)
 /* _Static_assert is a keyword since GCC 4.6.
  * However, don't use it if it looks like a build with `gcc -c99 -pedantic`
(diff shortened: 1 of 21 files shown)


core/tf_psa_crypto_common.h
```diff
@@ -382,30 +382,121 @@ static inline void mbedtls_xor_no_simd(unsigned char *r,
...
- * For non GCC compilers we check that C>=11 is used since C11 introduced _Static_assert.
+
+#elif defined(_MSC_VER)
+/* MSVC has `static_assert` as a keyword (not a macro) since
+ * Visual Studio 2010.
+ */
+#define MBEDTLS_STATIC_ASSERT(expr, msg)    static_assert(expr, msg)
+
+#elif defined(__GNUC__) && \
+    ((__GNUC__ == 4 && __GNUC_MINOR >= 6) || __GNUC__ > 4) && \
```
bensze01: Preexisting typo
```suggestion
    ((__GNUC__ == 4 && __GNUC_MINOR__ >= 6) || __GNUC__ > 4) && \
```

bensze01 (changes requested):

core/tf_psa_crypto_platform_requirements.h
```diff
@@ -0,0 +1,62 @@
...
+ * where snprintf does not always zero-terminate the buffer, and does
+ * not support formats such as "%zu" for size_t and "%lld" for long long.
+ *
+ * Defining __USE_MINGW_ANSI_STDIO=0 may work and provide a small code size
+ * and performance benefit for some combinations of older Mingw and Windows
+ * versions. Do this at your own risk and make sure that least
+ * test_suite_platform_printf passes.
+ */
+#if !defined(__USE_MINGW_ANSI_STDIO)
+#define __USE_MINGW_ANSI_STDIO 1
```
gilles-peskine-arm, line 59: I had missed that we need an extra step to avoid [breaking this branch of TF-PSA-Crypto with the current development branch of mbedtls](https://ci.trustedfirmware.org/job/mbed-tls-tf-psa-crypto-multibranch/job/PR-694-head/8/pipeline-console/log?nodeId=3388).

`mbedtls_common.h` includes `tf_psa_crypto_common.h` which now includes `tf_psa_crypto_platform_requirements.h‎` which defines `__USE_MINGW_ANSI_STDIO` causes `debug.c` to use a standard printf family. (That's in the default config. When `MBEDTLS_PLATFORM_VSNPRINTF_ALT` is enabled, what would matter is the printf family when `platform.c` is built.) This breaks the build in mbedtls since it's still assuming a legacy printf and setting `MBEDTLS_PRINTF_xxx` macros accordingly.

So I think that before merging this PR, we need to update mbedtls with the part of https://github.com/Mbed-TLS/mbedtls/pull/10606 that updates the MingW printf selection.

gilles-peskine-arm: > When `MBEDTLS_PLATFORM_VSNPRINTF_ALT` is enabled, what would matter is the printf family when platform.c is built.

Ah, but we automatically enable `MBEDTLS_PLATFORM_VSNPRINTF_ALT` (and `MBEDTLS_PLATFORM_SNPRINTF_ALT` but not the others) on MingW (and older, unsupported MSVC). Things can never be simple, can they?

gilles-peskine-arm: https://github.com/Mbed-TLS/mbedtls/pull/10625

core/tf_psa_crypto_common.h
```diff
@@ -382,30 +382,121 @@ static inline void mbedtls_xor_no_simd(unsigned char *r,
...
+ * - Chibicc 2020-12-07 ignores the assertion.
+ * - LC3 (trunk) ignores the assertion.
+ * - ppci 0.5.5 dies with a NotImplementedError on the `!` operator.
+ * - SDCC 4.0.0 through 4.5.0 complains if there are multiple assertions in
+ *   the same scope:
+ *   "extern definition for 'mbedtls_static_assert_anchor' mismatches with declaration."
+ */
 #define MBEDTLS_STATIC_ASSERT(expr, msg)                                \
-    struct ISO_C_does_not_allow_extra_semicolon_outside_of_a_function
+    extern void mbedtls_static_assert_anchor(char[(expr) ? 1 : -1])
```
bensze01: This expression doesn't assert if `expr` is not an integer-constant expression - eg. when running `gcc` with `-std=c99`. See:
https://godbolt.org/z/jKz48nTPh

Instead, we should use a statement derived from the current version of `STATIC_ASSERT_EXPR`, which is also fully valid C99:
https://github.com/Mbed-TLS/TF-PSA-Crypto/blob/20f2ca9dd2ffd1516fa62b4c0eff32a292048974/core/tf_psa_crypto_common.h#L82-L85

```suggestion
    struct mbedtls_static_assert_anchor { int MBEDTLS_STATIC_ASSERT : (expr) ? 1 : -1; }
```

core/tf_psa_crypto_common.h
```diff
@@ -382,30 +382,121 @@ static inline void mbedtls_xor_no_simd(unsigned char *r,
...
+#elif defined(__COUNTER__)
+/* Fall back to a hack that works in practice with non-ancient GCC-like
+ * compilers and MSVC, and doesn't trigger `-Wredundant-decls`.
+ *
+ * See the `#else` block below for an explanation. Here, we add another
+ * layer to make the declared name unique using the special preprocessor
+ * token `__COUNTER__`.
+ */
+#define MBEDTLS_STATIC_ASSERT_COUNTER(expr, counter) \
+    extern void mbedtls_static_assert_anchor##counter(char[1 - 2 * !(expr)])
```
bensze01: The integer constant expression here doesn't match the non-counter version below.

I assume the intent was to standardize on either  the arithmetic or the ternary expression - the ternary seems to be easier to parse for users when presented as an error message, so I'd prefer that one if there are no compatibility implications (both should be fully C99 compliant).

Also see: https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/694#discussion_r2865507542

```suggestion
    struct mbedtls_static_assert_anchor##counter { int MBEDTLS_STATIC_ASSERT : (expr) ? 1 : -1; }
```

tests/suites/test_suite_platform_unix.function
```diff
@@ -0,0 +1,45 @@
...
+/* END_CASE */
+
+/* BEGIN_CASE depends_on:MBEDTLS_PLATFORM_IS_UNIXLIKE */
+void clock_gettime_available(int clockid)
+{
+    struct timespec ts = { 0, 0 };
+    memset(&ts, 0, sizeof(ts));
+    int ret = clock_gettime(clockid, &ts);
+    TEST_ASSERT_ERRNO(ret == 0);
+    TEST_LE_S(1, ts.tv_sec);
```
bensze01: This will fail if the system clock is set to a value <= 1970-01-01T00:00:00 if `time_t` is signed :)

I think we can just remove this check, since we don't care about a functional realtime clock.

If we care about the clock being functional, we could test that `(ts.tv_sec | ts.tv_nsec) != 0`, which should cut down the probability to near zero, or assert that a second call to `clock_gettime` (after a short sleep)  doesn't return the same values.

tests/suites/test_suite_platform_unix.function
```diff
@@ -17,8 +17,12 @@
...
 
 /* Constants used in the test data need to be defined even if no test
- * functions that use them are enabled. */
+ * functions that use them are enabled.
+ * Undefine the macros first in case a system header does define them
+ * even though we haven't recognized the platform as Unix-like. */
+#undef CLOCK_REALTIME
 #define CLOCK_REALTIME 0
+#undef CLOCK_MONOTONIC
 #define CLOCK_MONOTONIC 0
```
bensze01, line 26: Redefining system macros feels like it might bite us in the future.

Why not just wrap the definitions with `#if !defined(FOO)` ?

core/tf_psa_crypto_common.h
```diff
@@ -104,7 +104,7 @@ extern void (*mbedtls_test_hook_test_fail)(const char *test, int line, const cha
 /* A compile-time constant with the value 0. If `const_expr` is not a
  * compile-time constant with a nonzero value, cause a compile-time error. */
 #define STATIC_ASSERT_EXPR(const_expr)                                \
-    (0 && sizeof(struct { unsigned int STATIC_ASSERT : 1 - 2 * !(const_expr); }))
+    (0 && sizeof(char[1 - 2 * !(const_expr)]))
```
bensze01: What version of GCC complains about anonymous structs?

I could only find a commit message (c86a16548c82c345d2784d6b79eecfcef71ce08a) that claimed MSVC was unhappy - if GCC is happy with the original definition, then in light of https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/694#discussion_r2865507542 I don't see any clear benefit from this patch.

On a second note, my suggestion in https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/694#discussion_r2865507542 doesn't use anonymous structs, so MSVC should be happy as well - either that approach, or a [__pragma](https://learn.microsoft.com/en-us/cpp/preprocessor/pragma-directives-and-the-pragma-keyword?view=msvc-170#the-pragma-keyword) could be used to silence the specific warning on MSVC.

bensze01 (changes requested):

core/tf_psa_crypto_common.h
gilles-peskine-arm: I couldn't get a struct to work when `STATIC_ASSERT_EXPR` is used inside of `ARRAY_LENGTH` which is used in `MBEDTLS_STATIC_ASSERT` (which we do in our code base, that's how I noticed), when `MBEDTLS_STATIC_ASSERT` is implemented in a way that doesn't rely on anything but C99. See the commit message for d5e30685a045bdf30c210d1d1d47151452207fe6.

If the struct is named then you can't have two assertions in the same scope. If the struct is not named then GCC complains that an anonymous struct in a “small” scope is silly.

core/tf_psa_crypto_common.h
gilles-peskine-arm: Same as https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/694#discussion_r2867415042 : I ddn't manage to get a struct to work. With your proposal, you can't have two assertions in the same scope.

core/tf_psa_crypto_common.h
gilles-peskine-arm: I'll need to go back and check, but IIRC on Godbolt I found a compiler that treated `array[(0) ? 1 : -1]` as a VLA but produced a compilation error with `array[1 - 2 * (0)]`.

tests/suites/test_suite_platform_unix.function
gilles-peskine-arm: I don't want to check that `clock_gettime()` is functional, but I would like to check that it's actually returning something that isn't all-bits-zero.

Indeed calling the function twice with a delay in between would be a better test, but I also don't want to make assumptions on delays. That tends to play badly in the CI.

I think just checking that the return value isn't all-bits-zero is indeed better. Not because I'm worried about the epoch, but because `clock_gettime` can have different clocks, including one that starts at boot, and running the test less than 1 second after boot is plausible.

tests/suites/test_suite_platform_unix.function
gilles-peskine-arm: This is only in the case where we believe that the platform isn't Unix-like, but system headers have defined these macros anyway. The definitions provided by system headers might not work either.

This could be solved with a level of indirection (use `MBEDTLS_TEST_CLOCK_REALTIME` in test data and define it differently depending on `MBEDTLS_PLATFORM_IS_UNIXLIKE`), but I don't think this warrants the additional complexity. These macro redefinitions are only done in a case where there's very little code in the `.c` file after them (since none of the test functions are compiled in).

core/tf_psa_crypto_common.h
yanesca: Based on your [previous comment](https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/694/changes/9ebf182ca45680137027ff93c95812b1665d9996#r2867417298
) some compilers would treat this as a VLA and wouldn't fail. Is this not a problem?

(This is just a random question, I am not doing a full review.)

valeriosetti (dismissed):

core/tf_psa_crypto_common.h
bensze01: I see - that's annoying. How would you feel about keeping the struct-based approach I suggested if `__COUNTER__` is available, and going with the array-length-based approach if it is not available?

core/tf_psa_crypto_common.h
bensze01: I think I managed to combine our approaches in a way that avoids false-negatives via misinterpretation as a VLA, and is accepted by old versions of GCC without warnings:

```suggestion
    extern void mbedtls_static_assert_anchor(char[sizeof(struct {int MBEDTLS_STATIC_ASSERT : 1 - 2 * !(expr);} )])
```

bensze01: ~~Whelp - it works in GCC, but ICC just ignores the violation inside the `sizeof` :(~~

bensze01: Nope, ICC handles this correctly, I was just testing it wrong.

gilles-peskine-arm: That doesn't work, see the commit message for d5e30685a045bdf30c210d1d1d47151452207fe6.

However, I may have been too focused on declaring a function.
```
extern int mbedtls_static_assert_anchor[sizeof(struct {int x : (expr) ? 1 : -1;})]
```
is looking promising so far. I still need to check more compilers.

gilles-peskine-arm: No, I did have a reason for declaring a function: if it's not a function, then in function scope, this is likely to cause an unused-variable warning. I can pretty much universally silence that warning with `(void) mbedtls_static_assert_anchor`, but that only works in function scope, not at file scope.

Maybe the right way to do this would be to have different macros, one for file scope and one for function scope?

I'm going to try a little more, but at some point I propose that we merge what we have, and improve it later. Before this pull request, `MBEDTLS_STATIC_ASSERT` was a no-op on compilers that were neither C11 nor GCC-like. So it's still an improvement over the existing situation if we now cover most common compilers that are neither C11 nor GCC-like.

----

By the way

> This expression doesn't assert if expr is not an integer-constant expression

is not a critical problem. We run builds where `MBEDTLS_STATIC_ASSERT` is a C11 static assert, and that will fail the build if `expr` is not a constant expression. The only way a problem would be unnoticed is if `expr` involes some user configuration or platform-dependent macro, and that macro is not a constant expression in some real-world usage, but it is a constant expression in all of our CI builds, and that macro is not used in any other place where a constant expression is required. It's better if we detect this, but it's fairly unlikely to happen.

gilles-peskine-arm: A function _returning_ a pointer to an array might avoid both pitfalls: compilers don't usually complain about an unused function declaration even at function scope, and they don't seem to complain about an anonymous struct inside an array length.

```
#define MBEDTLS_STATIC_ASSERT(expr, msg)                                \
    extern int (*mbedtls_static_assert_anchor(void))[sizeof(struct {int x : (expr) ? 1 : -1;})]
```

This is actually something I found online, but IIRC I had a reason against it as well. I don't recall what that reason was though, so I'm going to try it out some more.

gilles-peskine-arm: The only serious problem I can find with a function returning a pointer to an array whose size involves a struct with a potentially negative bit-field width (other than its complexity) is that MSVC doesn't like it. But fortunately we have a different implementation with (non-ancient) MSVC.

core/tf_psa_crypto_common.h
gilles-peskine-arm: At least up to version 15.1.0, GCC doesn't seem to complain about an anonymous struct inside sizeof inside an array that is not in a function argument list. So I've switched to that.

core/tf_psa_crypto_common.h
gilles-peskine-arm: Moot since I've now moved everything to a bit-field width.

tests/suites/test_suite_platform_unix.function
gilles-peskine-arm: I don't intend to change this. Neither of  the proposed simple solutions is 100% correct, and both are highly unlikely to fail in practice, so I think they're equally fine.

bensze01 (dismissed):
Thanks, LGTM - needs a rebase due to a submodule conflict.

gilles-peskine-arm:
I've rebased on top of `development`. The framework is already up to date there, so the framework updates have been dropped. No other change.

valeriosetti (approved):

bensze01 (approved):

PR #696, opened 2026-02-23
Repository reorganization

### LINKED ISSUE #597: Move out of `drivers/builtin/src/` non driver modules
### Suggested enhancement
The `drivers/builtin/src/` directory contains C modules that are not part of the built-in driver. The list of such files at time of writing is:
`asn1parse.c` , `asn1write.c`                                                                     
`base64.c`, `base64_internal.h`                                                                                                                                   
`lmots.c`, `lmots.h`, `lms.c`                                                                        
`md.c`, `md_psa.h`, `md_wrap.h`                                                                       
`memory_buffer_alloc.c`                                                           
`nist_kw.c`                                                                       
`oid.c`, `crypto_oid.h`                                                                           
`pem.c`                                                                           
`pk.c`, `pk_ecc.c`, `pk_internal.h`, `pkparse.c`, `pk_rsa.c`, `pk_wrap.c`, `pk_wrap.h`, `pkwrite.c`, `pkwrite.h`
`pkcs5.c` 
`platform.c`, `platform_util.c`                                                   
`threading.c`, `threading_internal.h`

The goal of this issue is to move them out of `drivers/builtin/src` and adapt accordingly the build systems (Mbed TLS legacy make included) and our various scripts.

The files should be moved to `core` that we rename `library` as it is going to contain more that just the PSA core code and to two new directories at repo root: `platform/` and `utilities/`. The platform code is moved under `platform/`, the module that call directly or indirectly the driver code are moved to `library/`, the other are moved to `utilities/`.

In `core/` renamed to `library/` we move:
`lmots.c`, `lmots.h`, `lms.c`                                                                        
`md.c`, `md_psa.h`, `md_wrap.h`                                                                                                                                  
`nist_kw.c`                                                                                                                                                                                                                             
`pk.c`, `pk_ecc.c`, `pk_internal.h`, `pkparse.c`, `pk_rsa.c`, `pk_wrap.c`, `pk_wrap.h`, `pkwrite.c`, `pkwrite.h`

In `platform/` we move:
`memory_buffer_alloc.c`
`platform.c`, `platform_util.c`                                                   
`threading.c`, `threading_internal.h`

In `utilities/` we move:
`asn1parse.c` , `asn1write.c`                                                                     
`base64.c`, `base64_internal.h`
`constant_time.c`, `constant_time_impl.h` and `constant_time_internal.h`                                                                                                                                                                                                                                                                                                                                   
`oid.c`, `crypto_oid.h`                                                                           
`pem.c`                                                                           
`pkcs5.c` 

### Justification
TF-PSA-Crypto needs this because the files mentioned above should not be in `drivers/builtin/src/`.

ronald-cron-arm:
## Description
Introduction of dispatch, extras, platform and utilities directories at the root of TF-PSA-Crypto repository.
Fixes https://github.com/Mbed-TLS/TF-PSA-Crypto/issues/597.

Mbed-TLS/mbedtls-framework#280 and Mbed-TLS/mbedtls#10617 need to be merged first for this PR to have a green CI.

## PR checklist
- [x] **changelog** provided
- [x] **framework PR** provided Mbed-TLS/mbedtls-framework#280
- [x] **mbedtls development PR** provided Mbed-TLS/mbedtls#10617
- [x] **mbedtls 3.6 PR** not required because: 4.x only work
- **tests**  not required because: no need for additional tests


ronald-cron-arm:
CI is red as expected as long as Mbed-TLS/mbedtls#10617 is not merged. But I have validated the PR with a parametrized job: https://ci.trustedfirmware.org/view/Mbed-TLS/job/mbed-tls-restricted-pr-test-parametrized/99/. The PR is thus ready for review.

core/CMakeLists.txt
```diff
@@ -1,15 +1,8 @@
...
-    psa_its_file.c
-    psa_util.c
-    tf_psa_crypto_config.c
-    tf_psa_crypto_version.c
-)
+file(GLOB src_crypto "${CMAKE_CURRENT_SOURCE_DIR}/*.c")
+list(FIND src_crypto "${CMAKE_CURRENT_SOURCE_DIR}/psa_crypto_driver_wrappers_no_static.c" _idx)
+if(_idx EQUAL -1)
+    list(APPEND src_crypto psa_crypto_driver_wrappers_no_static.c)
+endif()
```
valeriosetti: Can you please explain/add a comment for this?
My hypothesis is that in case `psa_crypto_driver_wrappers_no_static.c` is not present at the moment `GLOB` is run, then we make this is a dependency so that it's then generated few lines below. But I might be wrong of course. That's why I think a comment is beneficial in this case

valeriosetti (changes requested):
I only requested a comment but overall it looks good to me

configs/ext/crypto_config_profile_medium.h
```diff
@@ -158,7 +158,7 @@
  * \note This abstraction layer must be enabled on Windows (including MSYS2)
  * as other modules rely on it for a fixed snprintf implementation.
  *
- * Module:  drivers/builtin/src/platform.c
+ * Module:  platform/platform.c
```
minosgalanakis, line 161: Doing a quick grep I can still find [references to the old location](https://github.com/Mbed-TLS/TF-PSA-Crypto/blob/development/tests/scripts/components-configuration-platform.sh#L56) of this file on the tests.

I understand that both locations need to exist in the MbedTLS side for transition, but I fail to see why this was here.

core/CMakeLists.txt
```diff
@@ -1,15 +1,8 @@
...
-    psa_crypto_storage.c
-    psa_its_file.c
-    psa_util.c
-    tf_psa_crypto_config.c
-    tf_psa_crypto_version.c
-)
+file(GLOB src_crypto "${CMAKE_CURRENT_SOURCE_DIR}/*.c")
+list(FIND src_crypto "${CMAKE_CURRENT_SOURCE_DIR}/psa_crypto_driver_wrappers_no_static.c" _idx)
+if(_idx EQUAL -1)
+    list(APPEND src_crypto psa_crypto_driver_wrappers_no_static.c)
```
minosgalanakis: Shouldn't the "${CMAKE_CURRENT_SOURCE_DIR}/psa_crypto_driver_wrappers_no_static.c" be used when the cmake  falls-back here?

platform/CMakeLists.txt
```diff
@@ -0,0 +1,3 @@
+file(GLOB platform_src_files RELATIVE ${CMAKE_CURRENT_SOURCE_DIR} *.c)
```
minosgalanakis, line 1: Given that we probably want to reconfigure the driver files each time we modify the source, should we use `CONFIGURE_DEPENDS` ?

minosgalanakis (commented):
On first pass  I found only some  minor issue with stale code present. The CI failures are normal because it requires the prerequisite PR. 

Will revisit when re-based and the others have been merged

core/CMakeLists.txt
ronald-cron-arm: No because psa_crypto_driver_wrappers_no_static.c could be only in the build tree, not in the source tree.

configs/ext/crypto_config_profile_medium.h
ronald-cron-arm: Good catch, thanks. I've updated the paths to `platform*.o` files in components-configuration-platform.sh.

core/CMakeLists.txt
ronald-cron-arm: I've added a comment and tried to improve the code as well.

platform/CMakeLists.txt
ronald-cron-arm: I am not sure what you mean here. Could you elaborate?

valeriosetti (dismissed):

valeriosetti:
The PR looks good to me, but failures in the CI seems related to the changes being introduced here.

ronald-cron-arm:
> The PR looks good to me, but failures in the CI seems related to the changes being introduced here.

Yes I am on it. Adding a change log as well.

core/CMakeLists.txt
```diff
@@ -1,15 +1,17 @@
...
+# `src_crypto` if present, then append it using a relative path. This ensures
+# that the build system prefers the file from the build tree over the one in
+# the source tree.
+list(FIND src_crypto
+     "${CMAKE_CURRENT_SOURCE_DIR}/psa_crypto_driver_wrappers_no_static.c"
+     _idx)
+if(NOT _idx EQUAL -1)
+    list(REMOVE_AT src_crypto ${_idx})
+endif()
+list(APPEND src_crypto psa_crypto_driver_wrappers_no_static.c)
```
valeriosetti, line 14: If we want to ensure that CMake picks the file from the buildtree sholdn't we use something like `${CMAKE_CURRENT_BINARY_DIR}` in the path?

valeriosetti (commented):
Only 1 minor question, but the rest looks good

core/CMakeLists.txt
ronald-cron-arm: You mean ${CMAKE_CURRENT_BUILD_DIR} I guess. If GEN_FILES=ON that's this the right thing to do. But if GEN_FILES=OFF is should be ${CMAKE_CURRENT_SOURCE_DIR} in general but in all.sh components where GEN_FILES=OFF but the generated files are pre-generated in the build tree. Thus the relative path allows to not implement all this logic and I have documented it now thus I would prefer to stick with the last version if the CI is green now.

valeriosetti (approved):

platform/CMakeLists.txt
minosgalanakis: The equivalent could be done by using 

~~~~~~
file(GLOB platform_src_files
     CONFIGURE_DEPENDS
     "${CMAKE_CURRENT_SOURCE_DIR}/*.c")
~~~~~~

What it could do is when you add new platform files, or remove rename something cmake will reconfigure even if you call --build. 

But looking at it closer it does not fit our needs for this case and come with a bit of performance overhead each time we call make to build. So do not worry about this comment

minosgalanakis (approved):
Looks good thanks

PR #700, opened 2026-02-25
Add driver and dispatch for MLDSA

### LINKED ISSUE #633: Write PSA driver for ML-DSA using mldsa-native

gilles-peskine-arm:
Driver and dispatch for ML-DSA-87, pure only, deterministic signature only. Resolves https://github.com/Mbed-TLS/TF-PSA-Crypto/issues/633.

Continues https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/690.

Needs preceding PR:

* [x] https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/690
* [x] https://github.com/Mbed-TLS/mbedtls-framework/pull/282

## PR checklist

- [x] **changelog** provided | not required because: no user-visible feature yet
- [x] **framework PR** provided https://github.com/Mbed-TLS/mbedtls-framework/pull/282
- [x] **TF-PSA-Crypto development PR** provided https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/700
- [x] **TF-PSA-Crypto 1.1 PR** provided (partial) https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/750
- [x] **mbedtls development PR** (only creating `scripts/maintainer`) https://github.com/Mbed-TLS/mbedtls/pull/10684
- [x] **mbedtls 4.1 PR** (only creating `scripts/maintainer`) https://github.com/Mbed-TLS/mbedtls/pull/10685
- [x] **mbedtls 3.6 PR** (only creating `scripts/maintainer`) https://github.com/Mbed-TLS/mbedtls/pull/10686
- **tests**  provided


drivers/pqcp/src/psa_crypto_mldsa.c
```diff
@@ -8,8 +8,170 @@
...
+        /* The only documented error is an internal error which is not
+         * supposed to ever happen (failure of a self-test). */
+        goto cleanup;
+    }
+
+    const uint8_t prefix[2] = { 0, 0 }; // pure ML-DSA with empty context
+    const size_t prefix_length = sizeof(prefix);
+    const uint8_t rnd[MLDSA_RNDBYTES] = { 0 };
+
+    ret = tf_psa_crypto_pqcp_mldsa87_signature_internal(signature,
```
bjwtaylor, line 141: Not sure if this is by design or something missed, however it appears here you are capturing a return and then not returning it? As far as I can tell the called functions have error cases that we presumably should be taking note of, is there a reason we ignore them and return success?

drivers/pqcp/src/psa_crypto_mldsa.h
```diff
@@ -9,6 +9,136 @@
...
+    uint8_t *data, size_t data_size, size_t *data_length);
+
+/** Sign a message using pure-ML-DSA (without pre-hashing).
+ *
+ * \param[in] attributes        The key attributes.
+ * \param[in] key_buffer        The key material. This must be a key pair
+ *                              in the standard representation, i.e.
+ *                              just the 32-byte seed.
+ * \param key_buffer_size       The size of \p key_buffer, in bytes.
+ * \param alg                   The algorithm:
```
bjwtaylor, line 85: The docs currently advertise support for #PSA_ALG_ML_DSA or #PSA_ALG_DETERMINISTIC_ML_DSA, however on line drivers/pqcp/src/psa_crypto_mldsa.c:86 we appear to use  psa_get_key_type(attributes) != PSA_KEY_TYPE_ML_DSA_KEY_PAIR rejecting PSA_ALG_ML_DSA. Is it the case that this is planned for future? In which case is this just the API definition and maybe we just need a note this is planned and not implemented?

tests/suites/test_suite_dispatch_transparent.function
```diff
@@ -0,0 +1,213 @@
...
+
+        /* Try verifying a signature that's too long */
+        TEST_CALLOC(buf, signature->len + 1);
+        memcpy(buf, signature->x, signature->len);
+        TEST_EQUAL(psa_driver_wrapper_verify_message(
+                       &attributes,
+                       public_key->x, public_key->len,
+                       alg_arg,
+                       message->x, message->len,
+                       signature->x, signature->len + 1),
```
bjwtaylor, line 206: Not sure this is what is intended as I suspect this data now has two errors in some cases the flipped byte and the length. Is it better to do one at a time?

tests/suites/test_suite_psa_crypto_mldsa.function
```diff
@@ -16,8 +16,227 @@
...
+        /* Try verifying a signature whose last byte is wrong */
+        signature->x[signature->len - 1] ^= 1;
+        TEST_EQUAL(tf_psa_crypto_mldsa_verify_message(
+                       &attributes,
+                       public_key->x, public_key->len,
+                       alg_arg,
+                       message->x, message->len,
+                       signature->x, signature->len - 1),
+                   PSA_ERROR_INVALID_SIGNATURE);
+        signature->x[signature->len - 1] ^= 1;
```
bjwtaylor: Not sure this is what is intended as I suspect this data now has two errors in some cases the flipped byte and the length. Is it better to do one at a time?

bjwtaylor (commented):
Some initial minor comments/questions, though still need to do a full review.

tests/suites/test_suite_dispatch_transparent.function
gilles-peskine-arm: The last-byte-wrong code flips the same bit twice, so it restores the original signature. But I could make a flipped copy, it would be easier to follow.

bjwtaylor: Either that or do the other tests and then flip it? Maybe avoiding the flip?

drivers/pqcp/src/psa_crypto_mldsa.c
```diff
@@ -8,8 +8,182 @@
...
+        return PSA_ERROR_HARDWARE_FAILURE;
+    }
+}
+
+static psa_status_t seed_to_public_key(
+    size_t bits,
+    const uint8_t *key_buffer, size_t key_buffer_size,
+    uint8_t *data, size_t data_size, size_t *data_length)
+{
+    if (key_buffer_size != 32) {
```
bjwtaylor: minor: use of magic numbers

drivers/pqcp/src/psa_crypto_mldsa.c
```diff
@@ -8,8 +8,182 @@
...
+        return PSA_ERROR_INVALID_ARGUMENT;
+    }
+
+    if (bits != 87) {
```
bjwtaylor, line 53: minor: use of magic numbers

drivers/pqcp/src/psa_crypto_mldsa.h
```diff
@@ -9,6 +9,136 @@
...
+ *         The key type or size registered in \p attributes is not supported.
+ * \retval #PSA_ERROR_INVALID_ARGUMENT
+ *         The key material is invalid.
+ * \retval #PSA_ERROR_BUFFER_TOO_SMALL
+ *         \p data_size is too small.
+ */
+psa_status_t tf_psa_crypto_mldsa_export_public_key(
+    const psa_key_attributes_t *attributes,
+    const uint8_t *key_buffer, size_t key_buffer_size,
+    uint8_t *data, size_t data_size, size_t *data_length);
```
bjwtaylor, line 76: minor:  Is there any reason this is data rather than something more readable? Is is specified in the PSA API?

bjwtaylor (commented):
Some more comments/questions though still need to check a few more things.

drivers/pqcp/src/psa_crypto_mldsa.h
gilles-peskine-arm: The parameter name isn't part of the API specification, since it isn't part of the function's signature. But `data` is the name that the PSA API uses. It makes more sense for the [API](https://arm-software.github.io/psa-api/crypto/1.4/api/keys/management.html#c.psa_export_public_key) where the input is a handle, so the output is the only buffer. But the driver specification generally keeps the same names as the API for parameters that have the same meaning, for consistency. So it's a choice between making it easier to relate the different interface layers, or easier to understand one specific function. I don't see a clear winner.

drivers/pqcp/src/psa_crypto_mldsa.c
gilles-peskine-arm: This is cryptography. Some magic numbers are very much baked into the specification of the algorithm. Here the best we can do is `#define PARAMETER_SET_87 87`.

drivers/pqcp/src/psa_crypto_mldsa.c
gilles-peskine-arm: This is cryptography. Some magic numbers are very much baked into the specification of the algorithm. But here I could see an argument for `#define SEED_SIZE 32`.

bjwtaylor: Yes, defining it as a constant would be better as it'll make the code more readable.

drivers/pqcp/src/psa_crypto_mldsa.c
bjwtaylor: Yes, defining it as a constant would be better as it'll make the code more readable.

drivers/pqcp/src/psa_crypto_mldsa.h
bjwtaylor: Yeah, that's a tricky one. I hadn't realised it was part of the API, that being the case it probably makes sense to leave it as is as relating the various interface layers is generally trickier than understanding the variables. So I'll close this one.

drivers/pqcp/src/psa_crypto_mldsa.c
gilles-peskine-arm: Sorry, I'm looking at the code again and I really don't see a way in which having
```
    if (bits != PARAMETER_SET_87) {
        /* Other parameter sets are not supported yet. */
        return PSA_ERROR_NOT_SUPPORTED;
    }
```
in this file would make the code more readable. The string `87` is baked into a lot of function names and constant names. The role of `87` in this piece of code is clear in context. I just don't see how adding `PARAMETER_SET_` here, and requiring the reader to double check the value of `PARAMETER_SET_87`, could make the code more readable rather than less.

valeriosetti (dismissed):

drivers/pqcp/src/psa_crypto_mldsa.c
bjwtaylor: So, if I'm honest I don't have a strong opinion so I'll leave it up to you and you're right in this case the argument for/against is quite marginal. So it's a very minor point. I guess the pros are we have 87 a few times through the patch, so if we repurposed this code for something that wasn't 87 or 87 was wrong in some way we could easily change them (though suspect that's unlikely in this case). Also by giving it a readable name we can provide traceability back to where it came from, I guess in this is just the number of bits in the key? Which admittedly is not a massive thing to work out.  As you point out the cons are that you actually end up with more code, particularly on that line.

bjwtaylor (dismissed):
So long as the CI passes, the other comments are minor. So happy to approve and then leave it up to you.

drivers/pqcp/src/psa_crypto_mldsa.c
gilles-peskine-arm: 87 is the same number (or more accurately the same substring) that appears in the names of mldsa-native functions and in some macros (`MLDSA87_PUBLICKEYBYTES`, `tf_psa_crypto_pqcp_mldsa87_verify`, etc.).

We already know that at some point in the future (likely not soon), we'll want to support the other ML-DSA parameter sets (44 and 65). Then the code will need to be more complicated to allow for up to three different `bits` values, each of which can be independently enabled in the compile-time configuration, each of which requiring to call a different function. That will require checking pretty much every line of the code to make sure we're calling the right functions, using the right buffer sizes, etc. Using a symbolic constant for 87 where it appears as a number will not help us at all.

gilles-peskine-arm:
I rebased on top of the target branch and changed the framework submodule updat to take the head of `main`.

valeriosetti (approved):

bjwtaylor (approved):

PR #720, opened 2026-03-20
Remove pattern constraint from location field

parmi93:
According to the "PSA Certified Crypto Driver Interface" (Section 3.2) https://arm-software.github.io/psa-api/crypto-driver/1.0/body/manifest.html#driver-description-top-level-element, the location field may be either a non-negative integer or a C constant expression of type `psa_key_location_t`. For this reason, the pattern constraint is being removed, allowing the user to input any value.

## PR checklist
- [x] **changelog** not required because: minor enhancement to functionality that isn't officially documented
- [x] **TF-PSA-Crypto development PR** provided https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/720
- [x] **TF-PSA-Crypto 1.1 PR** provided https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/851
- [x] **mbedtls 3.6 PR** not required because: 3.6 has less than a year to live, so we don't encourage new driver developments
- **tests**  not required because: we don't test driver wrapper generators


parmi93:
Now user can set up a JSON manifest file like this:
```json
{
    "prefix": "stsafe_a110",
    "type": "opaque",
    "location": "STSAFE_A110_LOCATION", /* <- Previously, only hexadecimal values were permitted. */
    "headers": [
        "stsafe_a110_psa_driver.h"
    ],
    "capabilities": [
        {
            "entry_points": [
                "_comment": "bla bla"
            ],
        }
    ]
}
```

`stsafe_a110_psa_driver.h`:
```c
#define STSAFE_A110_LOCATION ((psa_key_location_t)0x800001)
```

gilles-peskine-arm (approved):
LGTM

gilles-peskine-arm:
I'm curious, are you finding the script to be usable as it is? I'm aware that only a small subset of the functionality is implemented.

parmi93:
The script (`generate_driver_wrappers.py`) does the job. As for the Jinja templates (`psa_crypto_driver_wrappers.h.jinja` and `psa_crypto_driver_wrappers_no_static.c.jinja`), they are still incomplete since only a small number of entry points are supported. So I end up patching the templates for my needs, but apart from that I didn’t have to touch anything else (except having to manually call the script (`generate_driver_wrappers.py`) because I need to pass it the path to my patched Jinja templates).

scripts/data_files/driver_jsons/driver_opaque_schema.json
```diff
@@ -14,8 +14,7 @@
       "const": "opaque"
     },
     "location": {
-      "type": ["integer","string"],
-      "pattern": "^(0x|0X)?[a-fA-F0-9]+$"
+      "type": ["integer","string"]
```
bjwtaylor, line 17: This as a standalone change looks good, though I think it may cause issues in areas such as scripts/data_files/driver_templates/OS-template-opaque.jinja:12. Currently we take the location as is, so if we change this to a string it is possible for this to emit invalid c. What do you think @gilles-peskine-arm my understanding of this code is not perfect, so this maybe covered by another part of the system. It could also be out of scope of this PR?

gilles-peskine-arm: Sure, it's possible to emit invalid C. But that can't be helped

This field is meant to allow values such as `"ACME_LOCATION_RANGE_START + 2"` where `ACME_LOCATION_RANGE_START` is a macro defined in some header that gets included during the build.

bjwtaylor: OK, if we're happy with this. Then it looks good.

bjwtaylor (approved):

gilles-peskine-arm:
The CI is failing. Would you mind rebasing your patch on top of the current head of the `development` branch?

Also, please make a pull request with the same patch targeting the `tf-psa-crypto-1.1` long-term support branch.

parmi93:
Rebased!

PR #724, opened 2026-03-22
cmake: make Thread package search quiet

valeriosetti:
## Description

This prevents printing message

"-- Could NOT find Threads (missing: Threads_FOUND)"

on platforms like Zephyr where threading is not provided by standard libraries.

## PR checklist

- [ ] **changelog** not required because: minor CMake message removal
- [ ] **mbedtls development PR** provided: https://github.com/Mbed-TLS/mbedtls/pull/10649
- [ ] **mbedtls 3.6 PR** not required because: not backported
- **tests** not required because: only cmake change

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -143,7 +143,7 @@ endif()
 # We now potentially need to link all executables against PThreads, if available
 set(CMAKE_THREAD_PREFER_PTHREAD TRUE)
 set(THREADS_PREFER_PTHREAD_FLAG TRUE)
-find_package(Threads)
+find_package(Threads QUIET)
 
 # If this is the root project add longer list of available CMAKE_BUILD_TYPE values
 if(NOT TF_PSA_CRYPTO_AS_SUBPROJECT)


CMakeLists.txt
```diff
@@ -141,9 +141,11 @@ else()
...
 
 # We now potentially need to link all executables against PThreads, if available
-set(CMAKE_THREAD_PREFER_PTHREAD TRUE)
-set(THREADS_PREFER_PTHREAD_FLAG TRUE)
-find_package(Threads)
+if(NOT CMAKE_SYSTEM_NAME STREQUAL "Generic")
+    set(CMAKE_THREAD_PREFER_PTHREAD TRUE)
+    set(THREADS_PREFER_PTHREAD_FLAG TRUE)
+    find_package(Threads)
+endif()
```
bjwtaylor: Not sure this is the best solution, it may work for this specific case however are we sure all Generic systems will not have threads? Maybe something like:

  set(CMAKE_THREAD_PREFER_PTHREAD TRUE)
  set(THREADS_PREFER_PTHREAD_FLAG TRUE)
  find_package(Threads QUIET)

might be better. You should probably also account for the scenario where LINK_WITH_PTHREAD=ON is defined by producing a FATAL_ERROR? 

Also the files core/CMakeLists.txt:82, tests/CMakeLists.txt:1, programs/psa/CMakeLists.txt:34, and
  programs/test/CMakeLists.txt:15 make unconditional assumption threading is available. Maybe this needs to be tightened? Other option is to key thread support of another variable, so we can make this zephr specific? What do you think?

valeriosetti: > find_package(Threads QUIET)

Oh right, this seems a better solution

> Also the files core/CMakeLists.txt:82, tests/CMakeLists.txt:1, programs/psa/CMakeLists.txt:34, and
programs/test/CMakeLists.txt:15 make unconditional assumption threading is available

But I don't get this: how making `find_package` quiet or using my solution resolve these problems? I think the build will fail in both cases, isn't it?

bjwtaylor: Sorry, probably should have done that as two separate comment as they are unrelated and you are correct the threads quiet will not solve the second solution and we will probably need something different to fix this.

valeriosetti: Ok, got it ;)
Thanks for the suggestion though! It's even more compact than what I proposed. Functionally it doesn't solve the problem, but it also doesn't make things worse, so that look reasonable to me

davidhorstmann-arm: > Also the files core/CMakeLists.txt:82, tests/CMakeLists.txt:1, programs/psa/CMakeLists.txt:34, and
programs/test/CMakeLists.txt:15 make unconditional assumption threading is available.

I'm not sure this is quite true. The use of threads in `core/CMakeLists.txt` is guarded by the setting of `LINK_WITH_PTHREAD`, an option which is [off by default](https://github.com/valeriosetti/TF-PSA-Crypto/blob/skip-thread-cmake-search/CMakeLists.txt#L117). The other references are in programs and tests, which can both be turned off (and indeed don't work on an embedded platform without an OS, threads, and Filesystem IO).

davidhorstmann-arm (approved):
LGTM, thanks!

CMakeLists.txt
bjwtaylor: Yeah, sounds reasonable. So I think we can close this one.

bjwtaylor (approved):
Approved subject to CI.

PR #727, opened 2026-03-26
Remove redundant declarations and update docs

### LINKED ISSUE #10376: Redundant declarations in psa/crypto.h and psa/crypto_struct.h
### Summary

When compiling with -Wredundant-decls and including "psa/crypto.h", we get many redundant declaration warnings.

### System information

Mbed TLS version (number or commit id): 3.6.4
Operating system and version: embedded
Configuration (if not default, please attach `mbedtls_config.h`): 
Compiler and options (if you used a pre-built binary, please indicate how you obtained it): -Wredundant-decls
Additional environment information: 

### Expected behavior

When compiling with -Wredundant-decls, no warnings shall be generated.

### Actual behavior

For example:
```
psa/crypto.h:126:29: error: redundant redeclaration of 'psa_key_attributes_init'
psa/crypto_struct.h:329:43: note: previous definition of 'psa_key_attributes_init' was here
```

This is due to the fact, that the functions that the functions in crypto.h are declared without inline, for example:
```
static psa_key_attributes_t psa_key_attributes_init(void);
```
While in crypto_structs.h they are defined as inline function, for example:
```
static inline struct psa_key_attributes_s psa_key_attributes_init(void)
```

### Steps to reproduce

compile a c file that includes psa/crypto.h with -Wredundant-decls.

bjwtaylor:
## Description

Currently we have a duplicate definition warning caused by the same definition being in crypto.h and crypto_structs.h My understanding from the docs is crypto.h is the public API and crypto_structs.h is the backend implementation. If I have understood correctly there will always be a crypto_structs.h or similar definition and this will always be the definition we need the system to use. Therefore what I can tell this means crypto.h exists purely for documentation. I have therefore resolved the issue by guarding the duplicate definitions with  __DOXYGEN_ONLY__. My understanding is this will remove them from compilation, however they will still be used in the API documentation.

Fix https://github.com/Mbed-TLS/mbedtls/issues/10376

## PR checklist

- [ ] **changelog** provided
- [ ] **mbedtls development PR** provided Mbed-TLS/mbedtls#https://github.com/Mbed-TLS/mbedtls/pull/10659
- [ ] **mbedtls 3.6 PR** not required because: No Backports
- **tests**  not required because: No changes.

### DIFF
--- a/include/psa/crypto.h
+++ b/include/psa/crypto.h
@@ -111,6 +111,7 @@ psa_status_t psa_crypto_init(void);
  * @{
  */
 
+#ifdef __DOXYGEN_ONLY__
 /** \def PSA_KEY_ATTRIBUTES_INIT
  *
  * This macro returns a suitable initializer for a key attribute structure
@@ -120,7 +121,9 @@ psa_status_t psa_crypto_init(void);
 /** Return an initial value for a key attributes structure.
  */
 static psa_key_attributes_t psa_key_attributes_init(void);
+#endif
 
+#ifdef __DOXYGEN_ONLY__
 /** Declare a key as persistent and set its key identifier.
  *
  * If the attribute structure currently declares the key as volatile (which
@@ -147,8 +150,10 @@ static psa_key_attributes_t psa_key_attributes_init(void);
  */
 static void psa_set_key_id(psa_key_attributes_t *attributes,
                            mbedtls_svc_key_id_t key);
+#endif
 
 #ifdef MBEDTLS_PSA_CRYPTO_KEY_ID_ENCODES_OWNER
+#if defined(__DOXYGEN_ONLY__)
 /** Set the owner identifier of a key.
  *
  * When key identifiers encode key owner identifiers, psa_set_key_id() does
@@ -166,7 +171,9 @@ static void psa_set_key_id(psa_key_attributes_t *attributes,
 static void mbedtls_set_key_owner_id(psa_key_attributes_t *attributes,
                                      mbedtls_key_owner_id_t owner);
 #endif
+#endif
 
+#ifdef __DOXYGEN_ONLY__
 /** Set the location of a persistent key.
  *
  * To make a key persistent, you must give it a persistent key identifier
@@ -195,7 +202,9 @@ static void mbedtls_set_key_owner_id(psa_key_attributes_t *attributes,
  */
 static void psa_set_key_lifetime(psa_key_attributes_t *attributes,
                                  psa_key_lifetime_t lifetime);
+#endif
 
+#ifdef __DOXYGEN_ONLY__
 /** Retrieve the key identifier from key attributes.
  *
  * This function may be declared as `static` (i.e. without external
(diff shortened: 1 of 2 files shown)


include/psa/crypto.h
```diff
@@ -111,6 +111,7 @@ psa_status_t psa_crypto_init(void);
  * @{
  */
 
+#ifdef __DOXYGEN_ONLY__
```
gilles-peskine-arm, line 114: Marking some function prototypes as Doxygen-only doesn't make sense. The reason to have prototypes is so that the C compiler sees them.

The reason some `static inline` functions have a prototype in `crypto.h` and a definition in `crypto_struct.h` is that the prototypes are the same for everyone, but the function definitions are not. By default, TF-PSA-Crypto presents itself a library. But it can also be integrated as a client and server, in which case the integrator makes some changes:

* On the client side:
    * Define `MBEDTLS_PSA_CRYPTO_CLIENT` but not `MBEDTLS_PSA_CRYPTO_C`.
    * Edit `crypto_struct.h` to change the definitions of operation structures, typically to have an operation handle instead of cryptographic material.
    * Provide implementations of the PSA API functions (excluding the `static inline` functions that manipulate a local struct) that makes remote procedure calls to the server.
* On the server side:
    * Probably define `MBEDTLS_PSA_CRYPTO_SPM` and likely `MBEDTLS_PSA_CRYPTO_KEY_ID_ENCODES_OWNER`.
    * Implement the server side of the remote procedure calls.

While the _definitions_ of the struct constructors in `crypto_struct.h` are customizable, their _prototypes_ are not. Hence the prototypes live in `crypto.h`. And the prototypes must be visible to the C compiler.

gilles-peskine-arm: That being said, is the current code really what it should be? No, there are two things that I never got around to fixing.

* `psa_key_attributes_t` is not supposed to be user-serviceable, unlike _operation_ structures. So it should be in `crypto_types.h`, and its inline accessor functions should not be defined in `crypto_struct.h`.
* The definitions of the struct constructor functions (`psa_xxx_init()`) don't actually need to be in `crypto_struct.h`, because their textual definition is actually the same for everyone. They're defined in terms of the corresponding `PSA_XXX_INIT` macro. Only the macro needs to be customized if you customize the struct layout.

We could fix this. It would be an incompatible change for integrators who customize `crypto_struct.h`. We don't explicitly promise any backward compatibility for such custom integrations, so I think it would be ok to do what I just wrote in the development branch. But I don't think we should do it in an LTS branch unless there's a compelling reason, and I don't see a compelling reason.

gilles-peskine-arm (changes requested):
The proposed change is not desirable.

It's not really clear to me what the goal is. If the goal is to compile with no warnings from `gcc -Wredundant-decls`, we need a different solution. That could be some header restructuring, or selectively disabling the warning using pragmas.

_If_ we make our code clean with `-Wredundant-decls`, we need to add this to our build scripts so that it's enforced. However, it's not fully clear to me why this is desirable. We've never promised to pay attention to this warning and I don't see why it's useful.

include/psa/crypto.h
bjwtaylor: OK, thanks for your comments. Let's put this one on hold pending design review of the original issue. There is also a related PR here: https://github.com/Mbed-TLS/mbedtls/pull/10659. It's the same issue, but not the same solution. Let me know what you think about this one, maybe we just decide we don't want to fix these warnings?

PR #737, opened 2026-04-03
Add metatests for the fork test helper as unit tests

gilles-peskine-arm:
Together with its framework prerequisite, resolves https://github.com/Mbed-TLS/mbedtls-framework/issues/295.

I am adding a test suite `test_suite_metatest` whose sole job is to test code that is in the test framework. Ideally, this test suite should be in the test framework. However we don't currently have a place for unit tests in the test framework, and I don't think it's worth the trouble to add them. So instead I am adding the tests in a consuming branch. They don't need to be backported unless the test code that's being tested starts having different behavior based on the consuming branch.

Needs preceding PR: https://github.com/Mbed-TLS/mbedtls-framework/pull/297

## PR checklist

- [x] **framework PR** provided https://github.com/Mbed-TLS/mbedtls-framework/pull/297
- [x] **TF-PSA-Crypto 1.1 PR** not required because: can be updated whenever
- [x] **mbedtls development PR** not required because: does not use fork helper
- [x] **mbedtls 4.1 PR** not required because: does not use fork helper
- [x] **mbedtls 3.6 PR** not required because: can be updated whenever
- **tests**  provided

### DIFF
--- a/tests/suites/test_suite_metatest.function
+++ b/tests/suites/test_suite_metatest.function
@@ -155,6 +155,18 @@ static void child_callback(void *param,
                            size_t *output_length)
 {
     child_instructions_t *instructions = param;
+    sigset_t sigs;
+    sigemptyset(&sigs);
+
+    /* If we're supposed to signal ourselves at the end, arrange to block
+     * the signal during this function (to avoid weird behavior if the signal
+     * is already blocked but pending for some reason), and unblock it later.
+     */
+    if (instructions->signal > 0) {
+        TEST_ASSERT_ERRNO(sigaddset(&sigs, instructions->signal) == 0);
+        TEST_ASSERT_ERRNO(sigprocmask(SIG_BLOCK, &sigs, NULL) == 0);
+        TEST_ASSERT_ERRNO(signal(instructions->signal, SIG_DFL));
+    }
 
     (void) output_size;
     if (instructions->data_size > 0) {
@@ -175,7 +187,12 @@ static void child_callback(void *param,
 
 exit:
     if (instructions->signal > 0) {
-        signal(instructions->signal, SIG_DFL);
+        if (sigprocmask(SIG_UNBLOCK, &sigs, NULL) != 0) {
+            /* We couldn't unblock the signal. This is really weird.
+             * We don't have much of a way to report what's happening.
+             * So exit with a unique status, which the parent can report. */
+            _exit(128);
+        }
         raise(instructions->signal);
     }
     if (instructions->exit >= 0) {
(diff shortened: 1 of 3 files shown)


tests/suites/test_suite_metatest.function
```diff
@@ -0,0 +1,417 @@
...
+    }
+
+    if (instructions->fail_write_result) {
+        extern int mbedtls_test_fork_child_fd;
+        close(mbedtls_test_fork_child_fd);
+    }
+
+exit:
+    if (instructions->signal > 0) {
+        signal(instructions->signal, SIG_DFL);
```
bjwtaylor: Do we also need something like 
  sigset_t set;
  sigemptyset(&set);
  sigaddset(&set, instructions->signal);
  sigprocmask(SIG_UNBLOCK, &set, NULL);
to unblock the signal mask? As it's possible the fork will inherit the masked state with an already existing block and at this point the signal will not terminate correctly? The commit already hints at the fact it could be masked, so is there a reason we don't unblock it?

tests/suites/test_suite_metatest.function
```diff
@@ -0,0 +1,417 @@
...
+    instructions.data_size = sizeof(output);
+    instructions.reported_length = sizeof(output);
+    int resource = -1;
+    unsigned long limit = 0;
+    struct rlimit saved_limit, new_limit;
+    void (*saved_handler)(int) = SIG_DFL;
+
+    switch (fault) {
+        case FORK_FAIL_FORK:
+            resource = RLIMIT_NPROC;
```
bjwtaylor, line 390: minor: If I've understood correctly this code begins a process which essentially sets RLIMIT_NPROC to 0, with the hope this will cause the fork to fail. Will this work for all cases as it can be bypassed for privileged users and can behave differently under container/runtime setups and do we care? If we do maybe a hook would be better?

bjwtaylor (commented):
Still need to do a full review, however some initial questions/comments.

tests/suites/test_suite_metatest.function
gilles-peskine-arm: This is getting into YAGNI territory, but I've made the signal masking more robust.

tests/suites/test_suite_metatest.function
gilles-peskine-arm: Resource limits are simple and relatively portable. (I've restricted them to Linux, but I think it's mostly a matter of coercing headers to declare the right things, but also possibly of some OSes not enforcing certain limits.)

Hooking into a system call is more dependent how you build the code and on the runtime environment. I'm not sure if it would work on WSL or Cygwin, for example, whereas resource limits should work (although I haven't verified that, and I don't care all that much if they don't work).

A container shouldn't matter (and can make hooking into system calls harder). Running those tests as a privileged user won't work, but that's not something we would do, so I don't think that matters, but I've added a comment.

PR #743, opened 2026-04-08
platform: fix -Wcast-align warnings in memory_buffer_alloc.c

M-Moawad:
Cast from 'unsigned char *' to 'memory_header *' through an intermediate 'void *' to suppress -Wcast-align warnings.

Some Clang-based toolchains (e.g. MetaWare/ARC) enable -Wcast-align as part of -Wall, unlike standard Clang on x86/ARM. Combined with -Werror this turns the casts into fatal build errors.

The casts are already alignment-safe at runtime:
- In mbedtls_memory_buffer_alloc_init(), buf is explicitly aligned to MBEDTLS_MEMORY_ALIGN_MULTIPLE before the cast.
- In buffer_alloc_calloc(), p is computed from an aligned base plus aligned offsets (sizeof(memory_header) and len are both multiples of MBEDTLS_MEMORY_ALIGN_MULTIPLE).
- In buffer_alloc_free(), p is derived from a previously aligned allocation pointer minus the aligned header size.

Original issue reported on zephyr: https://github.com/zephyrproject-rtos/zephyr/issues/106712

## PR checklist

- [x] **changelog** not required because: minor fix
- [x] **tf-psa-crypto 1.1 backport** https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/757
- [x] **mbedtls 3.6 PR** provided https://github.com/Mbed-TLS/mbedtls/pull/10692
- **tests** not required because: no new code to be tested

### DIFF
--- a/platform/memory_buffer_alloc.c
+++ b/platform/memory_buffer_alloc.c
@@ -294,7 +294,9 @@ static void *buffer_alloc_calloc(size_t n, size_t size)
     }
 
     p = ((unsigned char *) cur) + sizeof(memory_header) + len;
-    new = (memory_header *) p;
+    /* Double casting is required to prevent compilation warning on Clang-based
+     * compilers when "-Wcast-align" is used. */
+    new = (memory_header *) (void *) p;
 
     new->size = cur->size - len - sizeof(memory_header);
     new->alloc = 0;
@@ -375,7 +377,9 @@ static void buffer_alloc_free(void *ptr)
     }
 
     p -= sizeof(memory_header);
-    hdr = (memory_header *) p;
+    /* Double casting is required to prevent compilation warning on Clang-based
+     * compilers when "-Wcast-align" is used. */
+    hdr = (memory_header *) (void *) p;
 
     if (verify_header(hdr) != 0) {
         mbedtls_exit(1);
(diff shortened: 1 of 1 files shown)


valeriosetti:
@M-Moawad I suspect that using "Fixes" in the description of this PR will close the pointed issue once this PR is merged which is not correct. That issue will only be closed when this PR is cherry-picked to [Zephyr fork of this repo](https://github.com/zephyrproject-rtos/TF-PSA-Crypto)

M-Moawad:
Modified

platform/memory_buffer_alloc.c
```diff
@@ -294,7 +294,7 @@ static void *buffer_alloc_calloc(size_t n, size_t size)
     }
 
     p = ((unsigned char *) cur) + sizeof(memory_header) + len;
-    new = (memory_header *) p;
+    new = (memory_header *) (void *) p;
```
valeriosetti, line 299: Would you mind adding a comment here (and in other occurrences fixed in this PR) giving the reason for the double casting so that it doesn't get erroneously removed in the future?
Something like:

> Double casting is required to prevent compilation warning on Clang-based compilers when "-Wcast-align" is used.

valeriosetti (approved):

gilles-peskine-arm (approved):
LGTM

At runtime, we know that the start of the heap is aligned to `MBEDTLS_MEMORY_ALIGN_MULTIPLE`, which is supposed to be good enough for the platform's alignment requirements. So the warning is spurious, but the compiler can't know. This patch silences the warning in a sensible way.

OK to not have a changelog entry since silencing compiler warnings is on a best effort basis and this one doesn't have an issue on our bug tracker.

gilles-peskine-arm:
Thanks for the patch! Can you please backport it to the `tf-psa-crypto-1.1` long-time support branch, and to the `mbedtls-3.6` LTS branch on https://github.com/Mbed-TLS/mbedtls ?

valeriosetti:
> Thanks for the patch! Can you please backport it to the `tf-psa-crypto-1.1` long-time support branch, and to the `mbedtls-3.6` LTS branch on https://github.com/Mbed-TLS/mbedtls ?

I did those:
- https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/757
- https://github.com/Mbed-TLS/mbedtls/pull/10692

PR #751, opened 2026-04-10
Detect getrandom() on Android (Bionic libc)

### LINKED ISSUE #745: Allow Bionic or other libc to use getrandom() syscall
### Suggested enhancement

Currently, it looks like getrandom() is only supported if you're using glibc or (minipix?).
Could this maybe be expanded to Bionic and or other C libraries?

I'm not really sure what the best way to handle this would be, or even if maybe I'm the only one who has or will experience this one issue.

### Justification

I ran into a bug with an app I'm developing when run under Android wherein libcurl would hang during initialization.

As it turns out, my app gets stuck waiting on a call to read() from /dev/random and my embedded Android system doesn't seem to be very good at generating entropy on it's own.

After tracking down the path of libcurl->mbedtls->tf-psa-crypto, I noticed that tf-psa-crypto reads from /dev/random if getrandom() isn't available.
I did a quick hack to add \_\_BIONIC\_\_ to the "test for Linux getrandom() support" test and my app launched instantly.

﻿﻿Line in question: https://github.com/Mbed-TLS/TF-PSA-Crypto/blob/1e3d9f1ac9e4207a1166a1b5f6c790b7df3bb697/platform/platform_util.c#L297

gilles-peskine-arm:
Fixes #745

## PR checklist

- [x] **changelog** provided
- [x] **TF-PSA-Crypto development PR** provided here
- [ ] **TF-PSA-Crypto 1.1 PR** TODO
- [ ] **mbedtls 3.6 PR** TODO
- **tests**  not required because: we don't test with Bionic

### DIFF
--- a/ChangeLog.d/bionic-getrandom.txt
+++ b/ChangeLog.d/bionic-getrandom.txt
@@ -1,0 +1,2 @@
+Features
+   * Use getrandom() rather than /dev/random on Android. Fixes #745.
--- a/platform/platform_util.c
+++ b/platform/platform_util.c
@@ -294,9 +294,12 @@ int mbedtls_platform_get_entropy(psa_driver_get_entropy_flags_t flags,
  * Since there is no wrapper in the libc yet, use the generic syscall wrapper
  * available in GNU libc and compatible libc's (eg uClibc).
  */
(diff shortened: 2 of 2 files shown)


gilles-peskine-arm:
Subsumed by https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/843

PR #753, opened 2026-04-13
1.1 only: Fix the build when a driver calls the dispatch layer

gilles-peskine-arm:
Fix the build when a driver calls the dispatch layer. We don't currently do this in our code base, but partners who integrate their own drivers want to do this.

Done in `development` in https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/704.

## PR checklist

- [x] **changelog** not required because: fixing something that we don't officially promise and that hasn't been reported as an issue
- [x] **mbedtls 3.6 PR** not required because: only relevant after 1.0
- **tests**  not required because: fixing something that we don't officially promise

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -408,6 +408,8 @@ endif()
 # drivers, that are appended in `drivers/CMakeLists.txt`.
 #
 set(TF_PSA_CRYPTO_PRIVATE_INCLUDE_DIRS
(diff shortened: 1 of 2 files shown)


valeriosetti:
isn't the destination branch wrong?

gilles-peskine-arm:
Oops, yes, thanks.

valeriosetti (approved):

bjwtaylor (approved):

PR #755, opened 2026-04-14
tests: add CMake DESTDIR install coverage in components-build-system

yiwu0b11:
## Description

Support for DESTDIR env for build install is already in makefile.
Tests are added

Issue: https://github.com/Mbed-TLS/mbedtls/issues/10627
Side port of: https://github.com/Mbed-TLS/mbedtls/pull/10631

## PR checklist

- [x] **framework PR**  not required
- [x] **mbedtls development PR** provided: https://github.com/Mbed-TLS/mbedtls/pull/10631
- [x] **TF-PSA-Crypto development** provided here
- [x] **TF-PSA-Crypto 1.1** provided: https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/756
- [x] **mbedtls 3.6 PR** provided: https://github.com/Mbed-TLS/mbedtls/pull/10637
- **tests**  provided


tests/scripts/components-build-system.sh
```diff
@@ -164,6 +164,38 @@ component_tf_psa_crypto_build_config_name () {
...
+
+    DESTDIR="$OUT_OF_SOURCE_DIR/stage" make install
+
+    install_lib_subdir="$(sed -n 's/^CMAKE_INSTALL_LIBDIR:PATH=//p' CMakeCache.txt)"
+    [ -n "$install_lib_subdir" ] # Failed to read CMAKE_INSTALL_LIBDIR from CMakeCache.txt
+
+    install_lib_path="$OUT_OF_SOURCE_DIR/stage/usr/${install_lib_subdir}"
+
+    if [[ "$OSTYPE" == darwin* ]]; then
+        [ -f "$install_lib_path/libtfpsacrypto.a" ]
```
gilles-peskine-arm: Minor: the `.a` file is the same everywhere, so this line could be outside the if block.

gilles-peskine-arm (changes requested):
The content in TF-PSA-Crypto is very different from Mbed TLS due to differences with generated files and having a single library, so here I don't mind just writing similar code without copying the development history.

So ok for the history. There's just one line that it doesn't make sense to duplicate here given the simpler structure.

tests/scripts/components-build-system.sh
```diff
@@ -164,6 +164,37 @@ component_tf_psa_crypto_build_config_name () {
     rm -f include/psa/crypto_config_full.h
 }
 
+component_tf_psa_crypto_install_with_destdir () {
+    msg "install: cmake tf-psa-crypto with DESTDIR staging"
+    TF_PSA_CRYPTO_ROOT_DIR="$PWD"
+    cd "$OUT_OF_SOURCE_DIR"
+    cmake -DGEN_FILES=ON -DENABLE_PROGRAMS=OFF -DENABLE_TESTING=OFF -DUSE_SHARED_TF_PSA_CRYPTO_LIBRARY=ON -DCMAKE_INSTALL_PREFIX:PATH=/usr "$TF_PSA_CRYPTO_ROOT_DIR"
```
bjwtaylor: I think I'd be tempted to leave -DENABLE_PROGRAMS on. I suspect it'll make zero difference to the outcome, though I believe the programs are installed. So there's a possibility it'll effect the test.

tests/scripts/components-build-system.sh
```diff
@@ -164,6 +164,37 @@ component_tf_psa_crypto_build_config_name () {
...
+
+    DESTDIR="$OUT_OF_SOURCE_DIR/stage" make install
+
+    install_lib_subdir="$(sed -n 's/^CMAKE_INSTALL_LIBDIR:PATH=//p' CMakeCache.txt)"
+    [ -n "$install_lib_subdir" ] # Failed to read CMAKE_INSTALL_LIBDIR from CMakeCache.txt
+
+    install_lib_path="$OUT_OF_SOURCE_DIR/stage/usr/${install_lib_subdir}"
+
+    [ -f "$install_lib_path/libtfpsacrypto.a" ]
+    if [[ "$OSTYPE" == darwin* ]]; then
```
bjwtaylor, line 182: Are we required/expecting to support windows here?

yiwu0b11: Sorry for the long wait, after spend some time double check -  No, not expected - only test component that has "_mingw" in its name will be checked on Windows.
https://github.com/Mbed-TLS/mbedtls-framework/blob/c6610dde67ffd2a3a81cc204a73572b9c31a5775/scripts/all-core.sh#L876-L877
Additionally cmake "desdir" functionality only exist on UNIX system   https://cmake.org/cmake/help/latest/envvar/DESTDIR.html

gilles-peskine-arm: We don't try to run `all.sh` on Windows and I expect that a lot of things would break. It's not a problem to have one more place where Windows isn't supported.

We only run `all.sh` on Linux and FreeBSD in the CI, but some developers use macOS so it's better if the script also works on macOS.

Even the MinGW component doesn't run on Windows. It runs a cross-compiler on Linux that produces Windows binaries. We also do a MinGW build with a Windows host on the CI, but that doesn't use `all.sh`, it uses [`windows_testing.[y`](https://github.com/Mbed-TLS/mbedtls-test/blob/main/resources/windows/windows_testing.py).

gilles-peskine-arm (approved):
LGTM

bjwtaylor (approved):

Issue #381, opened 2025-07-17
Doxygen documentation built in the source tree instead of the build tree


gilles-peskine-arm:
In TF-PSA-Crypto, CMake builds the Doxygen documentation under the source tree instead of the build tree. Since the documentation depends on the configuration, it should be in the build tree. The “Documentation” section of `README.md` states that `apidoc` is in the build tree.

(And to start with I wondered where the doc was getting built: paths in CMake build trees are not obvious and `find build-default-cmake -name '*.html'` didn't find anything.)

```
git checkout da898a0d419d7b047d6d32966007c4e0fdbe907c # development right now
cmake -B build-default-cmake
make -C build-default-cmake tfpsacrypto-apidoc
```

This places the HTML documentation in the `apidoc` subdirectory of the source tree.

PR #760, opened 2026-04-15
[backport 1.1] doxygen: use CMake build dir intead of source dir to host generated files

valeriosetti:
## Description

Backport of #718. Resolves https://github.com/Mbed-TLS/TF-PSA-Crypto/issues/381.

## PR checklist

- [x] **changelog** provided
- [x] **TF-PSA-Crypto development PR** provided https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/718
- [x] **TF-PSA-Crypto 1.1 PR** here
- [x] **mbedtls 4.1 PR** not required because: no change there
- **tests**  not required because: no code change


bjwtaylor (commented):
Do we need to update framework/scripts/apidoc_full.sh:58, there appears to be at least a stale comment here. Though not sure if can just be removed or if there is a bigger issue?

valeriosetti:
> Do we need to update framework/scripts/apidoc_full.sh:58, there appears to be at least a stale comment here. Though not sure if can just be removed or if there is a bigger issue?

Thanks to this comment I made a quick comparison with what's being done in the `mbedtls` repo and I found that:
1. in `mbedtls` doxygen files are not compiled by CMake, whereas in `tf-psa-crypto` they are compiled.
2. in `mbedtls` the documentation is still built in-tree, whereas in `tf-psa-crypto` it won´t after this PR.

Point (2) suggest that:
- I need to open the corresponding PRs also in `mbedtls` (`development` and `4.1`) to fix this.
- I can drop that comment (and the associated action) from the framework script since it won´t apply anymore

gilles-peskine-arm (commented):
This is a technically correct backport of https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/718. However, [I'm not sure it's an acceptable change in an LTS branch](https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/718#discussion_r3065588527).

IIRC you suggested creating an `apidoc` symlink in the source tree, I think that would be an acceptable compromise (doing nothing if `apidoc` is an existing directory). (Although that doesn't work on Windows.)

valeriosetti:
> This is a technically correct backport of #718. However, [I'm not sure it's an acceptable change in an LTS branch](https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/718#discussion_r3065588527).

So just to be sure I understood correctly: the symlink should be done only for the LTS or also for `development` (i.e. #718)?

gilles-peskine-arm:
We don't have firm rules around build system changes. I think it's ok to keep things simple in the development branch, but I'm uncomfortable with changing the interface for the build of the documentation in an LTS branch. So my opinion is symlink in LTS, no symlink in development. But others may have a different opinion.

bjwtaylor (dismissed):

CMakeLists.txt
```diff
@@ -71,12 +71,25 @@ set(version_number_files
...
 endforeach(file)
 
 ADD_CUSTOM_TARGET(${TF_PSA_CRYPTO_TARGET_PREFIX}tfpsacrypto-apidoc
     COMMAND doxygen tfpsacrypto.doxyfile
-    WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/doxygen)
+    WORKING_DIRECTORY ${PROJECT_BINARY_DIR}/doxygen)
+
+# Create an in-tree symlink to the apidoc built in the binary folder. This is
+# only performed on Unix platforms and if "apidoc" doesn't exist in-tree.
+if(UNIX AND NOT EXISTS ${CMAKE_CURRENT_SOURCE_DIR}/apidoc)
```
gilles-peskine-arm: This checks for the existence of `apidoc` when you run `cmake`, not when you build the documentation.

One consequence is that if you do an in-tree build from a fresh checkout, cmake decides that you do want to create `apidoc`, but then it fails because that's an existing directory.

So at least this part should be skipped if the source tree is the build tree.

Furthermore, if you do an in-tree build after an out-of-tree build, the out-of-tree build overwrites the `apidoc` in the out-of-tree build. Admittedly that would be pretty unusual (I guess most users either do an in-tree build or one or more out-of-tree builds but not both), but it's definitely not desirable behavior.

Also, if you do two out-of-tree builds, which one will “win” `apidoc` depends on whether you do `cmake -B dir1; make -C dir1; cmake -B dir2; make -C dir2` or `cmake -B dir1; cmake -B dir2; make -B dir1; make -B dir2`. This isn't wrong (only one build can win and there's no particular reason to prefer a specific one), but it's weird.

ChangeLog.d/doxygen-build-folder.txt
```diff
@@ -0,0 +1,6 @@
+Bugfix
+   * Doxygen documentation `apidoc` is now built into CMake's binary folder,
+     not in-tree. For backward compatibility on Unix platforms an in-tree
+     symlink named `apidoc` is created pointing to the folder in the build
+     directory.
+     Fixes #381.
```
gilles-peskine-arm: Minor:

```suggestion
     directory. Fixes #381.
```

gilles-peskine-arm (changes requested):
LGTM except for in-tree builds

CMakeLists.txt
valeriosetti: this should be addressed now in https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/760/commits/db8d277d907d591f752e6a6902ba8530135953e0

CMakeLists.txt
```diff
@@ -79,15 +79,19 @@ ADD_CUSTOM_TARGET(${TF_PSA_CRYPTO_TARGET_PREFIX}tfpsacrypto-apidoc
...
+# symlink.
+if(UNIX AND NOT CMAKE_SOURCE_DIR STREQUAL CMAKE_BINARY_DIR)
     file(RELATIVE_PATH _apidoc_rel_path
         ${CMAKE_CURRENT_SOURCE_DIR}
         ${PROJECT_BINARY_DIR}/apidoc)
     add_custom_command(TARGET ${TF_PSA_CRYPTO_TARGET_PREFIX}tfpsacrypto-apidoc POST_BUILD
-        COMMAND ${CMAKE_COMMAND} -E create_symlink
-            ${_apidoc_rel_path}
-            ${CMAKE_CURRENT_SOURCE_DIR}/apidoc
+        COMMAND sh -c
```
gilles-peskine-arm:
```
$ cmake -B build-default-cmake
…
$ make -C build-default-cmake tfpsacrypto-apidoc
…
/bin/sh: 1:               /usr/bin/cmake -E create_symlink              build-default-cmake/apidoc /home/gilpes01/z/dev/crypto/reviewing/760-backport-pr718/apidoc: not found
make[3]: *** [CMakeFiles/tfpsacrypto-apidoc.dir/build.make:75: tfpsacrypto-apidoc] Error 127
```

Looks like there's a quoting problem here.

gilles-peskine-arm (changes requested):
The symlink creation doesn't work for me.

CMakeLists.txt
valeriosetti: Sorry, you're right. Since I didn't find a clean solution to this problem only relying on CMake stuff, I moved the symlink creation logic to a very small external shell script.

doxygen/create_symlink.sh
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env sh
+
+set -eu
+
+link_target="$1"
+link_path="$2"
+
+if [ ! -L "$link_path" ]; then
+    ln -s "$link_target" "$link_path"
+fi
```
gilles-peskine-arm: I think this should be
```suggestion
if [ ! -e "$link_path" ]; then
    ln -sf "$link_target" "$link_path"
fi
```
`-e rather than `-L` because if the path is an existing directory, we must not create a symlink inside it. Conversely, if the path is a broken symlink, it's desirable to overwrite it, hence `-f`.

gilles-peskine-arm (changes requested):
LGTM except for the condition for creating the symbolic link

gilles-peskine-arm (approved):
LGTM

bjwtaylor (approved):

Issue #744, opened 2026-04-08
MLDSA doesn't detect SHAKE failures


gilles-peskine-arm:
MLDSA, implemented through mldsa-native, doesn't detect failures of a SHAKE operation.

The SHAKE implementation in mldsa-native (used when `TF_PSA_CRYPTO_PQCP_OWN_SHAKE` is enabled) cannot fail. Our built-in SHAKE implementation (used otherwise, unless a PSA accelerator driver is enabled) cannot fail, except if the desired SHAKE variant is enabled, but we check for that at compile time. So the only configuration where a runtime failure is possible is when using a third-party SHAKE driver. (For testing, we can force a failure in our test driver.)

Since mldsa-native's SHAKE callbacks return `void`, there's no natural way to convey a failure.

PR #770, opened 2026-04-20
ML-DSA driver for multipart operations

### LINKED ISSUE #635: Multipart sign-message and verify-message in MLDSA driver
In the driver for ML-DSA using mldsa-native, add the PSA multipart sign-message and verify-message functions. The API is currently being finalized at https://github.com/ARM-software/psa-api/issues/319.

For now this is gated behind an experimental config option, default off but enabled in `full`, and not yet reachable from the API.

Since mldsa-native only provides a one-shot interface, we call its “external mu” interface (`signature_internal`, `verify_internal`). Use the `fips202.h` functions (`mld_shake256_init()` etc.) to calculate mu from the message.

Prototype: (which calls `mbedtls_sha3_xxx` directly, but we should instead call `fips202.h` functions): https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/599

Prerequisites: https://github.com/Mbed-TLS/TF-PSA-Crypto/issues/633, https://github.com/Mbed-TLS/TF-PSA-Crypto/issues/629, https://github.com/Mbed-TLS/TF-PSA-Crypto/issues/627

gilles-peskine-arm:
Add support for multipart pure ML-DSA signature at the driver layer. Resolves https://github.com/Mbed-TLS/TF-PSA-Crypto/issues/635

The signature operation allocates heap memory. We will revise this later, likely [when we give the signature function access to the expanded key](https://github.com/Mbed-TLS/TF-PSA-Crypto/issues/774).

Needs preceding PR: test case generation in https://github.com/Mbed-TLS/mbedtls-framework/pull/300

## PR checklist

- [x] **changelog** not required because: not user-facing yet
- [x] **framework PR** provided https://github.com/Mbed-TLS/mbedtls-framework/pull/300
- [x] **TF-PSA-Crypto 1.1 PR** not required because: new feature
- [x] **mbedtls 3.6 PR** not required because: crypto only
- [x] **tests**  provided


drivers/pqcp/src/psa_crypto_mldsa.h
```diff
@@ -141,4 +150,152 @@ psa_status_t tf_psa_crypto_mldsa_verify_message(
...
+    tf_psa_crypto_mldsa_operation_t *operation,
+    const uint8_t *input, size_t input_length);
+
+/** Finish a pure-ML-DSA signature operation.
+ *
+ * \param operation             An operation structure. It must have
+ *                              been set up for signing and not finished
+ *                              or aborted yet.
+ * \param[out] signature        On success, the exported key.
+ * \param[in] key_buffer        The key material. This must be the same seed
```
bjwtaylor, line 232: Key buffer and key_buffer are never used. Presumably they are specified in the API and then not needed by the function? Is there a plan to use these variables in the future and whatever the plan do we need to document the current reasoning behind this?

bjwtaylor (commented):
Need to do a full review, though a quick question I had noticed.

drivers/pqcp/src/psa_crypto_mldsa.h
gilles-peskine-arm: For ML-DSA, the signature calculation needs the public key during setup, and the expanded private key during finish. The standard PSA format for the private key is just the seed. Since the same mldsa-native function calculates both the expanded private key and the public key from the seed, we store the expanded private key during `sign_setup` and use it in `sign_finish` (where a comment explains this). It's a performance/memory trade-off: we could instead recalculate the private key in `sign_finish`.

In addition to this memory/performance tradeoff, there are other algorithms or key representation formats where it's more advantageous to not store any  key material between setup and finish. So I proposed that the driver interface would pass the key both to setup and finish. (The driver interface isn't officially defined yet.) https://github.com/ARM-software/psa-api/issues/350

We plan to also support a key representation that's fully expanded, in which case there's nothing to calculate and `sign_finish` would read the part that it needs from `key_buffer`. https://github.com/Mbed-TLS/TF-PSA-Crypto/issues/774

drivers/pqcp/src/psa_crypto_mldsa.h
```diff
@@ -141,4 +150,152 @@ psa_status_t tf_psa_crypto_mldsa_verify_message(
...
+ * \retval #PSA_ERROR_BAD_STATE
+ *         The state of \p operation has been detected as inconsistent
+ *         with the request. Note that this function does not guarantee
+ *         that an inconsistent state is detected.
+ */
+psa_status_t tf_psa_crypto_mldsa_update(
+    tf_psa_crypto_mldsa_operation_t *operation,
+    const uint8_t *input, size_t input_length);
+
+/** Finish a pure-ML-DSA signature operation.
```
bjwtaylor, line 227: Minor: There maybe a good reason for this, however the doxygen docs list the parameters in a different order to the prototype. If there's no good reason for it, then it may be worth re-ordering them to improve readability.

drivers/pqcp/src/psa_crypto_mldsa.c
```diff
@@ -193,4 +273,243 @@ psa_status_t tf_psa_crypto_mldsa_verify_message(
...
+
+    return PSA_SUCCESS;
+}
+
+psa_status_t tf_psa_crypto_mldsa_update(
+    tf_psa_crypto_mldsa_operation_t *operation,
+    const uint8_t *input, size_t input_length)
+{
+    mld_shake256_absorb(&operation->shake, input, input_length);
+    return PSA_SUCCESS;
```
bjwtaylor, line 424: This doesn't look right, if I've understood correctly the mld_shake256_absorb calls functions that can fail. I understand they are going through the driver wrapper though, so is there no way we can pass the error?

drivers/pqcp/src/psa_crypto_mldsa.h
bjwtaylor: OK is it worth adding a comment here and potentially at the void points to capture this?

drivers/pqcp/src/psa_crypto_mldsa.c
```diff
@@ -193,4 +273,243 @@ psa_status_t tf_psa_crypto_mldsa_verify_message(
...
+
+psa_status_t tf_psa_crypto_mldsa_sign_finish(
+    tf_psa_crypto_mldsa_operation_t *operation,
+    const uint8_t *key_buffer, size_t key_buffer_size,
+    uint8_t *signature, size_t signature_size, size_t *signature_length)
+{
+    *signature_length = 0;
+
+    if (operation->parameter_set != 87) {
+        return PSA_ERROR_NOT_SUPPORTED;
```
bjwtaylor, line 435: Do we have/need a test that covers this return?

drivers/pqcp/src/psa_crypto_mldsa.c
```diff
@@ -193,4 +273,243 @@ psa_status_t tf_psa_crypto_mldsa_verify_message(
...
+    if (signature_size < MLDSA87_BYTES) {
+        return PSA_ERROR_BUFFER_TOO_SMALL;
+    }
+
+    /* Rely on setup() having stored the expanded private key in the
+     * operation structure. This is a performance/memory trade-off:
+     * we could instead re-expand the private key from the seed
+     * in \p key_buffer here. */
+    if (operation->key_length != MLDSA87_SECRETKEYBYTES) {
+        return PSA_ERROR_CORRUPTION_DETECTED;
```
bjwtaylor, line 446: Do we have/need a test that covers this return?

drivers/pqcp/src/psa_crypto_mldsa.c
```diff
@@ -193,4 +273,243 @@ psa_status_t tf_psa_crypto_mldsa_verify_message(
...
+
+    int ret = tf_psa_crypto_pqcp_mldsa87_signature_internal(
+        signature, signature_length,
+        mu, sizeof(mu),
+        NULL, 0, rnd,
+        operation->key, 1);
+
+    psa_status_t abort_status = tf_psa_crypto_mldsa_abort(operation);
+    if (abort_status != PSA_SUCCESS) {
+        return abort_status;
```
bjwtaylor, line 467: Do we have/need a test that covers this return?

drivers/pqcp/src/psa_crypto_mldsa.c
```diff
@@ -193,4 +273,243 @@ psa_status_t tf_psa_crypto_mldsa_verify_message(
...
+    }
+    return pqcp_to_psa_error(ret);
```
bjwtaylor, line 469: Do we have/need a test that covers this return?

drivers/pqcp/src/psa_crypto_mldsa.c
```diff
@@ -193,4 +273,243 @@ psa_status_t tf_psa_crypto_mldsa_verify_message(
...
+}
+
+psa_status_t tf_psa_crypto_mldsa_verify_finish(
+    tf_psa_crypto_mldsa_operation_t *operation,
+    const uint8_t *key_buffer, size_t key_buffer_size,
+    const uint8_t *signature, size_t signature_length)
+{
+    if (operation->parameter_set != 87) {
+        return PSA_ERROR_NOT_SUPPORTED;
```
bjwtaylor, line 478: Do we have/need a test that covers this return?

drivers/pqcp/src/psa_crypto_mldsa.c
```diff
@@ -193,4 +273,243 @@ psa_status_t tf_psa_crypto_mldsa_verify_message(
...
+
+    int ret = tf_psa_crypto_pqcp_mldsa87_verify_internal(
+        signature, signature_length,
+        mu, sizeof(mu),
+        NULL, 0,
+        key_buffer, 1);
+
+    psa_status_t abort_status = tf_psa_crypto_mldsa_abort(operation);
+    if (abort_status != PSA_SUCCESS) {
+        return abort_status;
```
bjwtaylor, line 500: Do we have/need a test that covers this return?

drivers/pqcp/src/psa_crypto_mldsa.c
```diff
@@ -193,4 +273,243 @@ psa_status_t tf_psa_crypto_mldsa_verify_message(
...
+    }
+
+    if (ret == MLD_ERR_FAIL) {
+        return PSA_ERROR_INVALID_SIGNATURE;
```
bjwtaylor, line 504: Do we have/need a test that covers this return?

drivers/pqcp/src/psa_crypto_mldsa.c
```diff
@@ -193,4 +273,243 @@ psa_status_t tf_psa_crypto_mldsa_verify_message(
...
+    } else {
+        return pqcp_to_psa_error(ret);
```
bjwtaylor, line 506: Do we have/need a test that covers this return?

drivers/pqcp/src/psa_crypto_mldsa.h
gilles-peskine-arm: In the header, I don't think this warrants a comment: whether the parameter is used or not is an implementation detail. The API follows the applicable standard, or at least the current draft of that standard.

In the implementation, there's already a comment just above the void point.

drivers/pqcp/src/psa_crypto_mldsa.c
gilles-peskine-arm: `mld_xxx` functions return `void`.

The only cases where the functions can fail internally are:

* If there's a bug in our code and we end up calling `setup` in a configuration where the algorithm is not provided. This shouldn't happen thanks to compile-time guards.
* If SHAKE is provided by a third-party PSA driver (i.e. neither our implementation nor the one in mldsa-native) and that one has failure conditions, which is unusual, especially for an algorithm that doesn't use a secret key.

That second case is a bug, filed as https://github.com/Mbed-TLS/TF-PSA-Crypto/issues/744. Given how unusual it is for it to matter, it's very low priority.

drivers/pqcp/src/psa_crypto_mldsa.c
gilles-peskine-arm: No, because this can't happen except if there's an internal error somewhere. I put this check in partly for robustness, and partly as a signpost to indicate that in the future, when we support multiple sizes, we'll need to have a `switch` on the size.

drivers/pqcp/src/psa_crypto_mldsa.c
gilles-peskine-arm: No. Generally `PSA_ERROR_CORRUPTION_DETECTED` indicates that something happened that should not have happened, so there's no way to test it.

drivers/pqcp/src/psa_crypto_mldsa.c
gilles-peskine-arm: No. The only way `abort` can return a failure status is in case there's some internal error.

Arguably, given that the driver has no internal errors, we could make `tf_psa_crypto_mldsa_abort` return void. I followed the driver interface closely here because it's experimental and I wanted to make sure that it's suitable. Maybe in the future we can have a minor optimization phase where we remove this kind of won't-happen error path.

drivers/pqcp/src/psa_crypto_mldsa.c
gilles-peskine-arm: As above, there is no failure more for abort.

drivers/pqcp/src/psa_crypto_mldsa.c
gilles-peskine-arm: As above, it's currently impossible to get that far without having a supported size in the operation.

drivers/pqcp/src/psa_crypto_mldsa.c
gilles-peskine-arm: As above, there is no failure more for abort.

drivers/pqcp/src/psa_crypto_mldsa.c
gilles-peskine-arm: We do.

```
grep -c 'verify_multipart.*PSA_ERROR_INVALID_SIGNATURE' tests/suites/test_suite_psa_crypto_mldsa.*.data
```

drivers/pqcp/src/psa_crypto_mldsa.c
gilles-peskine-arm: Currently there are no possible errors from `verify`, other than the invalid-signature case. In the near future, there will be an `INSUFFICIENT_MEMORY` case.

bjwtaylor (dismissed):

bensze01 (dismissed):
LGTM, please rebase and update the framework pointer to point to the merge commit.

gilles-peskine-arm:
I rebased on top of current `development` to resolve the submodule conflict. The previous head is at https://github.com/gilles-peskine-arm/TF-PSA-Crypto/tree/mldsa-sign-multipart-driver-6

bensze01 (approved):
LGTM

bjwtaylor (approved):

PR #776, opened 2026-04-23
Add fix for memset_s issue discovered with ATFE

### LINKED ISSUE #754: Build broken with TF-M using ATfE in 1.1.0
TF-PSA-Crypto 1.1.0 breaks the build of TF-M (Trusted Firmware M) when using ATfE (Arm Toolchain for Embedded): after [upgrading](https://review.trustedfirmware.org/c/TF-M/trusted-firmware-m/+/49686), they needed a [workaround](https://review.trustedfirmware.org/c/TF-M/trusted-firmware-m/+/49740) to avoid querying for the presence of C11 Annex K functions. The [patch](https://review.trustedfirmware.org/plugins/gitiles/TF-M/trusted-firmware-m/+/d6e6cdee67c130ec999157baca8e37a52c06cfa7/lib/ext/tf-psa-crypto/0016-Do-not-define-__STDC_WANT_LIB_EXT1__-with-ATfE.patch) disables our `#define __STDC_WANT_LIB_EXT1__ 1`.

Goal of this issue: identify the problem and find a solution so that TF-M doesn't need to carry a patch.

bjwtaylor:
## Description

Add fix for memset_s issue discovered with ATFE resolves https://github.com/Mbed-TLS/TF-PSA-Crypto/issues/754. Adds a patch currently being used by TF-M. Open issues are do we want to add additional testing to cover their use cases and do we want to backport it? Let me know what people think?

## PR checklist

- [ ] **changelog** not required because: No public changes
- [ ] **framework PR** provided not required
- [ ] **TF-PSA-Crypto 1.1 PR** provided # | not required because: TBD
- [ ] **mbedtls 4.1 PR** not required because: No changes
- [ ] **mbedtls 3.6 PR** not required because: No changes
- **tests**  not required because: TBD

### DIFF
--- a/core/tf_psa_crypto_platform_requirements.h
+++ b/core/tf_psa_crypto_platform_requirements.h
@@ -15,10 +15,13 @@
 #ifndef TF_PSA_CRYPTO_TF_PSA_CRYPTO_PLATFORM_REQUIREMENTS_H
 #define TF_PSA_CRYPTO_TF_PSA_CRYPTO_PLATFORM_REQUIREMENTS_H
 
+/* Do not try to use the library extension with ATfE toolchain */
+#if !defined(__GNUC__) || !defined(__clang_major__)
 #ifndef __STDC_WANT_LIB_EXT1__
 /* Ask for the C11 gmtime_s() and memset_s() if available */
 #define __STDC_WANT_LIB_EXT1__ 1
 #endif
+#endif /* !__GNUC__ || !__clang_major__ */
 
 #if !defined(_POSIX_C_SOURCE)
 /* For standards-compliant access to


core/tf_psa_crypto_platform_requirements.h
```diff
@@ -15,10 +15,13 @@
 #ifndef TF_PSA_CRYPTO_TF_PSA_CRYPTO_PLATFORM_REQUIREMENTS_H
 #define TF_PSA_CRYPTO_TF_PSA_CRYPTO_PLATFORM_REQUIREMENTS_H
 
+/* Do not try to use the library extension with ATfE toolchain */
+#if !defined(__GNUC__) || !defined(__clang_major__)
```
gilles-peskine-arm, line 19: This is throwing the baby with the bathwater. We do want to enable `__STDC_WANT_LIB_EXT1__` so that we can use Annex K functions (`memset_s`, `gmtime_s`, …) where available. That inclues Clang with Glibc.

A conforming C implementation will not define `__STDC_LIB_EXT1__` in its headers if it doesn't provide the Annex K functions. This should be the case even for embedded platforms that don't have a full libc. Technically the fact that compilation fails is a bug in TF-M or in AtFE, but given that TF-M is not using an external libc, we may need to add a special case.

We need to identify why AtFE in TF-M has a problem, and either add a workaround for this specific case or tell TF-M how to fix their code or their compiler invocation.

gilles-peskine-arm (changes requested):
We need to fix an interoperability problem, not remove a feature.

core/tf_psa_crypto_platform_requirements.h
bjwtaylor: Apologies, I thought the plan was to verify the error and apply their fix. 

So the issue is caused because TF-M adds the option to the compiler -nostdlib  here https://github.com/TrustedFirmware-M/trusted-firmware-m/blob/c630c687aecaba792788b3d0cec17035981e0eb9/toolchain_ATFE.cmake#L134. If we do not want to accept their patch, they will have to change their build to something like this pattern:

`  clang --target=arm-none-eabi -mcpu=cortex-m33 -mthumb -nostartfiles \
   tf-psa-crypto_build/repro_zeroize.c \
    tf-psa-crypto_build/libc_stubs.c \
    f-psa-crypto_build/core/libtfpsacrypto.a \
    -Wl,-e,main \
    -Wl,--defsym=__heap_start=0x20000000 \
    -Wl,--defsym=__heap_end=0x20000000 \
    -o tf-psa-crypto_build/repro_zeroize_option1.axf
`
Though they may not be keen to do that, as I suspect this will increase the size of their binaries and they won't like that. Which is probably why they're using the option in the first place. We will also probably have to add a stub something like this:

`  void _exit(int status)
  {
      (void) status;
      for (;;) {
      }
  }
`
Let me know what you think?

gilles-peskine-arm: Why is there a problem only with AtFE, not with other toolchains?

bjwtaylor: So, we may have to work with tf-m to figure this out for sure as it appears to be a larger architectural problem with tf-m and only they will potentially know the exact reasoning behind their build config. However -nostdlib is gated on CONFIG_TFM_INCLUDE_STDLIBC , this is forced to be off when the compiler is clang here https://github.com/TrustedFirmware-M/trusted-firmware-m/blob/c630c687aecaba792788b3d0cec17035981e0eb9/config/check_config.cmake#L22 It appears based on this commit https://git.trustedfirmware.org/plugins/gitiles/TF-M/trusted-firmware-m/+/4537a2c584e1a496a215f37eecc22d37b1d9c22f that they are attempting to replace the stdlib with their own smaller more controlled implementation. This is being sourced from different places for different toolchains, which adds another level of variables and further explains the different behavior. However it looks like they are trying to add their own glibc stubs here https://github.com/TrustedFirmware-M/trusted-firmware-m/blob/c630c687aecaba792788b3d0cec17035981e0eb9/bl2/CMakeLists.txt#L162 and here https://github.com/TrustedFirmware-M/trusted-firmware-m/blob/main/bl2/src/mbedcrypto_stubs.c. However they have not added implementations for the functions that are currently causing the issue. So there's a few different ways they could fix it, though the exact solution will depend on their architectural objectives.

bjwtaylor: @gilles-peskine-arm, I've spoken to TF-M today and their preference is for use to disable the use of the ANNEX K functions in tf-psa-crypto. Though they would also reluctantly add further stubs to the existing  https://github.com/TrustedFirmware-M/trusted-firmware-m/blob/c630c687aecaba792788b3d0cec17035981e0eb9/bl2/CMakeLists.txt#L162 and here https://github.com/TrustedFirmware-M/trusted-firmware-m/blob/main/bl2/src/mbedcrypto_stubs.c. So I think it comes down to what we want to do, I guess we could have a conditional disable of the ANNEX K functions based on a config? Or we could just say the library requires ANNEX K and tf-m have to stub it, what do you think?

gilles-peskine-arm: TF-PSA-Crypto does not require annex K functions. It uses annex K functions if the system headers declare that annex K is available.

It's perfectly reasonable for TF-M to prefer not to implement `memset_s()`. But TF-PSA-Crypto already supports that. The problem is that in one specific build configuration, the TF-M platform (consisting of the TF-M build scripts, the TF-M code and the toolchain code) declares that `memset_()` is available, but does not provide it.

Why does this specific combination of project and compiler cause a discrepancy between system headers and actual function availability?

bjwtaylor: So, thanks of pointing this out, I hadn't realised we had this check. As far as I can tell the reason it fails because the check is a compile time check and the change that tf-m made keeps the compile time logic from atfe, which passes this check. They then change the linker to link in an implementation of libc which doesn't provide this function. So I guess they have further options of fixing the mismatch between compile time and link. Or we could just propose they use MBEDTLS_PLATFORM_ZEROIZE_ALT? Or we could maybe add a link based test, though this would probably not be an ideal for us. What do you think?

M-Moawad: Hi, I found another toolchain affected by this condition.

The Zephyr `benchmark.crypto.psa` test fails when built with the ARC MWDT toolchain:

```text
 error: call to undeclared function 'memset_s'
```

MWDT uses a Clang-based compiler, so `ccac` defines both macros checked by the current condition:

```text
 #define __clang_major__ 21
 #define __GNUC__ 4
 ```

This makes the condition false and prevents `__STDC_WANT_LIB_EXT1__` from being defined.

However, MWDT supports Annex K, defines `__STDC_LIB_EXT1__`, and provides `memset_s()`. Its `<string.h>` requires `__STDC_WANT_LIB_EXT1__` to expose the declaration. As a result, TF-PSA-Crypto selects `memset_s()` but the function is not declared.

MWDT provides the compiler-specific `__CCAC__` macro, so it could be handled separately:

```c
 #if !defined(__GNUC__) || !defined(__clang_major__) || defined(__CCAC__)
```

This keeps Annex K disabled for ATfE while enabling it for MWDT, where it is supported.

M-Moawad: I found this PR while preparing a TF-PSA-Crypto fix after the discussion in my Zephyr PR: https://github.com/zephyrproject-rtos/zephyr/pull/119764.

We concluded that the fix belongs in TF-PSA-Crypto rather than the Zephyr libc header.

PR #778, opened 2026-04-25
Fix ML-DSA signing error propagation

Nadav0077:
## Description

The ML-DSA PSA signing wrapper called mldsa-native to produce a signature, but
then overwrote the mldsa-native return value with success before returning.

This PR preserves the mldsa-native return value so internal signing failures are
reported through the PSA status code. It also maps mldsa-native allocation
failure to `PSA_ERROR_INSUFFICIENT_MEMORY`.

## PR checklist

- [x] **changelog** provided
- [x] **TF-PSA-Crypto development PR** provided #THIS_PR
- [x] **TF-PSA-Crypto 1.1 PR** not required because: ML-DSA support is only on the development branch
- [x] **mbedtls development PR** not required because: this PR only changes TF-PSA-Crypto source; Mbed TLS can update the submodule separately after this lands
- [x] **mbedtls 4.1 PR** not required because: this PR only changes TF-PSA-Crypto development
- [x] **mbedtls 3.6 PR** not required because: ML-DSA support is not present there
- **tests** not required because: the existing ML-DSA tests cover the normal signing path, and the failure path fixed here requires mldsa-native internal failure injection or custom allocation plumbing to trigger directly.

### DIFF
--- a/ChangeLog.d/mldsa-sign-error.txt
+++ b/ChangeLog.d/mldsa-sign-error.txt
@@ -1,3 +1,0 @@
-Bugfix
-   * Fix ML-DSA signing to report internal mldsa-native failures instead of
-     reporting success without a signature.
--- a/ChangeLog.d/mldsa-sign-error.txt
+++ b/ChangeLog.d/mldsa-sign-error.txt
@@ -1,0 +1,3 @@
+Bugfix
+   * Fix ML-DSA signing to report internal mldsa-native failures instead of
+     reporting success without a signature.
--- a/drivers/pqcp/src/psa_crypto_mldsa.c
+++ b/drivers/pqcp/src/psa_crypto_mldsa.c
@@ -27,18 +27,19 @@
 
 static psa_status_t pqcp_to_psa_error(int ret)
 {
-    /* At the time of writing, mldsa-native has very few documented error
-     * conditions: only invalid signature on verification, and self-test
-     * failure. But this will change when we update mldsa-native
-     * with support for heap allocation of intermediate values.
-     */
     if (ret == 0) {
         return PSA_SUCCESS;
     } else if (ret == MLD_ERR_OUT_OF_MEMORY) {
         return PSA_ERROR_INSUFFICIENT_MEMORY;
     } else {
-        /* Not really hardware, but this is the fallback error code for
-         * something unexpectedly going wrong in a driver. */
+        /* MLD_ERR_RNG_FAIL is intentionally not mapped: we don't install
+         * mldsa-native's RNG callback (mu is computed on our side for
+         * hedged ML-DSA), so it shouldn't be reachable. It would land
+         * here and be reported as a generic driver failure.
+         *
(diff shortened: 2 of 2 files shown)


drivers/pqcp/src/psa_crypto_mldsa.c
gilles-peskine-arm: I wrote this comment back when mldsa-native didn't define `MLD_ERR_xxx` constants (but they were being proposed) . This comment now is obsolete.

drivers/pqcp/src/psa_crypto_mldsa.c
```diff
@@ -34,6 +34,8 @@ static psa_status_t pqcp_to_psa_error(int ret)
      */
     if (ret == 0) {
         return PSA_SUCCESS;
+    } else if (ret == MLD_ERR_OUT_OF_MEMORY) {
+        return PSA_ERROR_INSUFFICIENT_MEMORY;
     } else {
```
gilles-peskine-arm, line 34: There's also `MLD_ERR_RNG_FAIL` nowadays. We haven't configured the RNG callback (I think we'll do the mu computation ourselves for hedged MLDSA, it's simpler overall), so we don't expect this error to ever appear. But it would be nice to mention in a comment that the omission is deliberate.

ChangeLog.d/mldsa-sign-error.txt
```diff
@@ -0,0 +1,3 @@
+Bugfix
```
gilles-peskine-arm: We haven't released MLDSA as a feature, so we shouldn't mention it in a changelog entry.

gilles-peskine-arm (changes requested):
Thanks for looking at our work in progress! The stray `ret = 0` is definitely problematic (although it has no consequences so far since the signature function can't fail at this stage). The error code update is just that this code is slightly behind upstream, we're going to use and test `MLD_ERR_OUT_OF_MEMORY` soon.

drivers/pqcp/src/psa_crypto_mldsa.c
Nadav0077: Added a comment in the fallback branch explaining MLD_ERR_RNG_FAIL is intentionally not mapped (no RNG callback installed, mu computed on our side). Let me know if you'd rather have it sit next to the OOM mapping.

Nadav0077 (commented):
Thanks for the review, Pushed a follow-up addressing the three points: dropped the changelog entry, removed the obsolete comment above the dispatcher, and added a note that MLD_ERR_RNG_FAIL is intentionally not mapped.

gilles-peskine-arm:
Please amend your commit(s) to add a sign-off line (`git commit --amend --signoff`). Otherwise I'm afraid we can't use them.

Nadav0077:
@gilles-peskine-arm Looks good?

gilles-peskine-arm (approved):
LGTM, thanks

bjwtaylor (approved):

PR #780, opened 2026-04-27
Mldsa functional tests

### LINKED ISSUE #641: Run mldsa-native functional tests in our CI
The goal of this issue is to run a small selection of functional tests of mldsa-native as part of `all.sh`. We only want coverage of functional testing on the platforms that we target.

What this means in terms of test coverage:

* Here, we're only testing mldsa-native, the way we _build_ it (compilation options, target platform, compiler flags), but not the way we _integrate_ it (we're not calling any of our integration code).
* We rely on the test coverage in mldsa-native for algorithmic tests, including edge cases.
* Maintainer tests (lint, portability, proofs, etc.) are out of scope here.

TBD: what about constant-time testing?

I'm still investigating, but at this stage I think what we need to do is to run [`scripts/tests`](https://github.com/pq-code-package/mldsa-native/blob/606e8ddf7be8a7284ad415fca464e4695fb19647/scripts/tests) with the right options, similar to the [`functest` action](https://github.com/pq-code-package/mldsa-native/blob/main/.github/actions/functest/action.yml).

Prerequisites: https://github.com/Mbed-TLS/TF-PSA-Crypto/issues/628

bjwtaylor:
## Description

Add mldsa functional tests to all.sh, resolves https://github.com/Mbed-TLS/TF-PSA-Crypto/issues/641

## PR checklist

- [ ] **changelog** not required because:  No functional changes.
- [ ] **TF-PSA-Crypto 1.1 PR** not required because:  No backports
- [ ] **mbedtls 4.1 PR** not required because: No changes
- [ ] **mbedtls 3.6 PR**not required because: No changes
- **tests**  provided

### DIFF
--- a/tests/scripts/components-pqcp.sh
+++ b/tests/scripts/components-pqcp.sh
@@ -16,11 +16,8 @@ support_test_pqcp_mldsa_native_upstream_all () {
 # This condition deliberately runs when all.sh sources this file. It hides the
 # component from both `--list-all-components` and explicit component requests on
 # runners whose Python is too old for mldsa-native's generated-test checks.
-if support_test_pqcp_mldsa_native_upstream_all; then
-    component_test_pqcp_mldsa_native_upstream_all () {
-        msg "test: mldsa-native upstream all --opt ALL --examples"
-
-        cd drivers/pqcp/mldsa-native
-        ./scripts/tests all --opt ALL --examples
-    }
-fi
+component_test_pqcp_mldsa_native_upstream_all () {
+    msg "test: mldsa-native upstream all --opt ALL --examples"
+     cd drivers/pqcp/mldsa-native
+    ./scripts/tests all --opt ALL --examples
+}
--- a/tests/scripts/components-pqcp.sh
+++ b/tests/scripts/components-pqcp.sh
@@ -16,5 +16,5 @@ support_test_pqcp_mldsa_native_upstream_all () {
 component_test_pqcp_mldsa_native_upstream_all () {
     msg "test: mldsa-native upstream all --opt ALL --examples --stack"
      cd drivers/pqcp/mldsa-native
-    ./scripts/tests all --opt ALL --examples --stack
+    ./scripts/tests all --opt ALL --examples --stack --alloc
 }
--- a/tests/scripts/components-pqcp.sh
+++ b/tests/scripts/components-pqcp.sh
@@ -15,6 +15,12 @@ support_test_pqcp_mldsa_native_upstream_all () {
 
 component_test_pqcp_mldsa_native_upstream_all () {
     msg "test: mldsa-native upstream all --opt ALL --examples --stack"
-     cd drivers/pqcp/mldsa-native
+    cd drivers/pqcp/mldsa-native
     ./scripts/tests all --opt ALL --examples --stack --alloc
+
+    make clean
+
+    msg "test: mldsa-native upstream all --opt ALL --stack with custom heap allocation"
+    ./scripts/tests all --opt ALL --stack --no-examples --no-alloc --no-rng-fail \
+        --cflags='-std=c11 -D_GNU_SOURCE -Itest -DMLD_CONFIG_FILE=\"configs/custom_heap_alloc_config.h\"'
 }
--- a/tests/scripts/components-pqcp.sh
+++ b/tests/scripts/components-pqcp.sh
@@ -13,9 +13,6 @@ support_test_pqcp_mldsa_native_upstream_all () {
     python3 -c 'import sys; exit(1 if sys.version_info < (3, 9) else 0)'
 }
 
-# This condition deliberately runs when all.sh sources this file. It hides the
-# component from both `--list-all-components` and explicit component requests on
-# runners whose Python is too old for mldsa-native's generated-test checks.
 component_test_pqcp_mldsa_native_upstream_all () {
     msg "test: mldsa-native upstream all --opt ALL --examples"
      cd drivers/pqcp/mldsa-native
--- a/tests/scripts/components-pqcp.sh
+++ b/tests/scripts/components-pqcp.sh
@@ -14,7 +14,7 @@ support_test_pqcp_mldsa_native_upstream_all () {
 }
 
 component_test_pqcp_mldsa_native_upstream_all () {
-    msg "test: mldsa-native upstream all --opt ALL --examples"
+    msg "test: mldsa-native upstream all --opt ALL --examples --stack"
      cd drivers/pqcp/mldsa-native
-    ./scripts/tests all --opt ALL --examples
+    ./scripts/tests all --opt ALL --examples --stack
 }
--- a/tests/scripts/components-pqcp.sh
+++ b/tests/scripts/components-pqcp.sh
@@ -1,0 +1,26 @@
+# components-pqcp.sh
+#
+# Copyright The Mbed TLS Contributors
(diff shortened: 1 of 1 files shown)


tests/scripts/components-pqcp.sh
```diff
@@ -0,0 +1,26 @@
...
+################################################################
+
+support_test_pqcp_mldsa_native_upstream_all () {
+    python3 -c 'import sys; exit(1 if sys.version_info < (3, 9) else 0)'
+}
+
+# This condition deliberately runs when all.sh sources this file. It hides the
+# component from both `--list-all-components` and explicit component requests on
+# runners whose Python is too old for mldsa-native's generated-test checks.
+if support_test_pqcp_mldsa_native_upstream_all; then
```
valeriosetti: Sure I'm missing something here, but is this `if` statement required?
I read the comment above that explains it, but as far as I remember the machinery that instruments our testing already does this. In support of my comment I don't see any other test `component_xxx` with an associated `support_xxx` with the `if` check.

valeriosetti (changes requested):

tests/scripts/components-pqcp.sh
bjwtaylor: So, it's been a while since I looked at this. However I think because the mldsa-native test cases require certain versions of python we need this to verify we are running that specific version and only run when we are.

valeriosetti: Yes, this is similar to what other `support_xxx` function do. IIRC the CI initially get a list of which `component_xxx` should be run on each specific ubuntu version, so there should be no need for extra logic.
However I might miss something so let's wait for the 2nd reviewer to confirm/deny my hypothesis.

gilles-peskine-arm: As Valerio says, this is exactly what `support_xxx` does. The `component_xxx` function should be defined unconditionally. It will only run on a platform where `support_xxx` is true. See for example `support_tf_psa_crypto_check_committed_generated_files` with the same code for the same reason.

bjwtaylor: OK, I'lll update it.

bjwtaylor: I've pushed the changes, lets see what the CI thinks.

bjwtaylor: CI is passing, so I think this one is good for re-review.

tests/scripts/components-pqcp.sh
```diff
@@ -0,0 +1,23 @@
...
+#### PQCP Testing
+################################################################
+
+support_test_pqcp_mldsa_native_upstream_all () {
+    python3 -c 'import sys; exit(1 if sys.version_info < (3, 9) else 0)'
+}
+
+# This condition deliberately runs when all.sh sources this file. It hides the
+# component from both `--list-all-components` and explicit component requests on
+# runners whose Python is too old for mldsa-native's generated-test checks.
```
gilles-peskine-arm: Obsolete comment, please remove it.

tests/scripts/components-pqcp.sh
```diff
@@ -0,0 +1,23 @@
...
+component_test_pqcp_mldsa_native_upstream_all () {
+    msg "test: mldsa-native upstream all --opt ALL --examples"
+     cd drivers/pqcp/mldsa-native
+    ./scripts/tests all --opt ALL --examples
```
gilles-peskine-arm: Do we care about examples?

Also, `--examples` is (currently) enabled by default.

gilles-peskine-arm: Should we pass `--stack` (off by default)? We do care about stack usage in some projects, so having the upstream information could be useful.

gilles-peskine-arm: We do want the alloc tests. `--alloc` is currently enabled by default, but we may want to make it explicit.

gilles-peskine-arm: Yes, we do care about examples. That's how mldsa-native tests itself with an external SHAKE. There's also an example that tests `MLD_CONFIG_REDUCE_RAM`.

tests/scripts/components-pqcp.sh
```diff
@@ -0,0 +1,23 @@
...
+################################################################
+
+support_test_pqcp_mldsa_native_upstream_all () {
+    python3 -c 'import sys; exit(1 if sys.version_info < (3, 9) else 0)'
+}
+
+# This condition deliberately runs when all.sh sources this file. It hides the
+# component from both `--list-all-components` and explicit component requests on
+# runners whose Python is too old for mldsa-native's generated-test checks.
+component_test_pqcp_mldsa_native_upstream_all () {
```
gilles-peskine-arm, line 16: This component tests mldsa-native in its default configuration, and with a set of variations, including an external SHAKE.

This set of variations does not cover one variant that we care about, which is using a custom memory allocator. So we should also test that. (Done via `.github/actions/config-variations/action.yml` in the mldsa-native CI.) We should (in the same component, given that it only takes a few minutes) run `scripts/tests all` with `--cflags='-DMLD_CONFIG_FILE "configs/custom_heap_alloc_config.h"'` (may require additional quoting) (plus the other options).

gilles-peskine-arm, line 16: Should we test other targets (architectures, compilers)? They already do upstream, but not with the same compilers, and anyway our goal here is to ensure that the exact version that we integrate is tested.

I think that we don't need to test additional targets that are already tested upstream. We should test IAR, which is not tested upstream, but that hinges on having IAR on the CI, which we currently don't. So that's a future action item.

gilles-peskine-arm (changes requested):
I've reviewed based on my analysis of the mldsa-native CI and on how we use it.

tests/scripts/components-pqcp.sh
bjwtaylor: OK, it's probably best to leave the --examples in then in case they disable it's default in the future? If so are you happy for this to be closed?

bjwtaylor: So, it is probably useful as it will monitor the stack usage which we are probably concerned about. I think it will increase CI time and will need valgrind, though if we don't have this already we can presumably add it. I've added the option let's see if the CI passes, or if it requires additional dependencies in the environment.

bjwtaylor: now added.

gilles-peskine-arm: We do have valgrind on the CI. Let's see how much load it adds.

tests/scripts/components-pqcp.sh
bjwtaylor: Now added

bjwtaylor: Yeah, I think a lot of this is already tested upstream. So we should focus on the things that really matter to us. So I'll leave these out, let me know if you really want them though.

tests/scripts/components-pqcp.sh
gilles-peskine-arm: The stack analysis takes about 3 second per build in https://ci.trustedfirmware.org/job/mbed-tls-tf-psa-crypto-multibranch/job/PR-780-head/10/cloudbees-pipeline-explorer/?filter=1299 . This is significantly less than the build (~30s). So no performance problem here.

gilles-peskine-arm (approved):
LGTM

valeriosetti (approved):

PR #793, opened 2026-05-17
Forward error from mldsa87_signature_internal in sign_message

dmitriy-bty:
## Summary

Commit [`21e93a5`](https://github.com/Mbed-TLS/TF-PSA-Crypto/commit/21e93a52) ("Make error detection more robust") converted `psa_crypto_mldsa.c` to forward `ret` from the internal mldsa-native functions into `pqcp_to_psa_error()`, so any future non-zero return becomes visible to PSA callers.

The conversion was applied correctly in `seed_to_public_key()` and at the `keypair_internal()` call in `tf_psa_crypto_mldsa_sign_message()`. At the `signature_internal()` call later in the same function, a bare `ret = 0;` was placed immediately after the call rather than inside a success-only branch, which would silently override any future non-zero return.

## Change

Mirror the existing pattern from the `keypair_internal()` call earlier in the same function (and from the entirety of `seed_to_public_key()`):

```c
     ret = tf_psa_crypto_pqcp_mldsa87_signature_internal(signature, ...);
+    if (ret != 0) {
+        goto cleanup;
+    }
     ret = 0;
```

## Tests

No new test. `signature_internal()` does not currently return errors — see the comment in `pqcp_to_psa_error()` and in `tf_psa_crypto_mldsa_verify_message()`. Adding a stub test for this would require wrapping or instrumenting the internal function, which seems out of proportion to a preparedness fix. Commit `21e93a5` itself added no tests for the same reason ("There are no errors to handle yet, but there will be in the future.").

## Changelog

Per `ChangeLog.d/00README.md`, no changelog entry: this is a not-yet-reachable code path being made future-safe, with no user-visible behavior change. Commit `21e93a5` similarly had none.

## Validation

- `git diff`: +3 lines, -0 lines, isolated to one function.
- New code is structurally identical to the `keypair_internal()` call's existing error path 12 lines above.

### DIFF
--- a/drivers/pqcp/src/psa_crypto_mldsa.c
+++ b/drivers/pqcp/src/psa_crypto_mldsa.c
@@ -145,6 +145,9 @@ psa_status_t tf_psa_crypto_mldsa_sign_message(
                                                         rnd,
                                                         secret,
                                                         0);
+    if (ret != 0) {
+        goto cleanup;
+    }
     ret = 0;
 
 cleanup:


gilles-peskine-arm:
Thanks! But we already have a pending fix in https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/778.

dmitriy-bty:
Closing — #778 covers the same bug with a broader fix (additionally maps `MLD_ERR_OUT_OF_MEMORY` to `PSA_ERROR_INSUFFICIENT_MEMORY` and removes the spurious `ret = 0;` rather than guarding it).

PR #809, opened 2026-06-11
Complete the move of driver wrappers to the dispatch/ directory

### LINKED ISSUE #833: Allow integrators to substitute their own dispatch layer
Some integrators of TF-PSA-Crypto want to plug in a set of drivers, and they want to maintain dispatch code for those drivers in a separate repository. Their dispatch code is C code, not generated using our script parametrized by a JSON description.

The goal of this issue is to allow integrators to substitute their own `dispatch` directory. This means we need to ensure that all of our the dispatch code is in the `dispatch` directory, and that directory contains no other code. Additionally, there should be a way to specify an alternative location instead of `dispatch` at build time.

Note that the dispatch code includes not only the files that are currently generated from jinja templates by `scripts/generate_driver_wrappers.py`, but also the multipart operation structure definitions in `include/psa/crypto_driver_contexts_*.h`.

asmellby:
## Description

The TF-PSA-Crypto 1.1 release introduced a `dispatch` directory with the following description:

> A new directory `dispatch` has been added at the root of the repository
     to eventually host all code that dispatches cryptographic operations to
     drivers, such as `psa_crypto_driver_wrappers_no_static.c`. For the time
     being, it only contains `psa_crypto_driver_wrappers_no_static.h`.

This PR completes the move of generated driver wrapper files and API headers to the dispatch directory.

Resolves https://github.com/Mbed-TLS/TF-PSA-Crypto/issues/833

## PR checklist

- [x] **changelog** provided
- [x] **framework PR** provided Mbed-TLS/mbedtls-framework#307
- [x] **TF-PSA-Crypto development PR** provided (this PR) 
- [x] **TF-PSA-Crypto 1.1 PR** not required because: not a bugfix, repository structure change is not entirely backwards compatible
- [x] **mbedtls development PR** provided Mbed-TLS/mbedtls#10780
- [x] **mbedtls 4.1 PR** not required because: does not affect mbedtls
- [x] **mbedtls 3.6 PR** not required because: does not affect mbedtls
- [x] **tests**  provided


core/CMakeLists.txt
```diff
@@ -154,6 +124,7 @@ foreach(target IN LISTS target_libraries)
         # so that generated headers in the build tree take precedence.
         PUBLIC $<BUILD_INTERFACE:${PROJECT_BINARY_DIR}/include/>
                $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/include/>
+               $<BUILD_INTERFACE:${TF_PSA_CRYPTO_DISPATCH_DIR}/include/>
```
gilles-peskine-arm, line 127: Please add a changelog entry file (see `ChangeLog.d/00README.md`) that briefly explains the user-visible aspect of the changes.

CMakeLists.txt
```diff
@@ -65,6 +65,10 @@ endif()
 # Set the project, Mbed TLS and framework root directory.
 set(TF_PSA_CRYPTO_FRAMEWORK_DIR ${CMAKE_CURRENT_SOURCE_DIR}/framework)
 
+if(NOT DEFINED TF_PSA_CRYPTO_DISPATCH_DIR)
```
gilles-peskine-arm, line 73: As I understand it, `TF_PSA_CRYPTO_DISPATCH_DIR` is now an option that can be set when running CMake. Please document it.

CMakeLists.txt
```diff
@@ -65,6 +65,10 @@ endif()
...
+    set(TF_PSA_CRYPTO_DISPATCH_DIR ${CMAKE_CURRENT_SOURCE_DIR}/dispatch)
```
gilles-peskine-arm, line 74: From prior experience, I think `CMakeLists.txt` in Mbed TLS will need to know about the new location to populate `TF_PSA_CRYPTO_PRIVATE_INCLUDE_DIRS`. It isn't properly extracting the information from TF-PSA-Crypto.

CMakeLists.txt
```diff
@@ -422,10 +428,13 @@ set(TF_PSA_CRYPTO_PRIVATE_INCLUDE_DIRS
...
 # constructed in `drivers/CMakeLists.txt`. This list may include a generated
 # test driver.
+# `dispatch` must be added after `drivers` because the dispatch implementation
+# must know of all drivers.
 #
 add_subdirectory(framework)
 add_subdirectory(include)
-add_subdirectory(drivers)   # must precede core
+add_subdirectory(drivers)   # must precede dispatch and core
+add_subdirectory(${TF_PSA_CRYPTO_DISPATCH_DIR})
```
gilles-peskine-arm: My understanding is that the point of having `TF_PSA_CRYPTO_DISPATCH_DIR` as a variable is that users can pass an out-of-tree directory to be used instead of `dispatch`, but it doesn't work.

```
$ cmake -DTF_PSA_CRYPTO_DISPATCH_DIR=/tmp/custom-dispatch-56449b86ac563df1daa921a7a929d522ecfc8256 -B build-customdispatch-cmake
-- The C compiler identification is GNU 13.3.0
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /usr/bin/cc - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Found Python3: /home/gilpes01/Packages/_/x86.64_Linux/bin/python3 (found version "3.12.3") found components: Interpreter 
-- Performing Test C_COMPILER_SUPPORTS_WFORMAT_SIGNEDNESS
-- Performing Test C_COMPILER_SUPPORTS_WFORMAT_SIGNEDNESS - Success
CMake Error at CMakeLists.txt:437 (add_subdirectory):
  add_subdirectory not given a binary directory but the given source
  directory "/tmp/custom-dispatch-56449b86ac563df1daa921a7a929d522ecfc8256"
  is not a subdirectory of
  "/tmp/reviewing/mbedtls+tf-psa-crypto-809-feature%dispatch-library/tf-psa-crypto".
  When specifying an out-of-tree source a binary directory must be explicitly
  specified.


-- Configuring incomplete, errors occurred!
```

1. Please either fix, or clarify the intended semantics.
2. Please add a test in `tests/scripts/components-build-system.sh` that demonstrates the intended feature.

dispatch/CMakeLists.txt
```diff
@@ -0,0 +1,46 @@
...
+            ${TF_PSA_CRYPTO_PYTHON_EXECUTABLE}
+                ${PROJECT_SOURCE_DIR}/scripts/generate_driver_wrappers.py
+                ${CMAKE_CURRENT_BINARY_DIR}
+        DEPENDS
+            ${PROJECT_SOURCE_DIR}/scripts/generate_driver_wrappers.py
+            ${PROJECT_SOURCE_DIR}/scripts/data_files/driver_templates/psa_crypto_driver_wrappers.h.jinja
+            ${PROJECT_SOURCE_DIR}/scripts/data_files/driver_templates/psa_crypto_driver_wrappers_no_static.c.jinja
+    )
+
+    add_custom_target(
```
gilles-peskine-arm: I asked Codex to review and it tells me:

> The generated wrapper rule can run twice in parallel CMake Makefile builds
> 
> `dispatch/CMakeLists.txt` creates a custom target for the generated wrapper
> files, and also lists the same generated outputs as sources of the dispatch
> object library. In a clean Makefile build:
> 
> ```sh
> cmake -S . -B /tmp/tf-psa-review-default \
>     -DENABLE_TESTING=OFF \
>     -DENABLE_PROGRAMS=OFF
> cmake --build /tmp/tf-psa-review-default --parallel 16
> ```
> 
> the output showed `generate_driver_wrappers.py` running twice for the same
> `dispatch/psa_crypto_driver_wrappers.*` outputs:
> 
> ```text
> Generating psa_crypto_driver_wrappers.h, psa_crypto_driver_wrappers_no_static.c
> Generating psa_crypto_driver_wrappers.h, psa_crypto_driver_wrappers_no_static.c
> ```
> 
> That is a race risk under recursive Makefile generation. The generated files
> should have a single owning custom command path, with other targets depending on
> that owner rather than independently triggering the same outputs.

This looks superficially legitimate to me, but I don't know much about CMake.

dispatch/include/psa/crypto_driver_contexts_composites.h
gilles-peskine-arm, line 1: There's a conflict because the file moved in this branch and changed in `development`. Modern git versions know how to resolve this, but not the ancient git that GitHub uses. Please rebase your branch on top of `development`, but keep the same commit structure and do rework in additional commits, so that `git range-diff development 56449b86ac563df1daa921a7a929d522ecfc8256 the_sha_after_rebase` remains simple.

gilles-peskine-arm (changes requested):
I think this is moving in the direction we want, thanks. But I've asked Ronald to review as well, as he's the architect of the directory structure and build scripts. We're currently focusing on some security fixes for our next release, so it might be a while before he can review.

I'm not sure what users are supposed to be able to do. Whatever it is, it needs to be documented and tested.

I reviewed 56449b86ac563df1daa921a7a929d522ecfc8256

core/CMakeLists.txt
asmellby: Added a changelog entry.

CMakeLists.txt
asmellby: Added documentation to the top of CMakeLists.txt, where the existing CMake variables were documented.

CMakeLists.txt
asmellby: Good catch! Fixed and added tests.

dispatch/CMakeLists.txt
asmellby: Fixed!

dispatch/include/psa/crypto_driver_contexts_composites.h
asmellby: Done. Rebased the original commits and added new commits addressing review feedback.

CMakeLists.txt
```diff
@@ -14,6 +14,11 @@
 #   calling CMake in order to avoid target name clashes, via the use of
 #   TF_PSA_CRYPTO_TARGET_PREFIX. The value of this variable is prefixed to the
 #   tfpsacrypto and tfpsacrypto-apidoc targets.
+# - TF_PSA_CRYPTO_DISPATCH_DIR: If set, this directory is used instead of the
```
gilles-peskine-arm: Please state that this must be an absolute path.

gilles-peskine-arm (changes requested):
Looks good to me, except I'd like one point clarified in the documentation. I don't know much about CMake, so my review of CMake code is superficial.

Our CI appears to be stuck today (ticket filed), so results might not be available until Monday.

gilles-peskine-arm:
Note that the CI isn't happy. There's at least a missing include directory in CMake builds.

CMakeLists.txt
asmellby: Opened Mbed-TLS/mbedtls#10780

CMakeLists.txt
asmellby: Done.

valeriosetti:
I manually kicked the CI because it didn't automatically start after the last push

ronald-cron-arm:
@asmellby Sorry for the delay.

I've started looking at this PR. It seems to me that it addresses two separate things:

1. Completing the move of the generated driver wrapper files and API headers to the `dispatch` directory.
2. Adding support for selecting a dispatch directory different from the repository's `dispatch` directory.

I would prefer to handle (1) first and (2) in a follow-up PR. Would you be able to split the changes accordingly?

asmellby:
> I would prefer to handle (1) first and (2) in a follow-up PR. Would you be able to split the changes accordingly?

I could do that, but that will require a rebase, which the contribution guidelines tell me not to do without approval from reviewers. So I'll need some guidance: To what degree should I clean up the commits when I rebase? Am I allowed to merge fixup commits into the initial commits to create a more reviewable changeset, or should I keep them separate to preserve the (now no longer entirely real) history?

gilles-peskine-arm:
I would prefer to keep (1) and (2) in the same pull request, because (2) checks that (1) has been done right.

But if (1) and (2) are separate, please provide histories that help me review incrementally:

* A history with the same commit structure, so I can review with `git range-diff development 1c6cde11420cf17340a385dfb43b9453dc211c44 head1`.
* A restructured history with a step `step1` in one PR, and a step `step2` such that `head1` and `step2` have the same content.

(Where rebases get really bad is if there's both a change of base and a change of commit structure.)

gilles-peskine-arm:
By the way, for reviewing, the merge conflict on the framework submodule is not a concern. However, I haven't made this review a priority due to the CI failure.

Note, I will be offline from tomorrow to 19 July.

asmellby:
> By the way, for reviewing, the merge conflict on the framework submodule is not a concern. However, I haven't made this review a priority due to the CI failure.

CI passes for the TF-PSA-Crypto tests as far as I can tell. The CI failure is on the Mbed TLS tests, which needs the corresponding change in https://github.com/Mbed-TLS/mbedtls/pull/10780 to pass. How can I tell the TF-PSA-Crypto CI to use that Mbed TLS PR?

ronald-cron-arm:
> > By the way, for reviewing, the merge conflict on the framework submodule is not a concern. However, I haven't made this review a priority due to the CI failure.
> 
> CI passes for the TF-PSA-Crypto tests as far as I can tell. The CI failure is on the Mbed TLS tests, which needs the corresponding change in [Mbed-TLS/mbedtls#10780](https://github.com/Mbed-TLS/mbedtls/pull/10780) to pass. How can I tell the TF-PSA-Crypto CI to use that Mbed TLS PR?

You cannot. We will need to merge the Mbed TLS PR first.

ronald-cron-arm:
> I would prefer to keep (1) and (2) in the same pull request, because (2) checks that (1) has been done right.

@gilles-peskine-arm I am not sure what do you mean by that. Could you elaborate?

gilles-peskine-arm:
If we've moved too few files or too many files, the test with an alternative dispatch directory has a good change of telling us. Without this test, nothing will tell us.

ronald-cron-arm:
> If we've moved too few files or too many files, the test with an alternative dispatch directory has a good change of telling us. Without this test, nothing will tell us.

The tests with an alternative dispatch directory mostly reproduce the content of the `dispatch` directory into a `custom` directory, so I am not sure how they would help check whether we have moved too few or too many files into that directory.

More generally, I may have missed some discussion, but I am not sure what we are trying to achieve with the `TF_PSA_CRYPTO_DISPATCH_DIR` option. What are the intended use cases for this option?

gilles-peskine-arm:
> The tests with an alternative dispatch directory mostly reproduce the content of the dispatch directory into a custom directory, so I am not sure how they would help check whether we have moved too few or too many files into that directory.

Ah, yes, they wouldn't really help with that, my bad.

> More generally, I may have missed some discussion, but I am not sure what we are trying to achieve with the TF_PSA_CRYPTO_DISPATCH_DIR option. What are the intended use cases for this option?

Ah, my bad, I've mentioned it before but I hadn't put it in writing. I've filed https://github.com/Mbed-TLS/TF-PSA-Crypto/issues/833.

dispatch/CMakeLists.txt
```diff
@@ -25,9 +25,21 @@ if(GEN_FILES)
         "${CMAKE_CURRENT_BINARY_DIR}/psa_crypto_driver_wrappers_no_static.c"
     )
 else()
-    list(APPEND dispatch_src_files
-        "${CMAKE_CURRENT_SOURCE_DIR}/psa_crypto_driver_wrappers_no_static.c"
-    )
+    if(EXISTS "${CMAKE_CURRENT_BINARY_DIR}/psa_crypto_driver_wrappers_no_static.c" AND NOT
+       EXISTS "${CMAKE_CURRENT_SOURCE_DIR}/psa_crypto_driver_wrappers_no_static.c")
+        # Some test cases (tf_psa_crypto_test_memsan_constant_flow_psa) rely on
```
gilles-peskine-arm, line 30: Not “test cases” (this would be a stanza in `tests/suites/*.data`, but “test components”.

But it's weird to justify a feature by a test component. I don't quite understand why this is needed.

dispatch/CMakeLists.txt
```diff
@@ -25,9 +25,21 @@ if(GEN_FILES)
         "${CMAKE_CURRENT_BINARY_DIR}/psa_crypto_driver_wrappers_no_static.c"
     )
 else()
-    list(APPEND dispatch_src_files
-        "${CMAKE_CURRENT_SOURCE_DIR}/psa_crypto_driver_wrappers_no_static.c"
-    )
+    if(EXISTS "${CMAKE_CURRENT_BINARY_DIR}/psa_crypto_driver_wrappers_no_static.c" AND NOT
+       EXISTS "${CMAKE_CURRENT_SOURCE_DIR}/psa_crypto_driver_wrappers_no_static.c")
```
gilles-peskine-arm, line 29: This should be an elseif.

gilles-peskine-arm, line 29: I don't understand why this condition is needed now. Is it fixing a preexisting bug?

We run our CI with `GEN_FILES=ON`. It's supposed to work with `GEN_FILES=OFF`, but that's not something we test.

gilles-peskine-arm (changes requested):
At 7655efb69f881912911083c8ff96650f9961f3cb, LGTM except that I don't understand 2de8d4e61c9d9fbb60461fb3ed578077191b3e65.

dispatch/CMakeLists.txt
asmellby: The test component `tf_psa_crypto_test_memsan_constant_flow_psa` runs CMake twice -- first with `GEN_FILES=ON` to generate the driver wrapper. Then, it runs again with `GEN_FILES=OFF` and `CMAKE_BUILD_TYPE=MemSan`, relying on the previous invocation's copy of the generated driver wrapper to still exist and be picked up by the build system.

Introduced in https://github.com/Mbed-TLS/TF-PSA-Crypto/commit/93a811427694293d75d032878d6c8384e544b5ce

This creates a weird scenario where the build system on that second pass needs to be able to pick up the file from the generation directory, despite `GEN_FILES=OFF`. The previous implementation in `core` handled this by using a relative path, which makes CMake check both the source and build directory: https://github.com/Mbed-TLS/TF-PSA-Crypto/blob/development/core/CMakeLists.txt#L3-L14
I can simplify this to match the way it was done in `core` before, they should be equivalent if my reading of https://cmake.org/cmake/help/v3.31/prop_tgt/SOURCES.html is correct -- what I did is just more explicit about the intention.

asmellby: https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/135#discussion_r1946824568 contains some discussion of why it's done this way for that test component.

dispatch/CMakeLists.txt
asmellby: Agree that it's weird. Added a comment below about why it's needed, but that it could be made less explicit by relying on implicit CMake search paths for relative paths, which was what the core did before. That would avoid explicitly calling out the test component -- the build system behavior would be exactly the same, though.

dispatch/CMakeLists.txt
gilles-peskine-arm: Thanks for the explanation.

I'm ok with this code if you improve the comments explaining why, mainly that it's about `GEN_FILES=ON` followed by `GEN_FILES=OFF`. I don't understand why we did that in 93a811427694293d75d032878d6c8384e544b5ce, as opposed to e.g. `framework/scripts/make_generated_files.py` or just directly running the build without touching `GEN_FILES`.

But I'm a CMake novice. Please wait for someone who is more experienced with CMake (Ronald) to give their opinion.

tests/scripts/components-build-system.sh
```diff
@@ -41,6 +41,18 @@ component_test_tf_psa_crypto_as_subdirectory () {
...
+component_test_tf_psa_crypto_as_subdirectory_custom_dispatch () {
+    msg "build: custom dispatch with cmake 'as-subdirectory' build"
+    cp -r dispatch/include programs/test/cmake_subproject_custom_dispatch/custom
+    cp dispatch/psa_crypto_driver_wrappers_no_static.h programs/test/cmake_subproject_custom_dispatch/custom
+    scripts/generate_driver_wrappers.py programs/test/cmake_subproject_custom_dispatch/custom
+    cd programs/test/cmake_subproject_custom_dispatch
+    # Note: Explicitly generate files as these are turned off in releases
+    cmake -D GEN_FILES=ON .
+    make
+    ./cmake_subproject_custom_dispatch
```
ronald-cron-arm:
```suggestion
    cp -r dispatch/. programs/test/cmake_subproject_custom_dispatch/custom/
    cp programs/test/cmake_subproject_custom_dispatch/custom.CMakeLists.txt \
         programs/test/cmake_subproject_custom_dispatch/custom/CMakeLists.txt
    scripts/generate_driver_wrappers.py programs/test/cmake_subproject_custom_dispatch/custom
    cd programs/test/cmake_subproject_custom_dispatch
    # Note: Explicitly generate files as these are turned off in releases
    cmake -D GEN_FILES=ON .
    make
    ./cmake_subproject_custom_dispatch
```
to minimize the knowledge of dispatch internals?

tests/scripts/components-build-system.sh
```diff
@@ -41,6 +41,18 @@ component_test_tf_psa_crypto_as_subdirectory () {
...
+}
```
ronald-cron-arm: Could we minimize this test’s knowledge of the dispatch directory? For example, copy the complete dispatch directory and then replace its CMakeLists.txt with the custom one:
```suggestion
component_test_tf_psa_crypto_as_subdirectory_custom_dispatch () {               
    msg "build: custom dispatch with cmake 'as-subdirectory' build"             
    cp -r dispatch/. programs/test/cmake_subproject_custom_dispatch/custom/     
    cp programs/test/cmake_subproject_custom_dispatch/custom.CMakeLists.txt \   
        programs/test/cmake_subproject_custom_dispatch/custom/CMakeLists.txt    
    scripts/generate_driver_wrappers.py programs/test/cmake_subproject_custom_dispatch/custom
    cd programs/test/cmake_subproject_custom_dispatch                           
    # Generate the remaining project-owned files as needed.                     
    cmake -D GEN_FILES=ON .                                                     
    make                                                                        
    ./cmake_subproject_custom_dispatch                                          
} 
```

tests/scripts/components-build-system.sh
```diff
@@ -67,6 +79,18 @@ component_test_tf_psa_crypto_cmake_as_package_install () {
     ./cmake_package_install
 }
 
+component_test_tf_psa_crypto_cmake_as_package_install_custom_dispatch () {
```
ronald-cron-arm: ditto

PR #811, opened 2026-06-12
Inconsistent return value of mbedtls_pk_parse_key breaks mbedtls_pk_check_pair sometimes

### LINKED ISSUE #807: Inconsistent return value of mbedtls_pk_parse_key breaks mbedtls_pk_check_pair sometimes
### Summary

In commit 6e237d5ffb12cd5e776c3109d197e03e68424f50 mbedtls_pk_check_pair was changed to require the prv key to contain a public key. But this breaks common usages like with mbedtls_pk_parse_keyfile as the change done to the test suite shows where the author added a call to mbedtls_pk_set_pubkey_from_prv to compensate.

I ran into the same problem but I do not have the option to call mbedtls_pk_set_pubkey_from_prv since that is internal.

The problem seems to be that mbedtls_pk_parse_key only calls mbedtls_pk_set_pubkey_from_prv _sometimes_.
In my case it seems to use the mbedtls_pk_parse_key_pkcs8_unencrypted_der path and it doesn't call mbedtls_pk_set_pubkey_from_prv. In other paths it does.

### System information

TF-PSA-Crypto version (number or commit id): v1.1.0
Operating system and version: Ubuntu 24.04
Configuration (if not default, please attach `crypto_config.h`): default 
Compiler and options (if you used a pre-built binary, please indicate how you obtained it): GCC, but also reproducible with clang 
Additional environment information:
This is from the OpenVPN code. An example failure can be seen in our Github actions:  https://github.com/flichtenheld/openvpn/actions/runs/27137270954

### Expected behavior

I think due to the change of check_pair it is important that parse_key now always returns prv in a state where the public key has been extracted. At the very least the PKCS8 case needs to be fixed, but there might be other cases affected.

### Actual behavior

Private key is returned from parse_key without the public key information set.

### Steps to reproduce

Just remove the mbedtls_pk_set_pubkey_from_prv call from the test suite, I assume? It really shouldn't be there and instead the parse functions fixed so that the test suite passes without that work-around.

If that is not enough to see the error I can provide reproduction based on the OpenVPN test suite. Let me know if that is required.

valeriosetti:
## Description

Resolves #807

## PR checklist

- [x] **changelog** provided
- [x] **TF-PSA-Crypto development PR** provided: this one
- [x] **TF-PSA-Crypto 1.1 PR** provided: https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/853
- [x] **mbedtls 4.1 PR** not required because: no change there
- **tests**  provided: the PR first changes tests removing calls to `mbedtls_pk_set_pubkey_from_prv`. This cause some test case to start failing. Rhen the fix is applied to `pkparse.c` to resolve the problem. No new test is added because I think the already existing ones are enough.

### DIFF
--- a/extras/pkparse.c
+++ b/extras/pkparse.c
@@ -796,9 +796,8 @@ MBEDTLS_STATIC_TESTABLE int mbedtls_pk_parse_key_pkcs8_unencrypted_der(
 
 #if defined(PSA_WANT_KEY_TYPE_RSA_PUBLIC_KEY)
     if (pk_alg == MBEDTLS_PK_RSA) {
-        ret = mbedtls_pk_rsa_set_key(pk, p, len) ||
-              mbedtls_pk_set_pubkey_from_prv(pk);
-        if (ret != 0) {
+        if (((ret = mbedtls_pk_rsa_set_key(pk, p, len)) != 0) ||
+            ((ret = mbedtls_pk_set_pubkey_from_prv(pk)) != 0)) {
             mbedtls_pk_free(pk);
             return ret;
         }
--- a/extras/pkparse.c
+++ b/extras/pkparse.c
@@ -796,7 +796,9 @@ MBEDTLS_STATIC_TESTABLE int mbedtls_pk_parse_key_pkcs8_unencrypted_der(
 
 #if defined(PSA_WANT_KEY_TYPE_RSA_PUBLIC_KEY)
     if (pk_alg == MBEDTLS_PK_RSA) {
-        if ((ret = mbedtls_pk_rsa_set_key(pk, p, len)) != 0) {
+        ret = mbedtls_pk_rsa_set_key(pk, p, len) ||
+              mbedtls_pk_set_pubkey_from_prv(pk);
+        if (ret != 0) {
             mbedtls_pk_free(pk);
             return ret;
         }
--- a/ChangeLog.d/pkcs8-rsa-public-key-parsing.txt
+++ b/ChangeLog.d/pkcs8-rsa-public-key-parsing.txt
@@ -1,0 +1,5 @@
+Bugfix
+   * Fixed a PKCS#8 bug where the public part of RSA key was not correctly
+     added to the PK context when the key was parsed. This could cause
+     other unexpected failures when using that PK context for other operations.
+     Fixes #807.
--- a/tests/suites/test_suite_pkparse.function
+++ b/tests/suites/test_suite_pkparse.function
@@ -131,8 +131,6 @@ void pk_parse_keyfile_rsa(char *key_file, char *password, int result)
         TEST_EQUAL(mbedtls_pk_get_bitlen(&ctx), ctx.bits);
         TEST_EQUAL(mbedtls_pk_get_len(&ctx), PSA_BITS_TO_BYTES(ctx.bits));
 
-        TEST_ASSERT(mbedtls_pk_set_pubkey_from_prv(&ctx) == 0);
-
 #if defined(MBEDTLS_PSA_CRYPTO_C)
         TEST_ASSERT(test_psa_bridge(&ctx, PSA_KEY_USAGE_SIGN_HASH));
         TEST_ASSERT(test_psa_bridge(&ctx, PSA_KEY_USAGE_SIGN_MESSAGE));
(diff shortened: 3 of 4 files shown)


gstrauss:
@valeriosetti should `pk_parse_key_rfc8410_der` check `if (pk->pub_raw_len == 0)` before redoing mbedtls_pk_set_pubkey_from_prv()?  From `pk_parse_key_rfc8410_der`:
```
    /* mbedtls_pk_parse_key_pkcs8_unencrypted_der() only supports version 1
     * PKCS8 keys, which never contain a public key. As such, derive the public
     * key unconditionally. */
    if ((ret = mbedtls_pk_set_pubkey_from_prv(pk)) != 0) {
        return ret;
    }
```

valeriosetti:
> @valeriosetti should `pk_parse_key_rfc8410_der` check `if (pk->pub_raw_len == 0)` before redoing mbedtls_pk_set_pubkey_from_prv()? From `pk_parse_key_rfc8410_der`:
> 
> ```
>     /* mbedtls_pk_parse_key_pkcs8_unencrypted_der() only supports version 1
>      * PKCS8 keys, which never contain a public key. As such, derive the public
>      * key unconditionally. */
>     if ((ret = mbedtls_pk_set_pubkey_from_prv(pk)) != 0) {
>         return ret;
>     }
> ```

Sorry, but I'm not sure I get it. According to the comment the public key never exist in PKCS8 v1, so we have to always derive it for sure. Having `pk->pub_raw_len != 0` would mean that the public part has already been filled somewhere else, but how can that happen?

Anyway it seems to me that the check is [already present](https://github.com/Mbed-TLS/TF-PSA-Crypto/blob/0ef26791ceb1445b0563c40b8e2c073000245a7d/extras/pk.c#L146) in `mbedtls_pk_set_pubkey_from_prv`, isn't it?

extras/pkparse.c
```diff
@@ -796,7 +796,9 @@ MBEDTLS_STATIC_TESTABLE int mbedtls_pk_parse_key_pkcs8_unencrypted_der(
 
 #if defined(PSA_WANT_KEY_TYPE_RSA_PUBLIC_KEY)
     if (pk_alg == MBEDTLS_PK_RSA) {
-        if ((ret = mbedtls_pk_rsa_set_key(pk, p, len)) != 0) {
+        ret = mbedtls_pk_rsa_set_key(pk, p, len) ||
+              mbedtls_pk_set_pubkey_from_prv(pk);
```
bjwtaylor: Should this be 
```
  if ((ret = mbedtls_pk_rsa_set_key(pk, p, len)) != 0 ||
      (ret = mbedtls_pk_set_pubkey_from_prv(pk)) != 0) {
      mbedtls_pk_free(pk);
      return ret;
  }
```
As currently will is loose the error code? Or is this intentional?

bjwtaylor (commented):
This looks like a good fix. Just one question/comment.

gstrauss:
> > @valeriosetti should `pk_parse_key_rfc8410_der` check `if (pk->pub_raw_len == 0)` before redoing mbedtls_pk_set_pubkey_from_prv()? From `pk_parse_key_rfc8410_der`:
> > ```
> >     /* mbedtls_pk_parse_key_pkcs8_unencrypted_der() only supports version 1
> >      * PKCS8 keys, which never contain a public key. As such, derive the public
> >      * key unconditionally. */
> >     if ((ret = mbedtls_pk_set_pubkey_from_prv(pk)) != 0) {
> >         return ret;
> >     }
> > ```
> 
> Sorry, but I'm not sure I get it. According to the comment the public key never exist in PKCS8 v1, so we have to always derive it for sure. Having `pk->pub_raw_len != 0` would mean that the public part has already been filled somewhere else, but how can that happen?
> 
> Anyway it seems to me that the check is [already present](https://github.com/Mbed-TLS/TF-PSA-Crypto/blob/0ef26791ceb1445b0563c40b8e2c073000245a7d/extras/pk.c#L146) in `mbedtls_pk_set_pubkey_from_prv`, isn't it?

Yes, sorry.  My mistake.  Your patch is also in another branch of code `mbedtls_pk_parse_key_pkcs8_unencrypted_der()`.  I had misread the code and thought there was a chance that `mbedtls_pk_set_pubkey_from_prv()` might be called more than once, and yes, `mbedtls_pk_set_pubkey_from_prv()` checks that.

extras/pkparse.c
valeriosetti: Yes, sorry for the mistake.

bjwtaylor (approved):

yiwu0b11 (approved):
Code change LGTM, thanks.

gilles-peskine-arm:
Please make a 1.1 backport

PR #824, opened 2026-06-28
Fix ML-DSA key type value

### LINKED ISSUE #823: Incorrrect value of PSA_KEY_TYPE_ML_DSA_PUBLIC_KEY
### Summary
The value is inconsistent with the PSA crypto spec: https://arm-software.github.io/psa-api/crypto/1.5/IHI0086-PSA_Certified_Crypto_API-1.5.0.pdf

### System information

TF-PSA-Crypto version (number or commit id): c35df962041731b08c6ff6a39a59e239c344e1ee

### Expected behavior
drivers/pqcp/src/psa_crypto_mldsa.h
#define PSA_KEY_TYPE_ML_DSA_PUBLIC_KEY ((psa_key_type_t) **0x4002**)

### Actual behavior
drivers/pqcp/src/psa_crypto_mldsa.h
#define PSA_KEY_TYPE_ML_DSA_PUBLIC_KEY ((psa_key_type_t) **0x6002**)

### LINKED ISSUE #7990: Remove mentions of DSA from Mbed TLS

gilles-peskine-arm:
Fixes https://github.com/Mbed-TLS/TF-PSA-Crypto/issues/823. Add a metadata test that would have caught something.

Also remove finite-field DSA constants, both to clean things up and to make room in flag bitmasks. Subsumes https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/321. Resolves https://github.com/Mbed-TLS/mbedtls/issues/7990.

## PR checklist

- [x] **changelog** not required because: ML-DSA is not a public feature yet
- [x] **TF-PSA-Crypto development PR** provided
- [x] **TF-PSA-Crypto 1.1 PR** provided # | not required because: ML-DSA in 1.1 has not reached a viability threshold, is not reachable in any way from public interfaces, and should just be removed; for DSA, we don't remove compile-time-only fluff in an LTS
- [x] **mbedtls 3.6 PR** not required because: no ML-DSA in 3.6; for DSA, we don't remove compile-time-only fluff in an LTS


bjwtaylor (approved):

tests/suites/test_suite_psa_crypto_metadata.function
```diff
@@ -11,6 +11,15 @@
 #include "psa/crypto.h"
 #include "psa_crypto_invasive.h"
 
+/* While we have an ML-DSA driver but no API support yet */
+#include "psa_crypto_mldsa.h"
+#if !defined(PSA_ALG_HASH_ML_DSA)
+/* While we the library doesn't support anything about HashML-DSA yet */
```
valeriosetti, line 17:
```suggestion
/* While the library doesn't support anything about HashML-DSA yet */
```

valeriosetti (approved):

tests/suites/test_suite_psa_crypto_metadata.function
gilles-peskine-arm: I agree, but let's not spend a rework-ci-review cycle just for that.

PR #834, opened 2026-07-08
ML-DSA driver: support the expanded private key format

### LINKED ISSUE #774: Support expanded key format in ML-DSA signature driver
In the ML-DSA signature driver, in addition to supporting the standard private key representation (just the seed), support a custom format that contains the expanded private key and the expanded public key.

We have a short-term business case for this through the multipart API, but in the medium term we also want it through the one-shot API, so we might as well do both.

Test this both through direct calls to the driver and through dispatch. (No library change is expected for dispatch.)

Prerequisites: https://github.com/Mbed-TLS/TF-PSA-Crypto/issues/635

Enhances: https://github.com/Mbed-TLS/TF-PSA-Crypto/issues/771 (when we have both dispatch-to-multipart and expanded key support, test dispatch to multipart with expanded key).

Follow-ups: the core actually taking advantage of this. We don't have a full architectural design yet. Interface discussion in https://github.com/ARM-software/psa-api/issues/308.

gilles-peskine-arm:
In the ML-DSA driver, support export-public and signing where the key input is the concatenation of the seed and the expanded private key, in addition to supporting the seed only. Resolves https://github.com/Mbed-TLS/TF-PSA-Crypto/issues/774.

Needs preceding PR: framework prerequisite https://github.com/Mbed-TLS/mbedtls-framework/pull/312

## PR checklist

- [x] **changelog** not required because: not exposed in the API yet
- [x] **framework PR** provided https://github.com/Mbed-TLS/mbedtls-framework/pull/312
- [x] **TF-PSA-Crypto 1.1 PR** not required because: new feature
- [x] **mbedtls development PR** not required because: new feature
- [x] **mbedtls 4.1 PR** not required because: new feature
- **tests**  provided


drivers/pqcp/src/psa_crypto_mldsa.c
bjwtaylor, line 221: The PR description says this PR adds support for export-public and signing where the key input is the concatenation of the seed and the expanded private key. However it looks like this function still calls to seed_to_public_key and doesn't accept the proposed new key input and the other changes seem related to signing? So is the description out of date, the code needs additional functionality or have I missed something?

bjwtaylor (commented):
One question on the scope, though otherwise looks good.

drivers/pqcp/src/psa_crypto_mldsa.c
gilles-peskine-arm: Oh, oops. That's what you get for working on a feature one hour at a time... I'll do it when I get back, either here or in a follow-up.

drivers/pqcp/src/psa_crypto_mldsa.c
```diff
@@ -105,12 +112,14 @@ MBEDTLS_STATIC_ASSERT(
     TF_PSA_CRYPTO_MLDSA_SIGNATURE_MAX_SIZE <= PSA_MLDSA_SIGNATURE_MAX_SIZE,
     "PSA and mldsa-native disagree on the maximum ML-DSA signature size");
 
-static psa_status_t pqcp_to_psa_error(int ret)
+static psa_status_t pqcp_to_psa_error(int ret, psa_status_t default_error)
```
gilles-peskine-arm, line 115: https://github.com/pq-code-package/mldsa-native/pull/1309 will remove the changing meaning of `MLD_ERR_FAIL`. So we can revert this change once that patch reaches our mldsa-native submodule.

drivers/pqcp/src/psa_crypto_mldsa.c
```diff
@@ -221,35 +314,23 @@ psa_status_t tf_psa_crypto_mldsa_sign_message(
...
-    status = tf_psa_crypto_pqcp_alloc_done();
-    mbedtls_platform_zeroize(secret, sizeof(secret));
-    if (status != PSA_SUCCESS) {
-        return status;
+    psa_status_t alloc_status = tf_psa_crypto_pqcp_alloc_done();
+    if (alloc_status != PSA_SUCCESS) {
+        return alloc_status;
     } else {
-        return pqcp_to_psa_error(ret);
+        return status;
```
bjwtaylor, line 334: Does this need to be mapped to the PSA error codes?

gilles-peskine-arm: I'm not sure what you mean. `psa_status_t status` is a PSA error code.

bjwtaylor: So, this PR adds sign_from_expanded() which returns a raw mldsa RC. It is correctly converted to a PSA error at psa_crypto_mldsa.c:282 using pqcp_to_psa_error. However in this case the return of sign_from_expanded is assigned directly to status which is returned here. Does this need a similar conversion, or have I missed something?

gilles-peskine-arm: Got it, fixed.

This alerted me to a test gap: I hadn't added poisoning tests with the private key in joined format (now done). Though it actually doesn't catch this specific mistake, because for an out-of-memory error, we return the allocator error code in preference to the mistakenly unconverted error code.

gilles-peskine-arm:
I'll push another commit to fix the code style. I'm waiting for the CI to complete, to see if there's anything else to fix.

bjwtaylor (dismissed):

drivers/pqcp/src/psa_crypto_mldsa.c
```diff
@@ -221,35 +314,24 @@ psa_status_t tf_psa_crypto_mldsa_sign_message(
...
-    uint8_t secret[TF_PSA_CRYPTO_MLDSA_EXPANDED_SECRET_MAX_SIZE];
-    uint8_t public[TF_PSA_CRYPTO_MLDSA_PUBLIC_KEY_MAX_SIZE];
-
-    int ret = tf_psa_crypto_pqcp_mldsa87_keypair_internal(public,
-                                                          secret,
-                                                          key_buffer);
-    if (ret != 0) {
-        goto cleanup;
+    if (key_buffer_size == SEED_SIZE) {
+        status = sign_from_seed(key_buffer,
```
minosgalanakis, line 318: Checking for key_size after we have called the `tf_psa_crypto_pqcp_alloc_start`. Effectively we permit input to unnecessarily acquire a shared allocator/mutex. Not sure if that can be exploited but I would make it to closely align the multipart setup. which reject invalid key lengths before allocating.
 e.g

~~~~~~~

# Capture invalid private-key length
if (key_buffer_size != SEED_SIZE &&
      key_buffer_size != SEED_SIZE + MLDSA87_SECRETKEYBYTES) {
      return PSA_ERROR_INVALID_ARGUMENT;
  }

  if (signature_size < actual_signature_length) {
      return PSA_ERROR_BUFFER_TOO_SMALL;
  }

# Start the PQCP allocator
  psa_status_t status = tf_psa_crypto_pqcp_alloc_start();

#Then we can then assume the key has a supported length:

  if (key_buffer_size == SEED_SIZE) {
      status = sign_from_seed(...);
  } else {
      status = pqcp_to_psa_error(
          sign_from_expanded(key_buffer + SEED_SIZE, ...),
          PSA_ERROR_HARDWARE_FAILURE);
  }
~~~~~~~


So ideally we should be doing the following:
  1. Validate algorithm, key type and parameter set.
  2. Check signature output size and private-key length.
  3. Acquire the PQCP allocator.
  4. Sign or return PSA_ERROR_INVALID_ARGUMENT.

drivers/pqcp/src/psa_crypto_mldsa.c
```diff
@@ -165,7 +174,47 @@ static psa_status_t seed_to_public_key(
...
+        return PSA_ERROR_BUFFER_TOO_SMALL;
+    }
+
+    psa_status_t status = tf_psa_crypto_pqcp_alloc_start();
+    if (status != PSA_SUCCESS) {
+        return status;
+    }
+
+    int ret = tf_psa_crypto_pqcp_mldsa87_pk_from_sk(data, sk);
+    status = pqcp_to_psa_error(ret, PSA_ERROR_INVALID_ARGUMENT);
```
minosgalanakis, line 208: Could we add a test covering the MLD_ERR_FAIL path from pk_from_sk() here? The non-matching-seed test only changes the leading seed, which this path skips, so the expanded private key remains valid.

We could achieve that using the current tests by corrupting a byte in tr, and checking that the test returns `PSA_ERROR_INVALID_ARGUMENT`

minosgalanakis (changes requested):
Looks good and functionally does what it should. I have only two concerns one about the library changes, and one for adding testing coverage.

drivers/pqcp/src/psa_crypto_mldsa.c
gilles-peskine-arm: I don't understand the concern here. Why does it matter if the code acquires a mutex? Especially one for which we don't expect contention in practice, since the mutex only exists when using a custom allocator, and that only happens in some very constrained environments?

gilles-peskine-arm:
Now that the framework prerequisite is merged, I'm going to rebase this branch. I'll add extra tests in a subsequent commit.

gilles-peskine-arm:
Actually, there's no conflict on the framework, so I'll keep the existing history for now, plus a new framework update commit. It's better to keep the history because there were several changes of the test generator, and if I collapse all the framework updates into one, the generated code and `generate_mldsa_tests.py` won't be consistent with the submodule state anymore.

drivers/pqcp/src/psa_crypto_mldsa.c
minosgalanakis: Mostly I pointed that out as a disparity to the way it is being handled in the other two places in the code. Also this change in order is introduced in this pr, so I am wondering if that was accidental.

My main concern is  weather it would be a good idea to call `tf_psa_crypto_pqcp_alloc_start()` when not validating a valid `key_buffer_size` and weather it could result in a dos, from a miss-configured client. 

I won't block merging this pr for it, but I wanted us to take a closer look to the allocator.

drivers/pqcp/src/psa_crypto_mldsa.c
minosgalanakis: Thank you for adressing this

drivers/pqcp/src/psa_crypto_mldsa.c
gilles-peskine-arm: Ok, it's good in principle to avoid more complex processing on error paths. But it's not like we can avoid having errors pop up with the allocator active. So it would only be a very minor robustness improvement. I don't think it's worth duplicating the `key_buffer_size` checks over this.

gilles-peskine-arm: I'm still fiddling with the follow-up related to expanded keys https://github.com/Mbed-TLS/TF-PSA-Crypto/issues/862 and I might end up refactoring this function (as part of being a bit more systematic about deducing seed/joined/invalid from the private key size and the bits value). Since it's only a minor robustness issue, I'd prefer to do it in the follow-up, if at all.

gilles-peskine-arm: I've done the refactoring I was thinking of when consuming a private key, and that actually fits into this PR thematically. So I might push it here. But I'm not sure I actually like it.

It pushes the (bits, private_key_size) → (validity, format) decision into a single function, so it's now fully uniform. Here, it does have the effect of moving one validation outside the buffer lock/unlock.

In terms of code size, it's a wash.

I'm not sure I like it because I find the new paths more convoluted. On the one hand, we now have a single point of validation, which is a good thing in principle. On the other hand, when you follow the individual paths, I originally had buffer size checks in the same function that relied on them. Now I find it harder to verify that we're passing valid buffer sizes when we call mldsa-native functions. That argument makes me still prefer the original design.

What do you think?

gilles-peskine-arm: Having thought about it overnight, I don't like having the buffer size checks distant from where they matter. Keeping them close means that they will sometimes happen with the buffer mutex held. So I don't plan to make any changes here.

We might want to do some refactoring of buffer size checks when we support multiple parameter sets, but even then, I think the checks may happen with the lock held.

minosgalanakis (approved):
LGTM

Thank you for addressing my comments

bensze01 (approved):
LGTM (incremental review of commits https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/834/changes/4af5f6cc6a9f6d40c5cf922ad10c40440d5be77e..31748d72247c564bfa6166cea173ee7d300163f2)

gilles-peskine-arm:
In the merge queue, one Windows job just disappeared into the ether. This seems like a glitch. I'm requeuing.

Issue #709, opened 2026-03-11
ARM PSA crypto 1.4.0: Implement WPA3-SAE support


krish2718:
### Suggested enhancement

Implement support for WPA3-SAE recently added to ARM PSA Crypto API in 1.4.0: https://arm-software.github.io/psa-api/crypto/1.4/appendix/history.html#:~:text=Added%20support%20for%20the%20WPA3%2DSAE%20PAKE%3A

Also from: https://github.com/Mbed-TLS/mbedtls-docs/blob/main/project/20251125-psa-spec-vs-implementation.pdf

<img width="594" height="103" alt="Image" src="https://github.com/user-attachments/assets/d7534314-49eb-43fe-8bf5-46633ea96e8a" />

### Justification

Mbed TLS needs this because WPA3-SAE is widely used in Wi-Fi and in fact becoming de-facto security standard. So, its essential that MbedTLS implements this PSA crypto API. Any Embedded IoT vendor including the ZephyrRTOS project that uses Wi-Fi will need this, multiple vendors like Nordic, NXP are interested in this feature.

MaochenWang1:
+1 from NXP. 
WPA3-SAE support in PSA Crypto API is important for our wireless connectivity product portfolio. We would appreciate seeing this feature implemented in the TF-PSA-Crypto library.

krish2718:
I've opened #837 with a **Phase 0 design proposal** for WPA3-SAE
(`docs/proposed/wpa3-sae.md`) — a design note only, no implementation
code, so we can agree the approach before any code is written.

**Scope**: hash-to-element only; ECC groups 19/20/21; FFDH via
`PSA_DH_FAMILY_RFC3526`; the `FIXED` and `GDH` variants and the
`WPA3_SAE_H2E` KDF. Protocol-level concerns (anti-clogging, group
negotiation, SAE-PK, mesh) are out of scope — they belong in the Wi-Fi stack.
The note also covers the 1.4 API surface (with spec citations), the new
primitives required (RFC 9380 hash-to-curve, a constant-time modular square
root, the SAE engine), a per-step constant-time strategy, a test plan, and a
proposed P1–P5 PR phasing.

I'd especially value maintainer input on three questions before proceeding:

1. **FFDH group set** — the 1.4 spec only defines `PSA_DH_FAMILY_RFC3526`
   (IANA groups 15–18) for SAE FFDH tokens. Is that the intended set?
[...]

PR #372, opened 2025-07-11
Remove deprecated items

bjwtaylor:
## Description

Remove deprecated items, contributes https://github.com/Mbed-TLS/mbedtls/issues/8138

This PR is part of a chain which needs to be merged in the following order:

1. https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/372
2. https://github.com/Mbed-TLS/mbedtls/pull/10298

## PR checklist

- [ ] **changelog** provided
- [ ] **framework PR** not required
- [ ] **mbedtls development PR** provided Mbed-TLS/mbedtls https://github.com/Mbed-TLS/mbedtls/pull/10298
- [ ] **mbedtls 3.6 PR** not required because: No Backports
- **tests**  provided

PR #837, opened 2026-07-12
docs: add WPA3-SAE (PSA Crypto API 1.4) design proposal

krish2718:
This adds a Phase 0 design proposal (`docs/proposed/wpa3-sae.md`) for
implementing the WPA3-SAE PAKE from the PSA Certified Crypto API 1.4
(§10.13.12–13) in the built-in driver.

It is a **design note for discussion — no implementation code** — intended
to reach agreement on scope, API surface, new primitives, constant-time
strategy, and PR phasing before any code is written. Relates to #709.

### What it covers
- **Scope**: hash-to-element only; ECC groups 19/20/21; FFDH via
  `PSA_DH_FAMILY_RFC3526`; the `FIXED` and `GDH` algorithm variants and the
  `WPA3_SAE_H2E` KDF. Protocol-level concerns (anti-clogging, group
  negotiation, SAE-PK, mesh) are out of scope — they belong in the Wi-Fi stack.
- **API surface** from the 1.4 spec, with section citations.
- **New primitives**: RFC 9380 hash-to-curve (Simplified SWU), a constant-time
  modular square root, and the SAE engine.
- **Constant-time strategy**, per secret-dependent step (Dragonblood is the
  motivating threat).
- **Test plan** and a proposed **P1–P5 PR phasing**.

### Open questions for maintainers (see §9 of the note)
1. **FFDH group set** — the 1.4 spec only defines `PSA_DH_FAMILY_RFC3526`
   (IANA 15–18) for SAE FFDH tokens; confirm this is the intended set.
2. **Looping variant** — proposal is to *not* implement it (no constant-time
   PWE; H2E supersedes it) and return `PSA_ERROR_NOT_SUPPORTED`.
3. **Built-in vs driver-only** — proposal is a full built-in implementation
   (SPAKE2+ is currently API-encoding-only).

### Related
- Spec example errata found during this work are filed as
  GlobalPlatform/psa-api#372.

### DIFF
--- a/docs/proposed/wpa3-sae.md
+++ b/docs/proposed/wpa3-sae.md
@@ -1,0 +1,441 @@
+WPA3-SAE (PSA Certified Crypto API 1.4) — design note
+=====================================================
+
+Status: **draft for maintainer review** (Phase 0). Tracking issue:
+[#709](https://github.com/Mbed-TLS/TF-PSA-Crypto/issues/709).
+
+This note proposes an implementation of the WPA3-SAE password-authenticated key
+exchange (PAKE), as specified by the [PSA Certified Crypto API
+1.4](https://arm-software.github.io/psa-api/crypto/1.4/) (§10.13.12–13,
+Appendix B), for the built-in software driver. It is a design proposal only: no
+implementation code is written yet. The goal is to agree the scope, API
+surface, new primitives, constant-time strategy, and PR phasing **before** any
+code lands.
+
+All PSA symbols below are quoted from the 1.4 specification with a section
+reference. Where the spec is ambiguous, it is flagged as an open question
+rather than guessed.
+
+## 1. Scope
+
+**In scope**
+
+- WPA3-SAE as defined by IEEE 802.11-2024 §12.4, exposed through the PSA 1.4
+  PAKE interface.
+- **Hash-to-element (H2E)** password-element derivation only.
+- ECC groups: **19 / 20 / 21** (`secp256r1` / `secp384r1` / `secp521r1`).
+- FFDH groups: **`PSA_DH_FAMILY_RFC3526`** (IANA groups 15–18) — see §9, open
+  question 1 (this differs from an earlier informal "groups 22/23/24" scope).
+- Both the **fixed-hash** (`PSA_ALG_WPA3_SAE_FIXED`) and
+  **group-dependent-hash** (`PSA_ALG_WPA3_SAE_GDH`) algorithm variants, and the
+  H2E password-token key-derivation (`PSA_ALG_WPA3_SAE_H2E`).
+
+**Explicitly out of scope** (these belong in the Wi-Fi supplicant/stack, not
+the crypto library):
+
+- The **Looping** ("hunting-and-pecking") PWE method. It is the method H2E was
+  designed to replace precisely because it cannot be implemented in constant
+  time (Dragonblood, CVE-2019-9494). See §9, open question 2.
+- Anti-clogging tokens, rejected-groups negotiation beyond the salt input,
+  password identifiers beyond the `INPUT_INFO` KDF input, group negotiation,
+  SAE-PK, and mesh peering.
+- The Hunting-and-Pecking (HNP) construction generally.
+
+## 2. API surface (PSA Crypto API 1.4)
+
+### 2.1 Algorithms — §10.13.13, Appendix B.1.8/B.1.13
+
+| Macro | Encoding | Purpose |
+|---|---|---|
+| `PSA_ALG_WPA3_SAE_FIXED(hash_alg)` | `0x0A0008hh` | Fixed-size-PMK variant (Looping + H2E) |
+| `PSA_ALG_WPA3_SAE_GDH(hash_alg)` | `0x0A0009hh` | Group-dependent-hash variant |
+| `PSA_ALG_WPA3_SAE_ANY` | `0x0A0088FF` | Wildcard key policy |
+| `PSA_ALG_IS_WPA3_SAE(alg)` / `_FIXED` / `_GDH` | — | Predicates |
+| `PSA_ALG_WPA3_SAE_H2E(hash_alg)` | `0x088004hh` | KDF: derive password token (PT) |
+
+`hh` = HASH-TYPE of `hash_alg`. IEEE variant → macro mapping: Looping →
+`FIXED(SHA_256)` (out of scope here); H2E → `FIXED(hash)`; group-dependent-hash
+→ `GDH(hash)`. Hash is fixed by the group prime size (IEEE Table 12-1):
+256→SHA-256, 384→SHA-384, 521→SHA-512.
+
+### 2.2 Primitive / cipher suite — §10.13.12
+
+Standard `PSA_PAKE_PRIMITIVE(type, family, bits)`:
+
+- ECC: `PSA_PAKE_PRIMITIVE_TYPE_ECC`, `PSA_ECC_FAMILY_SECP_R1`, bits ∈ {256,384,521}.
+- FFDH: `PSA_PAKE_PRIMITIVE_TYPE_DH`, `PSA_DH_FAMILY_RFC3526`.
+
+A hash/group mismatch in the cipher suite makes `psa_pake_setup()` return
+`PSA_ERROR_INVALID_ARGUMENT`.
+
+### 2.3 Keys — §10.13.12, key-types, Appendix B.2
+
+- **Password**: `PSA_KEY_TYPE_PASSWORD`, encoded per IEEE §12.4.3.
+- **Password token (PT)** — output of the H2E KDF:
+  - `PSA_KEY_TYPE_WPA3_SAE_ECC(curve)` — `SECP_R1`→`0x3292`,
+    `BRAINPOOL_P_R1`→`0x32b0` (only these two ECC families defined). Key format
+    is `x ‖ y`, each big-endian `m` bytes.
+  - `PSA_KEY_TYPE_WPA3_SAE_DH(group)` — `RFC3526`→`0x3305`.
+- The wildcard policy `PSA_ALG_WPA3_SAE_ANY` lets one password key serve any
+  suite/derivation, and one PT serve both `FIXED` and `GDH`.
+
+### 2.4 PT derivation (H2E KDF) — key-derivation §, IEEE §12.4.4
+
+`PSA_ALG_WPA3_SAE_H2E(hash_alg)` inputs **in order**:
+`PSA_KEY_DERIVATION_INPUT_SALT` = SSID, `PSA_KEY_DERIVATION_INPUT_PASSWORD` =
+password, `PSA_KEY_DERIVATION_INPUT_INFO` = password identifier (optional).
+Produces exactly one key via `psa_key_derivation_output_key()`, of type
+`PSA_KEY_TYPE_WPA3_SAE_ECC/DH`. This is a key→key derivation: the PT never
+materialises as a plaintext buffer.
+
+### 2.5 Operation flow and step types — §10.13.5, §10.13.12
+
+SAE is **balanced**: identical flow on both sides, no `psa_pake_set_role()`;
+`psa_pake_set_user()` / `set_peer()` carry the two **MAC addresses**;
+`psa_pake_set_context()` returns `PSA_ERROR_BAD_STATE`.
+
+| Step | Value | Dir | WPA3-SAE format |
+|---|---|---|---|
+| `PSA_PAKE_STEP_COMMIT` | `0x06` | out & in | `commit-scalar ‖ COMMIT-ELEMENT` (IEEE §12.4.7.3) |
+| `PSA_PAKE_STEP_SALT` | `0x05` | in | rejected-groups list (IEEE §12.4.5.4) |
+| `PSA_PAKE_STEP_CONFIRM_COUNT` | `0x07` | in | 2-byte LE send-confirm counter |
+| `PSA_PAKE_STEP_CONFIRM` | `0x04` | out & in | 2-byte LE counter ‖ confirm (hash output) |
+| `PSA_PAKE_STEP_KEY_ID` | `0x08` | out | 16-byte PMKID |
+
+Sequence: `setup(pt_key,&cs)` → `set_user(mac_a)` → `set_peer(mac_b)` →
+**Commit** `output(COMMIT)` / `input(COMMIT)` + `input(SALT)` → **Confirm**
+`input(CONFIRM_COUNT)` + `output(CONFIRM)`; verify peer via `input(CONFIRM)` →
+optional `output(KEY_ID)` → `get_shared_key()`. Commit-validation failure →
+`PSA_ERROR_INVALID_ARGUMENT`; confirm-verification failure →
+`PSA_ERROR_INVALID_SIGNATURE`.
+
+### 2.6 Shared key (PMK) — §10.13.12
+
+`psa_pake_get_shared_key(&op, &att, &pmk)` — the caller controls the output key
+attributes (type, usage, algorithm, lifetime). PMK is pseudorandom; the spec
+recommends using it as a key-derivation input. This directly supports the
+opacity requirement: a PMK created with `PSA_KEY_USAGE_DERIVE` and **without**
+`PSA_KEY_USAGE_EXPORT` cannot be exported but can still feed a KDF.
+
+### 2.7 Support macros — §10.13.7
+
+Generic and already present: `PSA_PAKE_{OUTPUT,INPUT}_SIZE(alg,primitive,step)`
+and `…_MAX_SIZE`. These currently hard-code J-PAKE/`secp256r1` (=65) in
+`crypto_extra.h` and **must be extended** to compute SAE sizes: COMMIT =
+`scalar ‖ element`, CONFIRM = `2 + hashlen`, KEY_ID = 16.
+
+## 3. Proposed file layout and files to touch
+
+Derived from the existing J-PAKE plumbing (built-in driver) and the SPAKE2+
+API-encoding precedent.
+
+**Public API (`include/psa/`)**
+
+- `crypto_extra.h` — new algorithm macros, `PSA_KEY_TYPE_WPA3_SAE_ECC/DH`, an
+  `sae` member in the operation-struct computation-stage union, and extended
+  `PSA_PAKE_*_SIZE` macros.
+- `crypto_values.h` — `PSA_DH_FAMILY_RFC3526` (FFDH, P3).
+- `crypto_config.h` — `PSA_WANT_ALG_WPA3_SAE_*` gates.
+
+**Config gating**
+
+- `include/tf-psa-crypto/private/crypto_adjust_config_derived.h` — feed
+  `PSA_WANT_ALG_SOME_PAKE`.
+- `drivers/builtin/include/mbedtls/private/crypto_adjust_config_enable_builtins.h`
+  — derive `MBEDTLS_PSA_BUILTIN_ALG_WPA3_SAE_*` / `MBEDTLS_PSA_BUILTIN_PAKE`,
+  pull in ECP/bignum (and, P3, FFDH/RFC3526) modules.
+- `core/check_crypto_config.h` — dependency `#error` checks.
+- `drivers/builtin/include/mbedtls/private/config_adjust_test_accelerators.h`.
+- `scripts/data_files/config-options-current.txt` — list the new
+  `PSA_WANT_ALG_WPA3_SAE_*` options.
+
+**Dispatch** — expected **no change**: `psa_driver_wrapper_pake_*` (generated
+from `scripts/data_files/driver_templates/psa_crypto_driver_wrappers*.jinja`)
+is algorithm-agnostic; per-algorithm branching lives inside the built-in. To be
+confirmed in P4.
+
+**Built-in driver + new primitive**
+
+- `drivers/builtin/src/psa_crypto_pake.c` / `.h` — branch the five entry points
+  (setup / output / input / get_implicit_key / abort) on the SAE algorithm; add
+  H2E KDF handling to the key-derivation path.
+- **new** `drivers/builtin/src/sae.c` + private header — the SAE engine
+  (PWE/commit/confirm/PMK/PMKID), analogue of `ecjpake.c`.
+- **new** `drivers/builtin/src/hash_to_curve.c` + private header — RFC 9380
+  `hash_to_field` + Simplified SWU.
+- `drivers/builtin/src/ecp.c` — refactor the existing modular square root out of
+  `mbedtls_ecp_sw_derive_y()` so it can be shared (see §4.1).
+
+No `CMakeLists.txt` change is needed: `drivers/builtin/CMakeLists.txt` globs
+`src/*.c`.
+
+**Tests**
+
+- **new** `tests/suites/test_suite_sae.{data,function}` — low-level KATs
+  (RFC 9380 + hostap SAE vectors).
+- `tests/suites/test_suite_psa_crypto_pake.{data,function}` — behavioural /
+  negative / opacity / policy cases.
+- `tests/suites/test_suite_psa_crypto_metadata.{data,function}` — algorithm
+  encodings, plus a new `ALG_IS_WPA3_SAE` classification flag in the
+  `.function` file.
+
+**Framework submodule (separate repo)**
+
+- `framework/scripts/mbedtls_framework/crypto_knowledge.py` — register the new
+  algorithm prefixes as `AlgorithmCategory.PAKE`; `macro_collector.py` /
+  `psa_information.py` as needed.
+- `framework/tests/src/psa_exercise_key.c` — a branch in the key-derivation
+  exerciser for the H2E KDF (`SALT` / `PASSWORD` / optional `INFO`).
+
+The framework PR is submitted and merged **first**, and is written to work with
+both the current code and the new code — which is straightforward for a purely
+additive feature. The coordination is only that we want CI on the consuming
+TF-PSA-Crypto PR to pass before the framework PR merges, as evidence that the
+framework side is complete.
+
+**Housekeeping**: `ChangeLog.d/*.txt` fragment; SPDX header
+`Apache-2.0 OR GPL-2.0-or-later` on every new `.c`/`.h`; DCO `Signed-off-by` on
+every commit.
+
+## 4. New primitives and how each will be implemented
+
+In dependency order.
+
+### 4.1 Modular square root + `is_square` (CT)
+
+A `p ≡ 3 (mod 4)` square root already exists inside
+`mbedtls_ecp_sw_derive_y()` (`drivers/builtin/src/ecp.c`), so this is a
+**refactor**, not new code: lift `w^((p+1)/4) mod p` into a reusable helper and
+have `mbedtls_ecp_sw_derive_y()` call it. Three deltas are needed for SAE/SSWU
+use, none of which change existing behaviour:
+
+- The existing function computes its own operand (`ecp_sw_rhs()`); SSWU needs
+  the square root of an arbitrary field element, so the operand becomes a
+  parameter.
+- The existing function documents that it "will return garbage in Y if X does
+  not correspond to a point on the curve" — it does not test squareness. SSWU
+  needs `is_square`, so the helper also returns a CT squareness flag (verify by
+  squaring the result back, rather than a second `exp_mod` for the Legendre
+  symbol).
+- Root selection stays in the callers, because they need different things:
+  `mbedtls_ecp_sw_derive_y()` selects on a parity bit from a public compressed
+  point and its existing `if` is fine there; SSWU selects on secret data and
+  uses `mbedtls_mpi_core_cond_assign`.
+
+All three curves in scope satisfy `p ≡ 3 (mod 4)`, so the existing
+prerequisite check carries over unchanged.
+
+The helper is written at the `bignum_core.h` level (limb array + limb count),
+for the reasons in §4.3. `mbedtls_ecp_sw_derive_y()` can call it directly by
+growing its operand to `grp->P.n` and passing `Y->p, Y->n` — the same thing
+`bignum.c` does internally when it delegates to `mbedtls_mpi_core_*`.
+
+### 4.2 Remaining components
+
+- **Hash-to-curve (RFC 9380, Simplified SWU).** The largest new component.
+  `hash_to_field` (expand_message_xmd over the cipher-suite hash) + SSWU
+  `map_to_curve`; cofactor is 1 for these curves. Implemented branch-free.
+- **SAE engine (`sae.c`).** PWE-from-PT (IEEE §12.4.4.2.3), commit
+  (scalar/element generation with blinded scalar mult via `mbedtls_ecp_mul`),
+  confirm (SAE KDF + HMAC over the transcript), PMK, PMKID, and peer-input
+  validation (`mbedtls_ecp_check_pubkey`, range/identity/reflection checks).
+- **SAE KDF.** IEEE `KDF-Hash-Length`, a thin HMAC-based expansion assembled
+  from the existing HMAC.
+- **FFDH** (separate PR, see §7). New `PSA_DH_FAMILY_RFC3526` + RFC 3526 MODP
+  group constants + the FFDH H2E direct method (IEEE §12.4.4.3.3, iterated
+  exponentiation + QR test). None of this exists today (only
+  `PSA_DH_FAMILY_RFC7919`).
+
+### 4.3 Bignum abstraction level
+
+`bignum_mod.h` is the preferred level for new code, and on paper it fits: SSWU
+works modulo one fixed prime for a long run of operations, which is exactly the
+case the modulus/residue structures are built for. In practice it does not
+currently reach, for four reasons:
+
+- **No exponentiation.** The module exposes `mul`, `sub`, `add`, `inv`,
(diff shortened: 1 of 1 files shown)


docs/proposed/wpa3-sae.md
```diff
@@ -0,0 +1,312 @@
...
+  `hash_to_field` + Simplified SWU, plus the modular-sqrt / `is_square` helper.
+- `CMakeLists.txt` / builtin cmake — register new sources.
+
+**Tests**
+
+- **new** `tests/suites/test_suite_sae.{data,function}` — low-level KATs
+  (RFC 9380 + hostap SAE vectors).
+- `tests/suites/test_suite_psa_crypto_pake.{data,function}` — behavioural /
+  negative / opacity cases.
+- `tests/suites/test_suite_psa_crypto_metadata.data` — algorithm encodings.
```
gilles-peskine-arm: Also a new flag `ALG_IS_WPA3_SAE` in `tests/suites/test_suite_psa_crypto_metadata.function`

docs/proposed/wpa3-sae.md
```diff
@@ -0,0 +1,312 @@
...
+**Built-in driver + new primitive**
+
+- `drivers/builtin/src/psa_crypto_pake.c` / `.h` — branch the five entry points
+  (setup / output / input / get_implicit_key / abort) on the SAE algorithm; add
+  H2E KDF handling to the key-derivation path.
+- **new** `drivers/builtin/src/sae.c` + private header — the SAE engine
+  (PWE/commit/confirm/PMK/PMKID), analogue of `ecjpake.c`.
+- **new** `drivers/builtin/src/hash_to_curve.c` + private header — RFC 9380
+  `hash_to_field` + Simplified SWU, plus the modular-sqrt / `is_square` helper.
+- `CMakeLists.txt` / builtin cmake — register new sources.
```
gilles-peskine-arm: Adding new files in an existing directory should not require any changes to `**/CMakeLists.txt`.

docs/proposed/wpa3-sae.md
```diff
@@ -0,0 +1,312 @@
...
+- `crypto_config.h` — `PSA_WANT_ALG_WPA3_SAE_*` gates.
+
+**Config gating**
+
+- `include/tf-psa-crypto/private/crypto_adjust_config_derived.h` — feed
+  `PSA_WANT_ALG_SOME_PAKE`.
+- `drivers/builtin/include/mbedtls/private/crypto_adjust_config_enable_builtins.h`
+  — derive `MBEDTLS_PSA_BUILTIN_ALG_WPA3_SAE_*` / `MBEDTLS_PSA_BUILTIN_PAKE`,
+  pull in ECP/bignum (and, P4, FFDH/RFC3526) modules.
+- `core/check_crypto_config.h` — dependency `#error` checks.
```
gilles-peskine-arm, line 147: Also `scripts/data_files/config-options-current.txt`

docs/proposed/wpa3-sae.md
```diff
@@ -0,0 +1,312 @@
...
+
+- **new** `tests/suites/test_suite_sae.{data,function}` — low-level KATs
+  (RFC 9380 + hostap SAE vectors).
+- `tests/suites/test_suite_psa_crypto_pake.{data,function}` — behavioural /
+  negative / opacity cases.
+- `tests/suites/test_suite_psa_crypto_metadata.data` — algorithm encodings.
+
+**Framework submodule (separate repo, coordinated PR)**
+
+- `framework/scripts/mbedtls_framework/crypto_knowledge.py` — register the new
```
gilles-peskine-arm, line 184: Also probably a block in `framework/tests/src/psa_exercise_key.c` for the new KDF.

docs/proposed/wpa3-sae.md
```diff
@@ -0,0 +1,312 @@
...
+
+- **Low-level KATs** in `test_suite_sae.*`: RFC 9380 App J vectors (checking
+  intermediate `u`, `Q0`, `Q1`, `P`, not just the final point) and hostap
+  (BSD-3-Clause) SAE vectors for PT/PWE/commit/confirm/PMK/PMKID.
+- **Behavioural** in `test_suite_psa_crypto_pake.*`: full `psa_pake_*` flow per
+  group; algorithm-encoding lines in `test_suite_psa_crypto_metadata.data`.
+- **Negative set**: point-not-on-curve, identity element, out-of-range scalar,
+  reflection, malformed COMMIT (length), wrong confirm/counter/replay, wrong
+  password, `set_context` rejected, incompatible hash/group.
+- **Opacity (acceptance criteria)**: (a) PMK from `get_shared_key()` with
```
gilles-peskine-arm, line 339: Also test algorithm policies: non-wildcard policies require an exact match, and test the extent of `PSA_ALG_WPA3_SAE_ANY`.

docs/proposed/wpa3-sae.md
```diff
@@ -0,0 +1,312 @@
...
+  fuzzing of peer inputs, the full negative-test set, and documentation.
+
+## 8. Process conformance (CONTRIBUTING.md)
+
+- **License**: dual `Apache-2.0 OR GPL-2.0-or-later`; SPDX identifier on every
+  new source file; "Copyright The Mbed TLS Contributors" comment.
+- **DCO**: `Signed-off-by:` on every commit (this repo's configured git
+  identity).
+- **ChangeLog**: a `ChangeLog.d/*.txt` fragment (Features section) per the
+  `ChangeLog.d/00README.md` format.
```
gilles-peskine-arm: Note: changelog entries only for things visible in the API, not for internal elements that are not reachable from the API yet.

docs/proposed/wpa3-sae.md
```diff
@@ -0,0 +1,312 @@
...
+- **Tests**: comprehensive tests required with the feature; coverage comparable
+  to existing code.
+- **Coding style**: Mbed TLS coding standards; `.uncrustify.cfg` in-tree.
```
gilles-peskine-arm: Specifically uncrustify 0.75.1. But it's almost K&amp;R2 so the details shouldn't be a problem.

docs/proposed/wpa3-sae.md
```diff
@@ -0,0 +1,312 @@
...
+
+**Tests**
+
+- **new** `tests/suites/test_suite_sae.{data,function}` — low-level KATs
+  (RFC 9380 + hostap SAE vectors).
+- `tests/suites/test_suite_psa_crypto_pake.{data,function}` — behavioural /
+  negative / opacity cases.
+- `tests/suites/test_suite_psa_crypto_metadata.data` — algorithm encodings.
+
+**Framework submodule (separate repo, coordinated PR)**
```
gilles-peskine-arm: The framework PR will be merged first, and needs to work both with the current code and with the new code (but this shouldn't be an issue when adding a new feature). It only needs to be coordinated in the sense that we like the CI on the using PR to pass before we merge the framework PR, because that gives assurance that the framework PR isn't missing something.

docs/proposed/wpa3-sae.md
```diff
@@ -0,0 +1,312 @@
...
+
+**Housekeeping**: `ChangeLog.d/*.txt` fragment; SPDX header
+`Apache-2.0 OR GPL-2.0-or-later` on every new `.c`/`.h`; DCO `Signed-off-by` on
+every commit.
+
+## 4. New primitives and how each will be implemented
+
+Dependency order:
+
+1. **Modular square root + `is_square` (CT).** None exists in the library.
```
gilles-peskine-arm: There is already a square root implementation in `mbedtls_ecp_sw_derive_y()` in `ecp.c`. Prefer refactoring this to make it usable for SAE.

docs/proposed/wpa3-sae.md
```diff
@@ -0,0 +1,312 @@
...
+   (scalar/element generation with blinded scalar mult via `mbedtls_ecp_mul`),
+   confirm (SAE KDF + HMAC over the transcript), PMK, PMKID, and peer-input
+   validation (`mbedtls_ecp_check_pubkey`, range/identity/reflection checks).
+4. **SAE KDF.** IEEE `KDF-Hash-Length`, a thin HMAC-based expansion assembled
+   from the existing HMAC.
+5. **(P4) FFDH.** New `PSA_DH_FAMILY_RFC3526` + RFC 3526 MODP group constants +
+   the FFDH H2E direct method (IEEE §12.4.4.3.3, iterated exponentiation + QR
+   test). None of this exists today (only `PSA_DH_FAMILY_RFC7919`).
+
+Existing building blocks we rely on: the constant-time module
```
gilles-peskine-arm, line 297: The library has several internal bignum interfaces, from lowest level to highest level: `bignum_core.h`, `bignum_mod_raw.h`, `bignum_mod.h`, `mbedtls/private/bignum.h`. Lower-level interfaces generally have less overhead and are easier to make constant-time, but have a higher risk of memory management mistakes.

The existing code in `ecp.c` mostly relies on `mbedtls/private/bignum.h` because that was the only one available when `ecp.c` was originally written. However, for new code, `bignum_mod.h` is generally preferred, with `bignum_mod_raw.h` or `bignum_core.h` available if `bignum_mod.h` is not flexible enough or induces too much overhead.

Given that the new code for this feature will have a large adherence to the existing code in `ecp.c`, it's not obvious to me which abstraction level to prefer.

docs/proposed/wpa3-sae.md
```diff
@@ -0,0 +1,312 @@
...
+- **ChangeLog**: a `ChangeLog.d/*.txt` fragment (Features section) per the
+  `ChangeLog.d/00README.md` format.
+- **Tests**: comprehensive tests required with the feature; coverage comparable
+  to existing code.
+- **Coding style**: Mbed TLS coding standards; `.uncrustify.cfg` in-tree.
+- **API stability**: this is additive (new macros/types/functions); it can only
+  merge at a major/minor release per the API-extension policy, and the framework
+  change requires a coordinated submodule PR.
+
+## 9. Open questions for maintainers
```
gilles-peskine-arm, line 405: The design choices made here match Arm's preferences.

docs/proposed/wpa3-sae.md
```diff
@@ -0,0 +1,312 @@
...
+  `tests/scripts/components-sanitizers.sh`. New CT cases must drive the
+  SAE/SSWU/mod-sqrt paths with poisoned secret inputs. Documented limit: this
+  catches secret-dependent branches/indexing, not micro-architectural timing.
+
+## 7. Proposed PR phasing
+
+Each phase is independently reviewable; earlier phases land without depending
+on later ones.
+
+- **P1 — test vectors + harness (standalone).** `test_suite_sae.*` skeleton
```
gilles-peskine-arm: Please avoid landing test vectors before the corresponding algorithm. In particular, merging tests that are not executed is technically possible, but it's a hassle.

docs/proposed/wpa3-sae.md
```diff
@@ -0,0 +1,312 @@
...
+  with RFC 9380 and hostap vectors captured as data; the framework
+  `crypto_knowledge.py` registration. Lands first, no algorithm code.
+- **P2 — core primitives.** mod-sqrt/`is_square`, RFC 9380 hash-to-curve,
+  PWE/PT, commit, confirm, SAE KDF — as low-level `sae.c` /
```
gilles-peskine-arm: As far as I can tell from reading this document, but without being familiar with the mechanisms that are implemented, it seems to me that `sae.c` (+ unit tests) and `hash_to_curve.c` (+ unit tests) should be separate PR. I'm not sure about the other bits.

docs/proposed/wpa3-sae.md
```diff
@@ -0,0 +1,312 @@
...
+  `hash_to_curve.c`, validated against P1 KATs. No PSA wiring yet.
+- **P3 — PSA integration: group 19 / H2E / fixed hash, end to end.** Algorithm
+  macros, key types, config gates, built-in driver entry points, `get_shared_key`
+  path; full `psa_pake_*` flow for `secp256r1`.
+- **P4 — remaining groups + variants.** ECC groups 20/21, GDH variant, then
+  FFDH (`PSA_DH_FAMILY_RFC3526` + MODP data + FFDH H2E).
+- **P5 — hardening.** Constant-time verification (MSan/Valgrind CT cases),
```
gilles-peskine-arm: Constant-time annotations don't involve much code. They should preferably be present from the start.

Fuzzing, on the other hand, can be added as a separate later PR.

docs/proposed/wpa3-sae.md
```diff
@@ -0,0 +1,312 @@
...
+  fuzzing of peer inputs, the full negative-test set, and documentation.
```
gilles-peskine-arm: Documentation should be provided as early as possible, not as an afterthought. Writing good documentation often helps remove ambiguities from the specification and helps with the test plan.

docs/proposed/wpa3-sae.md
```diff
@@ -0,0 +1,312 @@
...
+- **P1 — test vectors + harness (standalone).** `test_suite_sae.*` skeleton
+  with RFC 9380 and hostap vectors captured as data; the framework
+  `crypto_knowledge.py` registration. Lands first, no algorithm code.
+- **P2 — core primitives.** mod-sqrt/`is_square`, RFC 9380 hash-to-curve,
+  PWE/PT, commit, confirm, SAE KDF — as low-level `sae.c` /
+  `hash_to_curve.c`, validated against P1 KATs. No PSA wiring yet.
+- **P3 — PSA integration: group 19 / H2E / fixed hash, end to end.** Algorithm
+  macros, key types, config gates, built-in driver entry points, `get_shared_key`
+  path; full `psa_pake_*` flow for `secp256r1`.
+- **P4 — remaining groups + variants.** ECC groups 20/21, GDH variant, then
```
gilles-peskine-arm: `PSA_DH_FAMILY_RFC3526` should be added in a PR that doesn't add H2E/SAE code.

gilles-peskine-arm (changes requested):
Thank you for providing a detailed design plan.

I've reviewed it as someone who is familiar with PSA APIs in general and with TF-PSA-Crypto as a product, but not with the PAKE API and not with the WPA3 algorithm family. In particular, I have not confronted the proposal with the WPA3 and related specifications.

docs/proposed/wpa3-sae.md
krish2718: Added.

docs/proposed/wpa3-sae.md
krish2718: Right, it globs `src/*.c` — dropped that line.

docs/proposed/wpa3-sae.md
krish2718: Added.

docs/proposed/wpa3-sae.md
krish2718: Added.

docs/proposed/wpa3-sae.md
krish2718: Added — exact-match on non-wildcard policies (`FIXED(SHA_256)` key rejected for `GDH(SHA_256)` and for `FIXED(SHA_384)`), plus what `PSA_ALG_WPA3_SAE_ANY` does and does not cover.

docs/proposed/wpa3-sae.md
krish2718: Noted. P1/P2 are internal-only, so no entry there; the ChangeLog fragment goes with the PSA-integration PR.

docs/proposed/wpa3-sae.md
krish2718: Pinned 0.75.1 in the note.

docs/proposed/wpa3-sae.md
krish2718: Thanks, that's clearer than what I had — reworded: framework merges first, written to work with both old and new code, and CI green on the consuming PR as the signal it isn't missing anything.

docs/proposed/wpa3-sae.md
krish2718: Missed that one — will refactor it rather than add a second sqrt. Three deltas for the SSWU side, none changing existing behaviour: the operand becomes a parameter (SSWU needs the sqrt of an arbitrary field element, not `ecp_sw_rhs()`); it returns a CT `is_square` flag (cheapest by squaring the result back, not a second `exp_mod`); and root selection stays in the callers, since `sw_derive_y` selects on a public parity bit and SSWU selects on secret data. Written up as §4.1, planned at `bignum_core.h` level — ecp.c can call it by growing to `grp->P.n` and passing `Y->p, Y->n`.

docs/proposed/wpa3-sae.md
krish2718: I went through the headers on this and I think the choice is narrower than it first looks. `bignum_mod.h` fits SSWU conceptually — one fixed modulus, a long run of ops — but as it stands it can't express it:

- no exponentiation, and the square root is `w^((p+1)/4) mod p`. `exp_mod` exists only as `mbedtls_mpi_exp_mod()` / `mbedtls_mpi_core_exp_mod()`;
- no conditional assign — CMOV is `mbedtls_mpi_mod_raw_cond_assign()`, a layer down, and SSWU needs it throughout;
- no comparison — `lt_ct` / `check_zero_ct` are in `bignum_core.h`;
- residues may be in Montgomery form, so RFC 9380 `sgn0` (a parity test on the canonical value) reads the wrong thing without converting out.

So it's really `bignum_core.h` vs extending `bignum_mod.h` first. I'd lean `bignum_core.h` for `hash_to_curve.c` and the shared sqrt (it has exp_mod, cond_assign, lt_ct, check_zero_ct and explicit mont conversion, at fixed limb counts), and keep `sae.c` on `mbedtls_mpi`, since it's mostly orchestrating `mbedtls_ecp_*` and anything else just buys conversions at every boundary.

Fixed-limb matters here beyond style: `mbedtls_mpi` normalizes its limb count, so a secret value that's numerically small runs shorter and leaks its magnitude — same class as Dragonblood — so I'd rather not have it near the field arithmetic. (`mbedtls_mpi_exp_mod()` is fine on that point; it grows to `N->n` first.)

Happy to extend `bignum_mod.h` instead if you'd prefer the library to move that way, but I'd rather that be its own piece of work than a dependency of this one. Written up as §4.3 / open question 5 — leaving this thread open for your call.

docs/proposed/wpa3-sae.md
krish2718: Dropped the vectors-only phase — tests now land in the PR with the code they exercise.

docs/proposed/wpa3-sae.md
krish2718: Agreed, split. P1 = sqrt refactor + `hash_to_curve.c` (RFC 9380 App J vectors), P2 = `sae.c` (hostap vectors). P2 depends on P1; nothing else does.

docs/proposed/wpa3-sae.md
krish2718: Done — CT annotations are a per-phase rule now, not a phase; fuzzing is its own later PR.

docs/proposed/wpa3-sae.md
krish2718: Agreed — docs land with the code that introduces them.

docs/proposed/wpa3-sae.md
krish2718: Done — it's a standalone phase now, independent of the SAE work, can land in any order relative to it.

krish2718:
Thanks, this was a useful pass — all 16 points are in the updated note (force-pushed). Substantive changes:

- **§7 phasing** reworked: no vectors-only PR (tests land with the code they exercise), `hash_to_curve.c` and `sae.c` split into separate PRs, `PSA_DH_FAMILY_RFC3526` as a standalone phase with no SAE code, constant-time annotations and documentation as per-phase rules rather than a trailing hardening pass, fuzzing last.
- **§4.1**: the modular square root is now a refactor of `mbedtls_ecp_sw_derive_y()` rather than new code.
- **new §4.3 + open question 5**: bignum abstraction level.

On that last one — the answer came out more definite than "open question" suggests, so flagging it here too: `bignum_mod.h` has no exponentiation, no conditional assign and no comparison, and its Montgomery representation breaks RFC 9380 `sgn0`, so the real choice is `bignum_core.h` vs extending `bignum_mod.h` first. We'd propose the former; detail in the thread on §4. That's the one thread I've left open for your call — the rest of the review is addressed and resolved.

The three maintainer-facing open questions (§9: FFDH group set, omitting Looping, built-in vs encoding-only) are unchanged and still open.

Unrelated: the spec example errata (#372) merged, targeted for 1.5.1 — the note now reflects that.

docs/proposed/wpa3-sae.md
```diff
@@ -0,0 +1,442 @@
+WPA3-SAE (PSA Certified Crypto API 1.4) — design note
```
gilles-peskine-arm, line 1: It's ok for this time because all the changes are in a single new file, but for future reference: please avoid force-pushing to a pull request after the first review has started. Make new changes in separate commits. Each commit should only do one thing (no big commit “address all the review comments”).

docs/proposed/wpa3-sae.md
```diff
@@ -0,0 +1,442 @@
...
+- **No exponentiation.** The module exposes `mul`, `sub`, `add`, `inv`,
+  `random`, `read`, `write` and modulus/residue lifecycle — and nothing else.
+  The square root is `w^((p+1)/4) mod p`, so it cannot be expressed there.
+  Modular exponentiation exists only as `mbedtls_mpi_exp_mod()` and
+  `mbedtls_mpi_core_exp_mod()` (`bignum_core.h`, documented constant-time with
+  respect to `A`, `N` and `E`).
+- **No conditional assign.** CMOV is `mbedtls_mpi_mod_raw_cond_assign()`, one
+  layer down; SSWU needs it throughout.
+- **No comparison.** The CT predicates (`mbedtls_mpi_core_lt_ct`,
+  `mbedtls_mpi_core_check_zero_ct`) live in `bignum_core.h`.
```
gilles-peskine-arm, line 265: It would be easy to add comparison and conditional assignment, and not hard to add exponentiation. So I don't think those missing functions should weigh much in the decision.

docs/proposed/wpa3-sae.md
```diff
@@ -0,0 +1,442 @@
...
+- **Montgomery representation gets in the way of `sgn0`.** Residues may be held
+  in Montgomery form depending on the modulus `int_rep`, so a bit-level parity
+  test — which RFC 9380 `sgn0` needs — reads the wrong value unless converted
+  out first.
```
gilles-peskine-arm, line 269: A conversion step can be done. And conversion has to be done manually anyway when using bignum_core. How is bignum_mod worse in this respect?

docs/proposed/wpa3-sae.md
```diff
@@ -0,0 +1,442 @@
...
+
+Proposal, therefore:
+
+- **`hash_to_curve.c` and the square-root helper → `bignum_core.h`.** Fixed
+  limb counts, no allocation, and it has the whole toolkit this code needs:
+  `exp_mod`, `cond_assign`, `lt_ct`, `check_zero_ct`, and explicit
+  `to_mont_rep`/`from_mont_rep` when we want Montgomery form. The cost is the
+  one Gilles noted — manual management of the Montgomery form, the precomputed
```
gilles-peskine-arm: “the one Gilles noted” won't make sense outside of the context of the review of this pull request.

docs/proposed/wpa3-sae.md
```diff
@@ -0,0 +1,442 @@
...
+  `ChangeLog.d/00README.md` format, only for what is visible through the API.
+  The internal-primitive phases (P1, P2) are not reachable from the API and
+  therefore get no ChangeLog entry; the entry goes with P4.
+- **Tests**: comprehensive tests required with the feature; coverage comparable
+  to existing code.
+- **Coding style**: Mbed TLS coding standards, enforced with **uncrustify
+  0.75.1** and the in-tree `.uncrustify.cfg`.
+- **API stability**: this is additive (new macros/types/functions); it can only
+  merge at a major/minor release per the API-extension policy. The framework
+  submodule change is a separate PR — see §3.
```
gilles-peskine-arm, line 403: To be precise, the changes inside the framework are a separate PR since it's a separate repository. Updating the framework submodule in the TF-PSA-Crypto repository should be done in the TF-PSA-Crypto PR that needs it, but the need may go away if another TF-PSA-Crypto PR updates the framework for its own needs first.

docs/proposed/wpa3-sae.md
```diff
@@ -0,0 +1,442 @@
...
+   - §10.13.10 (SPAKE2+) example used the undefined `PSA_PAKE_STEP_KEY_CONFIRM`;
+     the defined step is `PSA_PAKE_STEP_CONFIRM` (0x04).
+5. **Bignum abstraction level.** See §4.3. `bignum_mod.h` is the preferred
+   level for new code, but as it stands it has no exponentiation, no
+   conditional assign and no comparison, and its Montgomery representation
+   breaks RFC 9380 `sgn0`. So the choice is really: use `bignum_core.h`, or
+   extend `bignum_mod.h` first. We propose `bignum_core.h` for
+   `hash_to_curve.c` and the shared square root, and `mbedtls_mpi` for
+   `sae.c`. If extending `bignum_mod.h` is wanted instead, we would rather do
+   it as its own piece of work than on this feature's critical path.
```
gilles-peskine-arm, line 438: Many functions in the bignum modules and in the ecp module have automatically generated tests (`framework/scripts/generate_bignum_tests.py`, `framework/scripts/generate_ecp_tests.py`). If this work introduces new functions that should have automatically generated tests, we'll need to arrange. to add new tested functions

Because the framework code and the crypto code can't be updated atomically, the first time those scripts need to cover a new test function, this will require preliminary work to split the script into two parts:

* `framework/scripts/mbedtls_framwork/xxx_generator.py` having almost all the code.
* A small harness `tests/scripts/generate_xxx.py` that calls an entry point in the framework and passes it arguments that describe which test functions should be covered.
* For a transition period, `framework/scripts/generate_xxx.py` still exists. It calls `tests/scripts/generate_xxx.py` from the consuming branch if it exists, and otherwise generates test cases for the historical set of test functions.

See `scripts/maintainer/generate_mldsa_tests.py` and `framework/util/mbedtls_maintainer/mldsa_test_generator.py` for an example of this split. Note that the file paths are different for mldsa because it needs extra python packages, and that would break some validation scripts that we run on `framework/scripts`, `scripts` and `tests/scripts`; that doesn't apply for bignum and ecp.

gilles-peskine-arm (changes requested):
Thanks for the update. I'm generally happy with the plan now, except that I'm unsure about `bignum_core` vs `bignum_mod`: the arguments given against `bignum_mod` feel very weak to me. It's hard for me to judge because I'm not at all familiar with the algorithm.

Do you have a reading suggestion (code or math+English) that would make it clear how the representation of numbers matter? Are all the numbers involved in the same $Z_p$, or is there a mix? At which steps are conversions to/from Montgomery form needed?

Note: when I say I'm happy with the plan, that's as a general plan for implementing some calculations as a new cryptographic mechanism in TF-PSA-Crypto. I'm not familiar with the WPA3-SAE family of algorithms, so I can't vouch for difficulties or incompleteness for implementing the desired algorithms.

docs/proposed/wpa3-sae.md
krish2718: Assitant (LLM) leakage :), will fix it thanks.

docs/proposed/wpa3-sae.md
krish2718: You're right, and two of my three arguments don't hold up:

- "missing functions" — fair, `add`/`sub`/`mul`/`inv` are there and cmp / cond-assign / exp are easy to add, so that shouldn't weigh much.
- the limb-normalization / magnitude-leak point — that's an argument against legacy `mbedtls_mpi`, not against `bignum_mod`. `bignum_mod` residues are fixed-size (limbs == modulus limbs), so it doesn't distinguish the two. I conflated two axes; scratch that one.

So it's really about ergonomics and blast radius, not capability.

Your questions:

**All in the same Z_p, or a mix?** For `hash_to_curve.c` it's essentially one field. Every step of `hash_to_field` + simple-SWU + the square root is arithmetic mod `p`, the curve's base-field prime — mul/add/sub/inv/sqrt/sgn0, all mod `p`; cofactor clearing is a no-op (h=1 for P-256/384/521). The only non-field integers are the exponents fed to `exp_mod` (e.g. `(p+1)/4`), which are bit-strings not residues, and the final point add `PT = P1 + P2`, a curve-group op best handed to `mbedtls_ecp`. Scalars mod the group order `n` show up in the SAE protocol (commit scalar etc.) but those live in `sae.c`, not here. So: single modulus, long run of ops — which is exactly the `bignum_mod` sweet spot, so your instinct is the right one.

**Where are Montgomery conversions needed?** With `bignum_mod` in Montgomery `int_rep`: convert in once when the field element is read (`mbedtls_mpi_mod_read`), then every mul/add/sub/inv stays in Montgomery — no per-op conversion — and convert out only at (a) `sgn0`, since RFC 9380's sign is the parity of the *canonical* representative, and (b) the final handoff to point arithmetic. `exp_mod` for the sqrt wants the base in Montgomery plus a precomputed `RR`, which the modulus object already carries. With `bignum_core` you do the identical conversions by hand and manage `RR` yourself — so on conversions the two are a wash, which was your point on the other thread.

So I'll drop the objection: `bignum_mod` is a good fit for `hash_to_curve.c` and the shared sqrt. The one tradeoff left is scope, not correctness — see my reply on the generated-tests comment.

Reading, if useful: RFC 9380 §4.1 (`hash_to_field`, `sgn0`) and §6.6.2 (simple SWU) for what the numbers are; the `bignum_mod.h` header doc block (`int_rep` / residue model) plus HAC §14.3.2 (Montgomery reduction) for why representation matters — short version: Montgomery form stores `x·R mod p` so reduction is shifts instead of division; it's stable under +/-/×, but any step that reads the integer itself (parity, comparison, serialization) has to convert back first, which is why `sgn0` is the one place we drop out.

docs/proposed/wpa3-sae.md
krish2718: Agreed — conceded on the main thread; that argument doesn't carry weight, the functions are cheap to add.

docs/proposed/wpa3-sae.md
krish2718: Right — conversions are manual either way; I walked through where they land on the main thread. Not a point in `bignum_core`'s favour.

docs/proposed/wpa3-sae.md
krish2718: This is the real deciding factor now that capability isn't. If we go `bignum_mod` we're adding `exp_mod` (and probably cmp / cond-assign) as first-class `bignum_mod` functions, which per your note means doing the generator split — a `mbedtls_framework/..._generator.py` plus a thin `tests/scripts/generate_*.py` harness, transition shim and all — following the mldsa example. `bignum_core` already has all of these, so that path adds no bignum API and touches no generator.

So the choice is: grow `bignum_mod` properly (better long-term home, but the first code PR then also owns a bignum-module extension + the generator split), or keep the first SSWU PR tight on `bignum_core` and leave `bignum_mod` growth as separate future work. I lean toward doing it properly on `bignum_mod` since you'd want those functions eventually anyway — but you're better placed to say whether you want that coupled into the first SSWU PR or kept out of it. Either way I'll write it up to match.

docs/proposed/wpa3-sae.md
krish2718: Understood — I'll fix §3 to say it precisely: the framework changes are a separate PR (separate repo), and the submodule bump lands in whichever TF-PSA-Crypto PR first needs it, so the "merged first" framing was too rigid.

docs/proposed/wpa3-sae.md
krish2718: Removed in 986f205.

krish2718:
@gilles-peskine-arm — I answered your three questions (representation/reading, single-field vs mix, where the Montgomery conversions land) on the bignum thread rather than here so they wouldn't get buried: https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/837#discussion_r3641366480

Short version: it's a single field (F_p) throughout `hash_to_curve.c`, conversions are in-once and out-only-at-`sgn0`, and I've dropped the objection — `bignum_mod` is a fine fit. The one open point is scope: grow `bignum_mod` (which pulls in the generator split) vs keep the first PR tight on `bignum_core`. Your call.

gilles-peskine-arm (approved):
LGTM at 986f205106f7a8602dae59afcb6ef91c9a5b315a, with the caveat that I am not familiar with WPA3-SAE and I have not checked the plan against the specification of the mechanisms.

On `bignum_core` vs `bignum_mod`, I could go either way based on my knowledge. I'll ask a colleague who's more familiar with ECC math and the planned evolution of bignum and ECP modules.

krish2718:
@gilles-peskine-arm Should I add any other reviewer to get this merged?

gilles-peskine-arm:
We do need a second reviewer. But I have no idea when someone will be available or who it will be.

By the way, given the size of the contribution, have you made arrangements with our product management to have the review of the code scheduled? We currently have very little bandwidth since most of the team is triaging and fixing LLM-assisted vulnerability reports.

krish2718:
> We do need a second reviewer. But I have no idea when someone will be available or who it will be.

Ah, any process to followup here or a list of people to nudge :)?

> By the way, given the size of the contribution, have you made arrangements with our product management to have the review of the code scheduled? We currently have very little bandwidth since most of the team is triaging and fixing LLM-assisted vulnerability reports.

No, but if you can guide me, I am happy to connect and take this forward with the product management.

PR #843, opened 2026-07-21
Fix getrandom detection with musl libc

okning:
## Description

On Linux, the platform entropy implementation currently enables the
getrandom syscall path only for glibc and Midipix. This excludes musl and
other libc implementations that can use the same syscall path, so affected
builds fall back to `/dev/random`. On older Linux kernels, a subsequent
process may block during PSA crypto initialization after earlier reads
have depleted the available entropy count.

Use `__has_include(<sys/syscall.h>)` on Linux when supported by the
compiler, and enable the existing syscall wrapper only when
`SYS_getrandom` is defined. Preserve the glibc/Midipix condition for older
toolchains and the existing fallback when the kernel returns `ENOSYS`.

The initial `<sys/random.h>`/libc-wrapper detection was replaced during
review because it broke dietlibc builds. The final implementation reuses
`syscall(SYS_getrandom, ...)` and adds no public configuration option.

## Validation

Manual validation of this development PR, reported on August 3:

- glibc 2.43 / GCC 16.1: build passed; syscall path selected; two consecutive initializations passed.
- musl / GCC 12.4: full build passed; syscall path selected; two consecutive initializations passed.
- dietlibc 0.34: build passed; `SYS_getrandom` is absent and the existing fallback remained selected.
- uClibc-ng 1.0.54 / GCC 14.3: static build passed; syscall path selected; two consecutive initializations passed.
- Android bionic / NDK r29 / API 21 / x86_64: dynamic and static builds passed; syscall path selected; the static binary passed two consecutive initializations.
- AppleClang 21 Debug build passed, followed by 132/132 CTest tests.
- ChangeLog assembly check passed.

The individual backport PRs document their separate validation results.

## PR checklist

- [x] **changelog** provided.
- [x] **framework PR** not required.
- [x] **TF-PSA-Crypto development PR** this PR.
- [x] **TF-PSA-Crypto 1.1 PR** Mbed-TLS/TF-PSA-Crypto#893.
- [x] **mbedtls development PR** not required because the implementation is maintained in TF-PSA-Crypto.
- [x] **mbedtls 4.1 PR** not provided; the corresponding submodule update can follow the TF-PSA-Crypto backport.
- [x] **mbedtls 3.6 PR** Mbed-TLS/mbedtls#10945.
- **tests** no new committed test cases; manual libc testing was accepted during review and is recorded above.

### DIFF
--- a/ChangeLog.d/musl_getrandom.txt
+++ b/ChangeLog.d/musl_getrandom.txt
@@ -1,0 +1,3 @@
+Bugfix
+   * Fix PSA crypto initialization needlessly blocking on `/dev/random` on
+     Linux 5.6 and earlier when using a libc such as musl.
--- a/ChangeLog.d/musl_getrandom.txt
+++ b/ChangeLog.d/musl_getrandom.txt
@@ -1,5 +1,5 @@
 Features
-   * Use the `getrandom` system call for PSA crypto initialization on Linux
+   * Use the getrandom() system call for PSA crypto initialization on Linux
      when supported by the C library, including musl, uClibc and Android
-     bionic, avoiding unnecessary reliance on `/dev/random` on Linux 5.6 and
+     bionic, avoiding unnecessary reliance on /dev/random on Linux 5.6 and
      earlier.
--- a/ChangeLog.d/musl_getrandom.txt
+++ b/ChangeLog.d/musl_getrandom.txt
@@ -1,3 +1,5 @@
-Bugfix
-   * Fix PSA crypto initialization needlessly blocking on `/dev/random` on
-     Linux 5.6 and earlier when using a libc such as musl.
+Features
+   * Use the `getrandom` system call for PSA crypto initialization on Linux
+     when supported by the C library, including musl, uClibc and Android
+     bionic, avoiding unnecessary reliance on `/dev/random` on Linux 5.6 and
+     earlier.
--- a/platform/platform_util.c
+++ b/platform/platform_util.c
@@ -291,26 +291,16 @@ int mbedtls_platform_get_entropy(psa_driver_get_entropy_flags_t flags,
 
 /*
  * Test for Linux getrandom() support.
- * Prefer the libc wrapper when sys/random.h is available. This includes
- * libcs such as musl which provide getrandom() but not __GLIBC__.
+ * Use the generic syscall wrapper when sys/syscall.h is available.
  */
 #if defined(__linux__) && defined(__has_include)
-#if __has_include(<sys/random.h>)
-#define MBEDTLS_PLATFORM_HAS_SYS_RANDOM_H
+#if __has_include(<sys/syscall.h>)
+#define MBEDTLS_PLATFORM_HAS_SYS_SYSCALL_H
 #endif
 #endif
 
-#if defined(MBEDTLS_PLATFORM_HAS_SYS_RANDOM_H)
-#include <errno.h>
-#include <sys/random.h>
-#define HAVE_GETRANDOM
-
-static int getrandom_wrapper(void *buf, size_t buflen, unsigned int flags)
-{
-    return (int) getrandom(buf, buflen, flags);
-}
-#elif ((defined(__linux__) && defined(__GLIBC__)) || defined(__midipix__))
-/* Use the generic syscall wrapper when the libc has no getrandom() wrapper. */
+#if defined(MBEDTLS_PLATFORM_HAS_SYS_SYSCALL_H) || \
+    (defined(__linux__) && defined(__GLIBC__)) || defined(__midipix__)
 #include <unistd.h>
 #include <sys/syscall.h>
 #if defined(SYS_getrandom)
@@ -328,9 +318,9 @@ static int getrandom_wrapper(void *buf, size_t buflen, unsigned int flags)
     return (int) syscall(SYS_getrandom, buf, buflen, flags);
 }
 #endif /* SYS_getrandom */
-#endif /* getrandom() wrapper or generic syscall wrapper */
+#endif /* sys/syscall.h or a known generic syscall wrapper */
 
-#undef MBEDTLS_PLATFORM_HAS_SYS_RANDOM_H
+#undef MBEDTLS_PLATFORM_HAS_SYS_SYSCALL_H
 
 #if defined(__FreeBSD__) || defined(__DragonFly__)
 #include <sys/param.h>
--- a/platform/platform_util.c
+++ b/platform/platform_util.c
@@ -291,10 +291,26 @@ int mbedtls_platform_get_entropy(psa_driver_get_entropy_flags_t flags,
 
 /*
  * Test for Linux getrandom() support.
- * Since there is no wrapper in the libc yet, use the generic syscall wrapper
- * available in GNU libc and compatible libc's (eg uClibc).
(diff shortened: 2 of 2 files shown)


platform/platform_util.c
```diff
@@ -291,10 +291,26 @@ int mbedtls_platform_get_entropy(psa_driver_get_entropy_flags_t flags,
 
 /*
  * Test for Linux getrandom() support.
- * Since there is no wrapper in the libc yet, use the generic syscall wrapper
- * available in GNU libc and compatible libc's (eg uClibc).
+ * Prefer the libc wrapper when sys/random.h is available. This includes
+ * libcs such as musl which provide getrandom() but not __GLIBC__.
  */
-#if ((defined(__linux__) && defined(__GLIBC__)) || defined(__midipix__))
+#if defined(__linux__) && defined(__has_include)
```
gilles-peskine-arm, line 296: This breaks the build with dietlibc.

```
  CC    source/platform/platform_util.c
In file included from source/platform/platform_util.c:305:
/usr/include/x86_64-linux-gnu/sys/random.h: In function ‘getrandom’:
/usr/include/x86_64-linux-gnu/sys/random.h:35:42: error: unknown type name ‘__wur’; did you mean ‘__wsum’?
   35 |                    unsigned int __flags) __wur
      |                                          ^~~~~
      |                                          __wsum
/usr/include/x86_64-linux-gnu/sys/random.h:36:35: error: expected declaration specifiers or ‘...’ before ‘(’ token
   36 |                    __attr_access ((__write_only__, 1, 2));
      |                                   ^
/usr/include/x86_64-linux-gnu/sys/random.h:40:50: error: expected ‘=’, ‘,’, ‘;’, ‘asm’ or ‘__attribute__’ before ‘__wur’
   40 | int getentropy (void *__buffer, size_t __length) __wur
      |                                                  ^~~~~
source/platform/platform_util.c:309:1: error: expected ‘=’, ‘,’, ‘;’, ‘asm’ or __attribute__’ before ‘{’ token
  309 | {
      | ^
source/platform/platform_util.c:387:1: error: parameter ‘mbedtls_platform_dev_random’ is initialized
  387 | const char *mbedtls_platform_dev_random = MBEDTLS_PLATFORM_DEV_RANDOM;
      | ^~~~~
source/platform/platform_util.c:392:1: error: expected ‘=’, ‘,’, ‘;’, ‘asm’ or __attribute__’ before ‘{’ token
  392 | {
      | ^
/usr/include/x86_64-linux-gnu/sys/random.h:34:9: error: old-style parameter declarations in prototyped function definition
   34 | ssize_t getrandom (void *__buffer, size_t __length,
      |         ^~~~~~~~~
source/platform/platform_util.c:447: error: expected ‘{’ at end of input
source/platform/platform_util.c:447: error: control reaches end of non-void function [-Werror=return-type]
cc1: all warnings being treated as errors
make: *** [Makefile:11984: platform/platform_util.o] Error 1
```

(on Ubuntu 24.04)

(I'm pretty sure that's why I didn't do that when I overhauled our random-on-Linux code recently.)

ChangeLog.d/musl_getrandom.txt
```diff
@@ -0,0 +1,3 @@
+Bugfix
```
gilles-peskine-arm: I wouldn't consider this a bug in TF-PSA-Crypto (but a bug in older Linux kernels). We're just not taking advantage of a nicer feature exposed by musl.

gilles-peskine-arm (changes requested):
I'm all for improving the library on musl, but not if it breaks the build elsewhere.

The right thing to do would be to guard the non-portable code with `#if __MUSL__` or some such. But [since this does not and will not exist](https://www.openwall.com/lists/musl/2026/04/11/5), I'm afraid we're stuck. We could add a way to declare “I swear I have `getrandom()`” at compile time. But that's even more complexity in code that's already more complex than I'd like. So I'd want to set up some testing if we do that, and that's complicated.

okning:
Thanks, you're right — checking only for the presence of `<sys/random.h>` is insufficient and causes a regression for dietlibc. I'll remove the automatic detection and reclassify the changelog entry as a feature.

Before updating the implementation, would you be open to an explicit opt-in configuration macro, for example `MBEDTLS_PLATFORM_HAS_GETRANDOM`? The default behavior would remain unchanged, and the libc wrapper would only be used when the integrator explicitly declares it available. I would add build coverage for both the enabled path and the unchanged fallback path, while retaining the musl/QEMU consecutive-initialization validation.

gilles-peskine-arm:
We prefer to keep the number of compile-time options down, but if that's the only way, we'll accept it.

But there's one thing I'm not sure I've tried: maybe the block that uses `syscall(SYS_getrandom, ...)` would work on the platforms where `__has_include(<sys/syscall.h>)` works? Would you mind checking how that goes with dietlibc and musl and anything else you have easy access to?

Ideally we'd like this to be covered in our CI, but that would require more work on our CI images than we have time for at the moment. So we'll settle for manual testing for now.

okning:
That's a great idea—thanks for pointing it out. I replaced the
`<sys/random.h>`/libc-wrapper detection with a guarded raw-syscall path:

- On Linux, use `__has_include(<sys/syscall.h>)` when available.
- Only enable the syscall path when `SYS_getrandom` is defined.
- Retain the existing glibc/Midipix condition for older toolchains.
- Otherwise preserve the existing `/dev/random` fallback.

I also reclassified the changelog entry as a feature.

Validation performed:

- glibc 2.43 / GCC 16.1: build passed; syscall path selected; two
  consecutive initializations passed.
- musl / GCC 12.4: full build passed; syscall path selected; two
  consecutive initializations passed.
- dietlibc 0.34: `<sys/syscall.h>` is present but `SYS_getrandom` is not
  defined; the build passed and the existing fallback remained selected.
  This avoids the `<sys/random.h>` regression.
- uClibc-ng 1.0.54 / GCC 14.3: static build passed; syscall path selected;
  two consecutive initializations passed.
- Android bionic / NDK r29 / API 21 / x86_64: dynamic and static builds
  passed; syscall path selected; the static binary passed two consecutive
  initializations.
- AppleClang 21 Debug build passed, followed by 132/132 CTest tests.

ChangeLog.d/musl_getrandom.txt
```diff
@@ -0,0 +1,5 @@
+Features
+   * Use the `getrandom` system call for PSA crypto initialization on Linux
```
gilles-peskine-arm: Cosmetic: changelog entries are text, not markdown, so they shouldn't use backticks. Also, it's conventional to write getrandom() with parentheses.

gilles-peskine-arm (dismissed):
LGTM

ChangeLog.d/musl_getrandom.txt
okning: I've removed the Markdown backticks and changed it to the conventional getrandom() spelling.

gilles-peskine-arm (approved):
LGTM

davidhorstmann-arm (approved):
LGTM, thanks!

Can you make a backport to 1.1 and Mbed TLS 3.6? Then we can merge.

okning:
Thanks! I’ll prepare the backports to TF-PSA-Crypto 1.1 and Mbed TLS 3.6.

PR #866, opened 2026-08-06
Hybrid Public Key Encryption (HPKE) Functionality

LeonieReichertABB:
## Description

This PR adds Hybrid Public Key Encryption (HPKE), as specified in RFC 9180, as functionality. 

The current implementation supports the base mode using the ciphersuite DHKEM(X25519, HKDF-SHA256), HKDF-SHA256, and AES-128-GCM.  Tests have been added for the supported ciphersuite. Support for the remaining RFC-defined KEM, KDF, and AEAD combinations will be added in follow-up pull requests.

During development, AI assistance by Github Copilot was used to understand the migration from the Mbed TLS 3.6 APIs to version 4 APIs, for writing the changelog and for understanding what defines are required in crypto_adjust_config_enable_builtins.h. 
The remainder of the implementation was written and verified manually.

## PR checklist

- [x] **changelog** provided 
- [x] **framework PR**  not required because:  no changes to framework 
- [x] **TF-PSA-Crypto development PR** provided # 
- [x] **TF-PSA-Crypto 1.1 PR**  not required because: new feature
- [x] **mbedtls development PR**  not required because: crypto only 
- [x] **mbedtls 4.1 PR**  not required because:  crypto only
- [x] **mbedtls 3.6 PR**  not required because: crypto only
- **tests**  provided


ChangeLog.d/hpke-implementation.txt
```diff
@@ -0,0 +1,18 @@
+Features
+   * Add support for HPKE (RFC 9180) in the built-in PSA crypto driver.
```
gilles-peskine-arm, line 2: We wouldn't want to add code to the built-in driver for things that can be done generically above the standard PSA API. Are there any parts of HPKE that cannot be built on top of the PSA API?

If HPKE needs new primitives, these will need to be [added to the API specification](https://github.com/GlobalPlatform/psa-api/issues). We're ok with adopting a specification draft, but the discussion must at least be underway.

ChangeLog.d/hpke-implementation.txt
```diff
@@ -0,0 +1,18 @@
...
+     encryption/decryption, and exporter support for HPKE cipher suites,
+     including validation with RFC test vectors for
+     DHKEM(X25519, HKDF-SHA256), HKDF-SHA256, and AES-128-GCM.
+
+Changes
+   * When PSA_WANT_ALG_HPKE is enabled, automatically request the dependent
+     PSA mechanisms (ECDH with Montgomery-255, HKDF/HKDF-Extract/
+     HKDF-Expand, HMAC-SHA-256, AES, and GCM) so HPKE prerequisites are
+     configured consistently.
+   * Rework HPKE-related built-in selection in
```
gilles-peskine-arm, line 11: This is an internal detail and doesn't belong in a changelog entry.

ChangeLog.d/hpke-implementation.txt
```diff
@@ -0,0 +1,18 @@
+Features
+   * Add support for HPKE (RFC 9180) in the built-in PSA crypto driver.
+     This adds HPKE context setup, encapsulation/decapsulation, authenticated
+     encryption/decryption, and exporter support for HPKE cipher suites,
+     including validation with RFC test vectors for
+     DHKEM(X25519, HKDF-SHA256), HKDF-SHA256, and AES-128-GCM.
+
+Changes
+   * When PSA_WANT_ALG_HPKE is enabled, automatically request the dependent
```
gilles-peskine-arm, line 9: `PSA_WANT_ALG_HPKE` means that `PSA_ALG_HPKE` is enabled, but there is no such algorithm. And HKPE does not seem to fit the PSA algorithm paradigm anyway.

gilles-peskine-arm, line 9: It looks like `PSA_WANT_ALG_HPKE` enables all of the algorithms for one specific HKPE suite, which doesn't make sense: what if users want a different suite?

gilles-peskine-arm (changes requested):
Thank you for offering this contribution. Before we start the review process, we're going to need to discuss the design and planning.

Due to a heavy security load (like many open source projects), we currently have extremely limited bandwidth for functionality. Due to the size of this feature, if Arm participates, we will need a business case to be discussed with our product manager. It is extremely unlikely that this would be prioritized over the basics of PQC support, and that already claims most of our 2027 bandwidth.

In terms of design, if the PSA Crypto API needs to be extended (but I think it doesn't), a discussion needs to happen in GlobalPlatform. Arm can drive the discussion on your behalf if you are not a member (once we have a business case for spending our bandwidth).

Then we need to have a discussion on the design of the software interface where not covered by PSA (e.g. higher-level functions, configuration mechanism), and on the design of the implementation. The discussion would be with a member of the TF-PSA-Crypto maintainer team, probably me.

ChangeLog.d/hpke-implementation.txt
LeonieReichertABB: I assume, this would be defined in crypto_values.h ? I will add it there.

LeonieReichertABB: So far only one ciphersuite is implemented. I want to add support for the other ciphersuites later and would adapt the defines accordingly.

 Would you prefer a different approach?

gilles-peskine-arm: What would `PSA_ALG_HKPE` do? What category of algorithms does it fit under? What functions perform operations using this algorithm?

gilles-peskine-arm: `PSA_WANT_ALG_HPKE` should enable the generic functionality, not all the cipher suites. Like `PSA_WANT_ALG_ECDH` enables ECDH but not all the supported elliptic curves, which need to be enabled separately as desired.

ChangeLog.d/hpke-implementation.txt
LeonieReichertABB: HPKE relies on some mbedtls_mpi functionalities for deriving nonces and therefore requires private access.  All other calls are function calls of the format psa_[function_name].

LeonieReichertABB:
Thank you for all the detailed comments! I will address them one after another and adapt the PR accordingly. 
This is my first time contributing to MbedTLS, I am sorry for any inconveniences.

ChangeLog.d/hpke-implementation.txt
LeonieReichertABB: Maybe my phrasing of built-in is wrong here?

ChangeLog.d/hpke-implementation.txt
LeonieReichertABB: HPKE uses a set of building bocks like a KEM, a hash function and a block cipher to build an encryption scheme that simplifies sending large blocks of data if only a public key of the receiver is available. 
HPKE has its own key schedule, that's why private access is required. 

I am new to this, so I looked at a some PSA_WANT_ALG_* definitions as a blueprint of how to define the algorithm. 

Following that logic (and assuming that HPKE belongs with the built-in functions), this would mean the following changes have to be made.
1) The defines in _crypto_adust_config_enalbe_buitins.h_ need to look as follows: 

```
#if defined(PSA_WANT_ALG_HPKE)
#define PSA_HPKE_C

#if defined(PSA_WANT_ALG_HPKE_DHKEM_X25519_HKDF_SHA256_AES_128_GCM)
#define PSA_WANT_ALG_ECDH
#define PSA_WANT_ECC_MONTGOMERY_255
.....
#endif /* PSA_WANT_ALG_HPKE_DHKEM_X25519_HKDF_SHA256_AES_128_GCM */
#endif /* PSA_WANT_ALG_HPKE *
```
 
2) Both PSA_WANT_ALG_HPKE and PSA_WANT_ALG_HPKE_DHKEM_X25519_HKDF_SHA256_AES_128_GCMneed to be defined in _include/psa/crypto_config.h_. 

3) Correctness checks of enabled features are required in _core/check_crypto_config.h._ 

Is this correct?

PR #868, opened 2026-08-11
CMake: add support for config options

yiwu0b11:
## Description

Similar implementation of https://github.com/Mbed-TLS/mbedtls/pull/10860
Fixes: Fixes #10838.

Add support for customizing TF-PSA-Crypto compile-time configs in CMake with:
  - `TF_PSA_CRYPTO_CONFIG_BASE_FILE`
  - `TF_PSA_CRYPTO_CONFIG_NAME`
  - `TF_PSA_CRYPTO_CONFIG_SET`
  - `TF_PSA_CRYPTO_CONFIG_UNSET`

CMake copies the selected base config into the build tree before applying the transformations. The original file is uncganged.

`TF_PSA_CRYPTO_CONFIG_FILE` still working as the config header passed and cannot be combined with the transformation options.

Generated Mbed TLS and PSA configs are also installed correctly, so installed CMake targets do not reference the producer build tree.

## PR checklist

- [x] **changelog** provided 
- [x] **TF-PSA-Crypto development PR** provided: Here
- [x] **TF-PSA-Crypto 1.1 PR** provided: https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/882
- [x] **mbedtls development PR** provided: https://github.com/Mbed-TLS/mbedtls/pull/10860
- [ ] **mbedtls 4.1 PR** provided: TODO
- [x] **mbedtls 3.6 PR** provided: https://github.com/Mbed-TLS/mbedtls/pull/10912
- **tests**  provided:

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -152,12 +152,24 @@ if(NOT TF_PSA_CRYPTO_AS_SUBPROJECT)
         FORCE)
 endif()
 
-# Make TF_PSA_CRYPTO_CONFIG_FILE and TF_PSA_CRYPTO_USER_CONFIG_FILE into PATHs
+# Make configuration file options into PATHs
 set(TF_PSA_CRYPTO_CONFIG_FILE "" CACHE FILEPATH "TF-PSA-Crypto config file (overrides default).")
 set(TF_PSA_CRYPTO_USER_CONFIG_FILE "" CACHE FILEPATH "TF-PSA-Crypto user config file (appended to default).")
+set(TF_PSA_CRYPTO_CONFIG_BASE_FILE "" CACHE FILEPATH "TF-PSA-Crypto config to copy and optionally transform.")
 set(TF_PSA_CRYPTO_CONFIG_NAME "" CACHE STRING "Named TF-PSA-Crypto configuration (see config.py).")
 set(TF_PSA_CRYPTO_CONFIG_SET "" CACHE STRING "Options to set, separated by semicolons (OPTION or OPTION=VALUE).")
 set(TF_PSA_CRYPTO_CONFIG_UNSET "" CACHE STRING "Options to unset, separated by semicolons.")
+
+if(NOT "${TF_PSA_CRYPTO_CONFIG_FILE}" STREQUAL "" AND
+   (NOT "${TF_PSA_CRYPTO_CONFIG_BASE_FILE}" STREQUAL "" OR
+    NOT "${TF_PSA_CRYPTO_CONFIG_NAME}" STREQUAL "" OR
+    NOT "${TF_PSA_CRYPTO_CONFIG_SET}" STREQUAL "" OR
+    NOT "${TF_PSA_CRYPTO_CONFIG_UNSET}" STREQUAL ""))
+    message(FATAL_ERROR
+        "TF_PSA_CRYPTO_CONFIG_FILE cannot be combined with "
+        "TF_PSA_CRYPTO_CONFIG_BASE_FILE, TF_PSA_CRYPTO_CONFIG_NAME, "
+        "TF_PSA_CRYPTO_CONFIG_SET or TF_PSA_CRYPTO_CONFIG_UNSET.")
+endif()
 
 # Create a symbolic link from ${base_name} in the binary directory
 # to the corresponding path in the source directory.
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -155,7 +155,7 @@ endif()
 # Make configuration file options into PATHs
 set(TF_PSA_CRYPTO_CONFIG_FILE "" CACHE FILEPATH "TF-PSA-Crypto config file (overrides default).")
 set(TF_PSA_CRYPTO_USER_CONFIG_FILE "" CACHE FILEPATH "TF-PSA-Crypto user config file (appended to default).")
-set(TF_PSA_CRYPTO_CONFIG_BASE_FILE "" CACHE STRING "TF-PSA-Crypto config to copy and optionally transform.")
+set(TF_PSA_CRYPTO_CONFIG_BASE_FILE "" CACHE FILEPATH "TF-PSA-Crypto config to copy and optionally transform.")
 set(TF_PSA_CRYPTO_CONFIG_NAME "" CACHE STRING "Named TF-PSA-Crypto configuration (see config.py).")
 set(TF_PSA_CRYPTO_CONFIG_SET "" CACHE STRING "Options to set, separated by semicolons (OPTION or OPTION=VALUE).")
 set(TF_PSA_CRYPTO_CONFIG_UNSET "" CACHE STRING "Options to unset, separated by semicolons.")
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -155,13 +155,9 @@ endif()
 # Make TF_PSA_CRYPTO_CONFIG_FILE and TF_PSA_CRYPTO_USER_CONFIG_FILE into PATHs
 set(TF_PSA_CRYPTO_CONFIG_FILE "" CACHE FILEPATH "TF-PSA-Crypto config file (overrides default).")
 set(TF_PSA_CRYPTO_USER_CONFIG_FILE "" CACHE FILEPATH "TF-PSA-Crypto user config file (appended to default).")
-set(TF_PSA_CRYPTO_CONFIG_NAME "" CACHE STRING "Name of the configuration to use (see config.py documentation)")
-
-if(NOT "${TF_PSA_CRYPTO_CONFIG_NAME}" STREQUAL "")
-    if(NOT "${TF_PSA_CRYPTO_CONFIG_FILE}" STREQUAL "")
-        message(FATAL_ERROR "TF_PSA_CRYPTO_CONFIG_NAME and TF_PSA_CRYPTO_CONFIG_FILE are incompatible.")
-    endif()
-endif(NOT "${TF_PSA_CRYPTO_CONFIG_NAME}" STREQUAL "")
+set(TF_PSA_CRYPTO_CONFIG_NAME "" CACHE STRING "Named TF-PSA-Crypto configuration (see config.py).")
+set(TF_PSA_CRYPTO_CONFIG_SET "" CACHE STRING "Options to set, separated by semicolons (OPTION or OPTION=VALUE).")
+set(TF_PSA_CRYPTO_CONFIG_UNSET "" CACHE STRING "Options to unset, separated by semicolons.")
 
 # Create a symbolic link from ${base_name} in the binary directory
 # to the corresponding path in the source directory.
@@ -425,6 +421,10 @@ set(TF_PSA_CRYPTO_PRIVATE_INCLUDE_DIRS
 #
 add_subdirectory(framework)
 add_subdirectory(include)
+if(TF_PSA_CRYPTO_AS_SUBPROJECT AND TF_PSA_CRYPTO_CONFIG_TRANSFORMED)
+    set(TF_PSA_CRYPTO_CONFIG_FILE "${TF_PSA_CRYPTO_CONFIG_FILE}"
+        PARENT_SCOPE)
+endif()
 add_subdirectory(drivers)   # must precede core
 add_subdirectory(extras)
 add_subdirectory(platform)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -155,20 +155,23 @@ endif()
 # Make configuration file options into PATHs
 set(TF_PSA_CRYPTO_CONFIG_FILE "" CACHE FILEPATH "TF-PSA-Crypto config file (overrides default).")
 set(TF_PSA_CRYPTO_USER_CONFIG_FILE "" CACHE FILEPATH "TF-PSA-Crypto user config file (appended to default).")
-set(TF_PSA_CRYPTO_CONFIG_BASE_FILE "" CACHE FILEPATH "TF-PSA-Crypto config to copy and optionally transform.")
+set(TF_PSA_CRYPTO_CONFIG_BASE_FILE "" CACHE STRING "TF-PSA-Crypto config to copy and optionally transform.")
 set(TF_PSA_CRYPTO_CONFIG_NAME "" CACHE STRING "Named TF-PSA-Crypto configuration (see config.py).")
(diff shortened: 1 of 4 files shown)


tests/scripts/components-build-system.sh
```diff
@@ -164,6 +164,25 @@ component_tf_psa_crypto_build_config_name () {
...
+          -DTF_PSA_CRYPTO_CONFIG_UNSET=PSA_WANT_ALG_RIPEMD160 \
+          '-DTF_PSA_CRYPTO_CONFIG_SET=PSA_WANT_ALG_XTS;PSA_WANT_ALG_SHA_256=2' \
+          "$TF_PSA_CRYPTO_ROOT_DIR"
+    cmake --build . --target tfpsacrypto
+
+    grep -E '^#define PSA_WANT_ALG_SHA_256[[:space:]]+2$' \
+         include/psa/crypto_config.h
+    grep -E '^#define PSA_WANT_ALG_XTS[[:space:]]+1$' \
+         include/psa/crypto_config.h
+    not grep -q '^#define PSA_WANT_ALG_RIPEMD160' include/psa/crypto_config.h
```
gilles-peskine-arm, line 219: Please also assert that `configs/crypto-config-symmetric-only.h` is not modified.

tests/scripts/components-build-system.sh
```diff
@@ -164,6 +164,25 @@ component_tf_psa_crypto_build_config_name () {
     rm -f include/psa/crypto_config_full.h
 }
 
+component_tf_psa_crypto_build_config_options () {
+    msg "build: cmake with a base config, TF_PSA_CRYPTO_CONFIG_SET and TF_PSA_CRYPTO_CONFIG_UNSET"
+    cd "$OUT_OF_SOURCE_DIR"
+    cmake -DTF_PSA_CRYPTO_CONFIG_FILE=configs/crypto-config-symmetric-only.h \
+          -DTF_PSA_CRYPTO_CONFIG_UNSET=PSA_WANT_ALG_RIPEMD160 \
+          '-DTF_PSA_CRYPTO_CONFIG_SET=PSA_WANT_ALG_XTS;PSA_WANT_ALG_SHA_256=2' \
```
gilles-peskine-arm: Rather than XTS, which is currently not implemented but will be implemented later, it would be better to check something that is in the default config but not in `crypto-config-symmetric-only.h`, for example `PSA_WANT_KEY_TYPE_DH_KEY_PAIR_BASIC`.

bjwtaylor (dismissed):

include/CMakeLists.txt
```diff
@@ -1,28 +1,78 @@
 option(INSTALL_TF_PSA_CRYPTO_HEADERS "Install TF PSA Crypto headers." ON)
 
-# The handling of TF_PSA_CRYPTO_CONFIG_NAME assumes that the default
-# configuration file 'include/psa/crypto_config.h' exists in the source tree.
-# That file is copied into the build tree and then modified with 'config.py'
-# according to the configuration name provided via TF_PSA_CRYPTO_CONFIG_NAME.
+# Generate a build-tree configuration when transformations are requested.
+if(TF_PSA_CRYPTO_CONFIG_NAME OR TF_PSA_CRYPTO_CONFIG_SET OR
+   TF_PSA_CRYPTO_CONFIG_UNSET)
```
gilles-peskine-arm: It's an edge case, but [some strings are interpreted as false](https://cmake.org/cmake/help/latest/command/if.html#constant). For example `cmake -DTF_PSA_CRYPTO_CONFIG_SET=NO` is equivalent to leaving it empty, not to defining `NO`.

I think the correct check is `if ("" STREQUAL "${TF_PSA_CRYPTO_CONFIG_NAME}" OR ...)` but even this might not be quite right, I'm not a CMake expert.

include/CMakeLists.txt
```diff
@@ -1,28 +1,78 @@
...
+            "TF-PSA-Crypto configuration file not found: ${TF_PSA_CRYPTO_BASE_CONFIG_FILE}")
+    endif()
 
-if(NOT "${TF_PSA_CRYPTO_CONFIG_NAME}" STREQUAL "")
     file(MAKE_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/psa)
-    file(COPY ${PROJECT_SOURCE_DIR}/include/psa/crypto_config.h
-         DESTINATION ${CMAKE_CURRENT_BINARY_DIR}/psa)
-
-    message(STATUS "Setting ${TF_PSA_CRYPTO_CONFIG_NAME} configuration")
+    set(TF_PSA_CRYPTO_GENERATED_CONFIG_FILE
```
gilles-peskine-arm, line 43: When doing an in-tree build, if NAME/SET/UNSET is set, `include/psa/crypto_config.h` gets modified even if `TF_PSA_CRYPTO_CONFIG_FILE` is set. This is surprising behavior.

I only find NAME/SET/UNSET useful to make multiple build trees with different configurations. Otherwise I'd just run `scripts/config.py`. So I propose to reject them when doing an in-tree build.

include/CMakeLists.txt
```diff
@@ -1,28 +1,78 @@
 option(INSTALL_TF_PSA_CRYPTO_HEADERS "Install TF PSA Crypto headers." ON)
 
-# The handling of TF_PSA_CRYPTO_CONFIG_NAME assumes that the default
-# configuration file 'include/psa/crypto_config.h' exists in the source tree.
-# That file is copied into the build tree and then modified with 'config.py'
-# according to the configuration name provided via TF_PSA_CRYPTO_CONFIG_NAME.
+# Generate a build-tree configuration when transformations are requested.
+if(TF_PSA_CRYPTO_CONFIG_NAME OR TF_PSA_CRYPTO_CONFIG_SET OR
+   TF_PSA_CRYPTO_CONFIG_UNSET)
+    if(TF_PSA_CRYPTO_CONFIG_FILE)
```
gilles-peskine-arm: Since the introduction of `TF_PSA_CRYPTO_CONFIG_NAME`, the CMake variable `TF_PSA_CRYPTO_CONFIG_FILE` has two meanings:

* It causes `-DTF_PSA_CRYPTO_CONFIG_FILE=...` to be passed in `CFLAGS` when invoking the C compiler.
* It is the basis for the configuration, which may be modified according to `TF_PSA_CRYPTO_CONFIG_NAME`, and now also `TF_PSA_CRYPTO_CONFIG_SET` and `TF_PSA_CRYPTO_CONFIG_UNSET`.

Using a non-default base config file is not really useful with a config name, but it is definitely useful with set/unset. But having the two meanings is a bit confusing, and it means that [install doesn't quite work](https://github.com/Mbed-TLS/mbedtls/pull/10860#discussion_r3775542636).

How about using a different variable name for the two meanings?

* `TF_PSA_CRYPTO_CONFIG_FILE` is what passed in `CFLAGS`.
* `TF_PSA_CRYPTO_CONFIG_BASE_FILE` is the base for NAME/SET/UNSET.
* `TF_PSA_CRYPTO_CONFIG_FILE` is incompatible with `BASE_FILE`, `NAME`, `SET` and `UNSET`.

Then I think there's no surprising behavior, but all the useful behavior is available.

gilles-peskine-arm (changes requested):
Thank you very much for implementing this! I'm happy that CMake can now do most of what I use `mbedtls-prepare-build` for on a daily basis.

There are a few corner cases that I personally don't care much about, but that are likely to cause problems: in-tree builds, config file name confusion.

Other than that, I'd prefer some improvements to the testing, but it isn't critical.

tests/scripts/components-build-system.sh
yiwu0b11: Added a copy-and-compare check

tests/scripts/components-build-system.sh
yiwu0b11: done

include/CMakeLists.txt
yiwu0b11: done, it now uses `if(NOT "${TF_PSA_CRYPTO_CONFIG_BASE_FILE}" STREQUAL "")`

include/CMakeLists.txt
yiwu0b11: done, BASE_FILE, NAME, SET and UNSET are rejected for an in-tree build

include/CMakeLists.txt
yiwu0b11: done, yeah I agree this makes it cleaner

include/CMakeLists.txt
```diff
@@ -1,28 +1,89 @@
...
+        NOT "${TF_PSA_CRYPTO_CONFIG_SET}" STREQUAL "" OR
+        NOT "${TF_PSA_CRYPTO_CONFIG_UNSET}" STREQUAL "") AND
+       "${TF_PSA_CRYPTO_PYTHON_EXECUTABLE}" STREQUAL "")
+        message(FATAL_ERROR
+            "Python 3 is required to generate a TF-PSA-Crypto configuration file.")
+    endif()
+
+    if(NOT "${TF_PSA_CRYPTO_CONFIG_BASE_FILE}" STREQUAL "")
+        set(TF_PSA_CRYPTO_BASE_CONFIG_FILE
+            "${TF_PSA_CRYPTO_CONFIG_BASE_FILE}")
```
gilles-peskine-arm: Why use two variables with a similar-looking name `TF_PSA_CRYPTO_BASE_CONFIG_FILE` and `TF_PSA_CRYPTO_CONFIG_BASE_FILE`? It would make sense to use `"${PROJECT_SOURCE_DIR}/include/psa/crypto_config.h"` as the default value of `TF_PSA_CRYPTO_CONFIG_BASE_FILE` if it's unset.

bjwtaylor (dismissed):

CMakeLists.txt
```diff
@@ -152,16 +152,24 @@ if(NOT TF_PSA_CRYPTO_AS_SUBPROJECT)
...
 set(TF_PSA_CRYPTO_CONFIG_FILE "" CACHE FILEPATH "TF-PSA-Crypto config file (overrides default).")
 set(TF_PSA_CRYPTO_USER_CONFIG_FILE "" CACHE FILEPATH "TF-PSA-Crypto user config file (appended to default).")
-set(TF_PSA_CRYPTO_CONFIG_NAME "" CACHE STRING "Name of the configuration to use (see config.py documentation)")
-
-if(NOT "${TF_PSA_CRYPTO_CONFIG_NAME}" STREQUAL "")
-    if(NOT "${TF_PSA_CRYPTO_CONFIG_FILE}" STREQUAL "")
-        message(FATAL_ERROR "TF_PSA_CRYPTO_CONFIG_NAME and TF_PSA_CRYPTO_CONFIG_FILE are incompatible.")
-    endif()
-endif(NOT "${TF_PSA_CRYPTO_CONFIG_NAME}" STREQUAL "")
+set(TF_PSA_CRYPTO_CONFIG_BASE_FILE "" CACHE FILEPATH "TF-PSA-Crypto config to copy and optionally transform.")
```
gilles-peskine-arm: Optional and partly preexisting — we can handle this in a follow-up.

I don't think the documentation of `TF_PSA_CRYPTO_CONFIG_FILE`, `TF_PSA_CRYPTO_USER_CONFIG_FILE` and `TF_PSA_CRYPTO_CONFIG_BASE_FILE` makes it clear which one to use when.

We should either improve the documentation of these options, or document it elsewhere (there's already some documentation in `README.md` which we should expand), or both.

gilles-peskine-arm (dismissed):
LGTM

I have a couple of remarks but they can be handled later.

include/CMakeLists.txt
yiwu0b11: `TF_PSA_CRYPTO_CONFIG_BASE_FILE` is the public user-selected base as before. 
A new `TF_PSA_CRYPTO_CONFIG_BASE_FILE_OVERRIDE` is used only for the intermediate configuration produced by MbedTLS parent.

CMakeLists.txt
```diff
@@ -155,20 +155,23 @@ endif()
 # Make configuration file options into PATHs
 set(TF_PSA_CRYPTO_CONFIG_FILE "" CACHE FILEPATH "TF-PSA-Crypto config file (overrides default).")
 set(TF_PSA_CRYPTO_USER_CONFIG_FILE "" CACHE FILEPATH "TF-PSA-Crypto user config file (appended to default).")
-set(TF_PSA_CRYPTO_CONFIG_BASE_FILE "" CACHE FILEPATH "TF-PSA-Crypto config to copy and optionally transform.")
+set(TF_PSA_CRYPTO_CONFIG_BASE_FILE "" CACHE STRING "TF-PSA-Crypto config to copy and optionally transform.")
```
gilles-peskine-arm: Why `STRING`? I think `FILEPATH` is correct. This is a file that is consumed by CMake. As far as I understand the documentation, `FILEPATH` means that CMake will convert it to an absolute path, relative to the directory where CMake runs. It's not ideal because that means you can't relocate the build tree, but that's a CMake defect, it's not our job to fix that. Making it a STRING means we have to handle relative paths ourselves, which I don't think is being done correctly:

```
# start in the tf-psa-crypto source root
mkdir foo
cd foo
cmake -DTF_PSA_CRYPTO_CONFIG_BASE_FILE=../configs/crypto-config-ccm-aes-sha256.h ..
-- The C compiler identification is GNU 13.3.0
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /usr/bin/cc - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Found Python3: /home/gilpes01/Packages/_/x86.64_Linux/bin/python3 (found version "3.12.3") found components: Interpreter 
CMake Error at include/CMakeLists.txt:36 (message):
  TF-PSA-Crypto configuration file not found:
  /home/gilpes01/z/dev/crypto/reviewing/configs/crypto-config-ccm-aes-sha256.h


-- Configuring incomplete, errors occurred!
```

gilles-peskine-arm (changes requested):
8604f6b6b50b89d894f5d42db3958b194844f4f3 broke `TF_PSA_CRYPTO_CONFIG_BASE_FILE` with a relative path (it worked on 34538973d123e1995acf691606475bea205f1e0f). LGTM otherwise.

gilles-peskine-arm (approved):
LGTM

bjwtaylor (approved):

PR #875, opened 2026-08-26
Add design assumption for the use of PSA_ITS_STORAGE_PREFIX

bjwtaylor:
## Description

Add design assumption for the use of PSA_ITS_STORAGE_PREFIX

## PR checklist

- [ ] **changelog** provided
- [ ] **TF-PSA-Crypto 1.1 PR** provided # | not required because: TODO if required
- [ ] **mbedtls development PR** not required because: crypto only
- [ ] **mbedtls 4.1 PR** not required because: crypto only
- [ ] **mbedtls 3.6 PR**not required because: crypto only 
- **tests**  not required because: docs only


include/psa/crypto_config.h
```diff
@@ -1287,6 +1287,14 @@
  * Enable the emulation of the Platform Security Architecture
  * Internal Trusted Storage (PSA ITS) over files.
  *
+ * This implementation does not support concurrent access by multiple
+ * processes. Each process must use a separate storage directory.
+ *
+ * The storage directory is determined when the library is built. By default,
+ * files are stored in the process's current working directory. To select a
+ * different directory, define PSA_ITS_STORAGE_PREFIX as a string containing
```
gilles-peskine-arm: Minor: link to the documentation of the option:
```suggestion
 * different directory, define #PSA_ITS_STORAGE_PREFIX as a string containing
```

gilles-peskine-arm (dismissed):
LGTM

gilles-peskine-arm (dismissed):
LGTM

include/psa/crypto_config.h
bjwtaylor: I think this is causing a doxygen warning and the ci to fail. I've put something up that I think might resolve it, let's see what the CI does though and let me know what you think?

gilles-peskine-arm: Ah, my bad, I'd missed that `PSA_ITS_STORAGE_PREFIX` is not an official configuration option.

Since it's now referenced in public documentation, we should make it an official option. It means we commit to keep supporting it, but that's ok by me, I see no reason to retire or change it.

We should make it match our current naming pattern, however. The only `PSA_xxx` options we have are things that could potentially become a PSA standard. This one isn't. So we should call it either `TF_PSA_CRTYPTO_ITS_STORAGE_PREFIX` or `MBEDTLS_PSA_ITS_STORAGE_PREFIX`. I prefer `MBEDTLS_PSA_ITS_STORAGE_PREFIX` for consistency with the option name.

We should still recognize `PSA_ITS_STORAGE_PREFIX` as a fallback if `MBEDTLS_PSA_ITS_STORAGE_PREFIX` is not defined. Strictly speaking we could retire it at any time, but there is at least [nuttx-apps](https://github.com/open-vela/nuttx-apps) using it, and I see no reason to gratuitously break them. `PSA_ITS_STORAGE_PREFIX` shouldn't be documented except in a comment in `psa_its_file.c`.

bjwtaylor: I've put up some changes let's see if the CI likes them.

ChangeLog.d/psa-its-storage-prefix.txt
```diff
@@ -0,0 +1,4 @@
+Features
+   * Allow applications using file-based PSA ITS to configure the storage
+     directory with MBEDTLS_PSA_ITS_STORAGE_PREFIX. This allows each process
+     to use a separate directory where concurrent access is not supported.
```
gilles-peskine-arm: “where concurrent access is not supported” is ambiguous. The most natural grammatical interpretation is that “where” refers to the separate directory, and concurrent access is not supported there, with the implication that it would be supported elsewhere — although there is no semantic basis for thinking that concurrent access would be supported elsewhere, and indeed there is no other location where concurrent access is supported. It could also refer to platforms where concurrent access is not supported, as opposed to platforms where concurrent access is supported — but in fact multi-process concurrent access is not supported on any platform. What is true is that, independently of this change, multi-process concurrent access is not supported.

We don't necessarily need to mention that in the changelog entry (since that hasn't changed), but we could. Suggested wording:

> Allow applications using file-based PSA ITS to configure the storage directory with MBEDTLS_PSA_ITS_STORAGE_PREFIX.

or

> Allow applications using file-based PSA ITS to configure the storage directory with MBEDTLS_PSA_ITS_STORAGE_PREFIX. The default remains the current directory. Note that each concurrent process needs a separate directory.

ChangeLog.d/psa-its-storage-prefix.txt
```diff
@@ -0,0 +1,4 @@
+Features
+   * Allow applications using file-based PSA ITS to configure the storage
```
gilles-peskine-arm, line 2: “File-based PSA ITS” is correct, but it might be obsure terminology. The ostensible API feature is persistent keys combined with the compilation option `MBEDTLS_PSA_ITS_FILE_C` which is enabled by default. Suggested wording:

> Officially provide a compile-time option MBEDTLS_PSA_ITS_STORAGE_PREFIX to select the directory for persistent keys when using the default file-based backend. …

docs/architecture/mbed-crypto-storage-specification.md
```diff
@@ -182,10 +182,10 @@ Changes since [1.0.0](#mbed-crypto-1.0.0):
 
 Assumption: C stdio, allowing names containing lowercase letters, digits and underscores, of length up to 23.
 
-An undocumented build-time configuration value `PSA_ITS_STORAGE_PREFIX` allows storing the key files in a directory other than the current directory. This value is simply prepended to the file name (so it must end with a directory separator to put the keys in a different directory).
+The build-time configuration value `MBEDTLS_PSA_ITS_STORAGE_PREFIX` allows storing the key files in a directory other than the current directory. This value is simply prepended to the file name (so it must end with a directory separator to put the keys in a different directory).
```
gilles-peskine-arm: This document keeps a historical record, because users might upgrade an existing system from an older version. The old versions referenced here did not support `MBEDTLS_PSA_ITS_STORAGE_PREFIX`, so these sections should not change. The information about the change needs to go into a new section.

Given that the only thing that's changing is the new option name, we don't need a full copy of the section, just the new option name and a reference to the existing section which remains fully applicable.

include/psa/crypto_config.h
```diff
@@ -1281,12 +1281,29 @@
...
+//#define MBEDTLS_PSA_ITS_STORAGE_PREFIX ""
+
 /**
  * \def MBEDTLS_PSA_ITS_FILE_C
  *
  * Enable the emulation of the Platform Security Architecture
  * Internal Trusted Storage (PSA ITS) over files.
  *
+ * This implementation does not support concurrent access by multiple
+ * processes. Each process must use a separate storage directory. Configure
```
gilles-peskine-arm, line 1304: We may want to mention that multithreaded access within the same process is supported.

gilles-peskine-arm (changes requested):
LGTM for the code, needs work for the documentation

docs/architecture/mbed-crypto-storage-specification.md
bjwtaylor: I've added a suggestion, let me know what you think.

include/psa/crypto_config.h
bjwtaylor: I've added a clarification.

ChangeLog.d/psa-its-storage-prefix.txt
```diff
@@ -0,0 +1,8 @@
...
+     directory with MBEDTLS_PSA_ITS_STORAGE_PREFIX. The default remains the
+     current directory. Note that each concurrent process needs a separate
+     directory.
```
gilles-peskine-arm, line 5: This needs to be removed in favor of the rewritten version below.

docs/architecture/mbed-crypto-storage-specification.md
```diff
@@ -470,3 +470,7 @@ The layout of a transaction file is:
 ## TF-PSA-Crypto 1.0
 
 The dynamic secure element interface `MBEDTLS_PSA_CRYPTO_SE_C` is removed. As a consequence files identifiers 0xfffffe02 through 0xfffffeff are no more used by the secure element driver storage.
+
+### File namespace on stdio for TF-PSA-Crypto 1.0
```
gilles-peskine-arm, line 474:
```suggestion
## TF-PSA-Crypto 1.3

### Configuration for the file namespace on stdio since TF-PSA-Crypto 1.3.0

```

gilles-peskine-arm (changes requested):
LGTM except a couple of minor documentation problems. Also the CI is unhappy.

bjwtaylor:
Rebased in order to try to resolve the CI failures

ChangeLog.d/psa-its-storage-prefix.txt
gilles-peskine-arm: This still needs to be removed in favor of the rewritten version below.

gilles-peskine-arm (changes requested):
LGTM except the changelog

ChangeLog.d/psa-its-storage-prefix.txt
bjwtaylor: Apologies, I lost track of what you were after in the github thread. Hopefully I've understood it now and have uploaded the changes. Let me know what you think though?

gilles-peskine-arm (approved):
LGTM

PR #880, opened 2026-08-26
Enhance the PSA driver test jinja templates for hashing, AES, signature and key agreement functions

oubaid-a:
## Description

Enhance the jinja templates for the transparent driver using the MbedTLS transparent driver test:

* Enhance OS-template-transparent.jinja to:
  * Support fallback behavior when the driver do not fully support a cryptographic mechanism
  * Handle driver ID and and operation context for multipart operations

* Update psa_crypto_driver_wrappers.h.jinja to provide entry point for:
  * HASH operations including multi-part operations
  * AES cipher and AEAD encryption and decryption operations
  * Hash and message signature generation and verification operations
  * Key agreement operation.

* Rename the test_driver_ctx to in  crypto_driver_contexts_primitives.h to align with the transparent driver templates:
  * mbedtls_test_driver_ctx : driver.prefix + _driver_ctx

## PR checklist

- [x] **changelog** not required 
- [x] **framework PR**  not required
- [x] **TF-PSA-Crypto development PR** provided here 
- [x] **TF-PSA-Crypto 1.1 PR** provided https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/880
- [x] **mbedtls 3.6 PR** not required
- **tests**  not required because: No regression, generated files are the same

### DIFF
--- a/include/psa/crypto_driver_contexts_primitives.h
+++ b/include/psa/crypto_driver_contexts_primitives.h
@@ -110,7 +110,7 @@ typedef union {
     unsigned dummy; /* Make sure this union is always non-empty */
     mbedtls_psa_hash_operation_t mbedtls_ctx;
 #if defined(PSA_CRYPTO_DRIVER_TEST)
-    mbedtls_transparent_test_driver_hash_operation_t test_driver_ctx;
+    mbedtls_transparent_test_driver_hash_operation_t mbedtls_test_driver_ctx;
 #endif
 } psa_driver_hash_context_t;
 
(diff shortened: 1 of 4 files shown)


gilles-peskine-arm:
Thank you very much for the contribution!

Note that to allow us to take your code, you need to accept the developer certificate of origin in `dco.txt` by signing off your commits (`git rebase --signoff …`).

Also, note that we want to get rid of the Jinja dependency. We're planning to rewrite the Python script to not use a text template, and instead use the `psa_wrapper.py` module which is a flexible generator of C code that wraps around/inside PSA API functions. This has been a long-term wish that we didn't have time for, but we have just found out that we might have time for that next quarter, although it isn't certain yet. But it means that effort on the Jinja template and code might not be very efficient.

PR #897, opened 2026-09-06
Fix built-in MAC operation initialization

romart:
## Description

Zero-initialize the full built-in MAC context. Add a regression test that fails before the fix and passes afterward across six GCC optimization levels, including -O0. The

## PR checklist

  - [x] **changelog** not required because: internal initializer fix; no public API behavior change demonstrated.
  - [x] **TF-PSA-Crypto development PR** provided
  - [ ] **TF-PSA-Crypto 1.1 PR** applicability pending review
  - [ ] **mbedtls development PR** applicability pending review
  - [ ] **mbedtls 4.1 PR** applicability pending review
  - [ ] **mbedtls 3.6 PR** applicability pending review
  - **tests** provided: regression fails before the fix and passes afterward with GCC at -O0, -O1, -O2, -O3, -Os, and -Og. The PSA crypto suite passes at -O0.

### DIFF
--- a/drivers/builtin/include/mbedtls/private/crypto_builtin_composites.h
+++ b/drivers/builtin/include/mbedtls/private/crypto_builtin_composites.h
@@ -69,7 +69,11 @@ typedef struct {
     } MBEDTLS_PRIVATE(ctx);
 } mbedtls_psa_mac_operation_t;
 
-#define MBEDTLS_PSA_MAC_OPERATION_INIT { 0, { 0 } }
+/* Leave ctx implicitly initialized so the whole union is zeroed
+ * (C17 6.7.9 paragraphs 10 and 21). An explicit { 0 } initializes only
+ * dummy (6.7.9 paragraph 17), leaving bytes beyond it unspecified
+ * (6.2.6.1 paragraph 7; see also DR 283). */
+#define MBEDTLS_PSA_MAC_OPERATION_INIT { 0 }
 
 #if defined(MBEDTLS_PSA_BUILTIN_ALG_GCM) || \
     defined(MBEDTLS_PSA_BUILTIN_ALG_CCM) || \
--- a/tests/suites/test_suite_psa_crypto.data
+++ b/tests/suites/test_suite_psa_crypto.data
@@ -1774,6 +1774,10 @@ hash_clone_target_state:
 MAC operation object initializers zero properly
 mac_operation_init:
 
+Built-in MAC initializer clears a nonzero HMAC context
+depends_on:MBEDTLS_PSA_BUILTIN_ALG_HMAC
+builtin_mac_operation_init:
+
 PSA MAC setup: good, HMAC-SHA-256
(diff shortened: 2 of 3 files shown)


drivers/builtin/include/mbedtls/private/crypto_builtin_composites.h
```diff
@@ -69,7 +69,11 @@ typedef struct {
     } MBEDTLS_PRIVATE(ctx);
 } mbedtls_psa_mac_operation_t;
 
-#define MBEDTLS_PSA_MAC_OPERATION_INIT { 0, { 0 } }
+/* Leave ctx implicitly initialized so the whole union is zeroed
+ * (C17 6.7.9 paragraphs 10 and 21). An explicit { 0 } initializes only
+ * dummy (6.7.9 paragraph 17), leaving bytes beyond it unspecified
+ * (6.2.6.1 paragraph 7; see also DR 283). */
+#define MBEDTLS_PSA_MAC_OPERATION_INIT { 0 }
```
gilles-peskine-arm, line 76: We use full initializers because older versions of Clang don't like partial initializers with `-Wextra` (specifically `-Wmissing-field-initializers`). (See all the failures on the CI.)

drivers/builtin/include/mbedtls/private/crypto_builtin_composites.h
```diff
@@ -69,7 +69,11 @@ typedef struct {
     } MBEDTLS_PRIVATE(ctx);
 } mbedtls_psa_mac_operation_t;
 
-#define MBEDTLS_PSA_MAC_OPERATION_INIT { 0, { 0 } }
+/* Leave ctx implicitly initialized so the whole union is zeroed
+ * (C17 6.7.9 paragraphs 10 and 21). An explicit { 0 } initializes only
+ * dummy (6.7.9 paragraph 17), leaving bytes beyond it unspecified
+ * (6.2.6.1 paragraph 7; see also DR 283). */
```
gilles-peskine-arm, line 75: Unfortunately, leaving the union explicitly initialized doesn't help according to the C standard. ¶10 states:

> if it is a union, the first named member is initialized (recursively) according to these
rules, and any padding is initialized to zero bits;

The content of the union that does not overlap with the first member remains uninitialized. (This content is not padding. Padding is memory content that is not part of the value of anything, not memory content that is part of an inactive union member.)

In practice, `{ 0 }` vs `{ 0, { 0 } }` does seem to make a difference with GCC 15 (the most popular compiler that actually only initializes the first (or designated) member of a union.) So, even though it isn't guaranteed, this does make the code more robust in case we forget to explicitly zero out the whole union somewhere. However, we can't just break the build with older Clang.

gilles-peskine-arm (changes requested):
This breaks older clang.

This is just a minor robustness improvement, since there could be compilers where it doesn't help. (I'm not aware of one. Maybe Compcert, which has been only initializing the first member of unions for a long time?) So I'm not convinced it's worth making the code more complex over this.

PR #898, opened 2026-09-07
Add support for Blake2 MAC

valeriosetti:
## Description

That's the follow-up of https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/867 and it completes the support for Blake2 in tf-psa-crypto.

This PR depends on https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/867 and therefore most of the initial commits come from that PR.

## Current limitations

Only positive testing has been added so far. Negative testing is TODO.

## PR checklist

- [ ] **changelog** provided: TODO
- [ ] **TF-PSA-Crypto development PR** provided: this one
- [ ] **TF-PSA-Crypto 1.1 PR** not required because: new feature
- [ ] **mbedtls development PR** not required because: hopefully not, but we'll see what the CI thinks
- [ ] **mbedtls 4.1 PR** not required because: new feature
- [ ] **mbedtls 3.6 PR** not required because: new feature
- **tests**  provided


drivers/builtin/src/psa_crypto_mac.c
```diff
@@ -153,6 +154,111 @@ static psa_status_t psa_hmac_finish_internal(
...
+static psa_status_t psa_blake2_mac_finish_internal(mbedtls_psa_blake2_mac_operation_t *op,
+                                                   psa_algorithm_t alg,
+                                                   uint8_t *mac,
+                                                   size_t mac_size)
+{
+    psa_algorithm_t hash_alg = PSA_ALG_BLAKE2_MAC_GET_HASH(alg);
+
+#if defined(MBEDTLS_PSA_BUILTIN_ALG_BLAKE2S_HASH256)
+    if (hash_alg == PSA_ALG_BLAKE2S_HASH256) {
+        tf_psa_crypto_blake2s_finish(&op->ctx.s, mac, mac_size);
```
bjwtaylor, line 224: This call and the call below do not capture the return value. This function can return TF_PSA_CRYPTO_ERR_BUFFER_TOO_SMALL. This is in itself not good and will need fixing, however I think it will mean that you can manipulate the tag to perform a authentication bypass of the PSA MAC boundary by passing multiple forged tags. It may also be possible and advisable to initialise the keyed blake2 context with the requested tag length to mitigate this further. What do you think?

include/psa/crypto_values.h
bjwtaylor, line 442: Do we need to define PSA_KEY_TYPE_BLAKE2 here as per: https://arm-software.github.io/psa-api/crypto/1.5/api/ops/mac.html#c.PSA_ALG_BLAKE2_MAC

core/psa_crypto.c
bjwtaylor, line 577: Do we need to have a PSA_KEY_TYPE_BLAKE2 case here to prevent BLAKE2 keys returning PSA_ERROR_INVALID_ARGUMENT when a import is performed or is that out of scope?

scripts/data_files/config-options-current.txt
```diff
@@ -131,6 +131,8 @@ MBEDTLS_THREADING_ALT
 MBEDTLS_THREADING_C
 MBEDTLS_THREADING_PTHREAD
 PSA_CRYPTO_CONFIG_H
+PSA_WANT_ALG_BLAKE2B_HASH512
+PSA_WANT_ALG_BLAKE2S_HASH256
```
bjwtaylor, line 135: Do we also need PSA_WANT_ALG_BLAKE2_MAC here, or is it expected to be added later?

ChangeLog.d/blake2.txt
```diff
@@ -0,0 +1,4 @@
+Features
+   * Support for Blake2 hash algorithms Blake2s and Blake2p has been added.
+     They are enabled via PSA_WANT_ALG_BLAKE2S_HASH256 and
+     PSA_WANT_ALG_BLAKE2B_HASH512 respectively.
```
bjwtaylor, line 4: Should PSA_WANT_ALG_BLAKE2_MAC be added here, or is it deliberately ommited?

bjwtaylor (commented):
Still need to do a full review, but some initial comments?

valeriosetti:
Rebased to align to recent changes done in https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/867 (which is the base PR for this one)

include/psa/crypto_values.h
valeriosetti: That's a good point and that's something I came across while working on this (but forgot to ask before finalizing the PR). The problem is that there is no `PSA_KEY_TYPE_BLAKE2_MAC` in the 1.5 specs, so I'm currently using `PSA_KEY_TYPE_HMAC`. In terms of key type and content I don't think there is a difference between them, since the key is only a predefined amount of bytes that will be taken into account when computing the hash. However, for sake of uniformity, I think that using `PSA_KEY_TYPE_BLAKE2_MAC` with `PSA_ALG_BLAKE2_MAC` is clearer, than using `PSA_KEY_TYPE_HMAC`. 
@gilles-peskine-arm wdyt about this? Should I open a PR in the PSA API repo to fill this gap?

core/psa_crypto.c
valeriosetti: Currently I'm relying on HMAC key checks before using that for Blake2 (which as [mentioned above](https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/898#discussion_r4061673463) it might not be correct). If we add `PSA_KEY_TYPE_BLAKE2_MAC` then I will update the PR adding key management as well.

include/psa/crypto_values.h
gilles-peskine-arm: The key type is called [`PSA_KEY_TYPE_BLAKE2`](https://arm-software.github.io/psa-api/crypto/1.5/api/keys/types.html#c.PSA_KEY_TYPE_BLAKE2).

valeriosetti: Oh, I thought I searched for it and it didn't show up. Perhaps I used the wrong search string on the web page. Anyway, sorry for the noise: I will update the PR ASAP.

PR #911, opened 2026-09-15
tests: host_test: add the possibility to filter test cases on a suite

valeriosetti:
## Description

Add the "-f"/"--filter" command line option to each test suite. This is helpful during debug sessions when a user want to execute only a specific series of test cases and skip the others.

I know that I can get a similar result by creating a new file named `test_suite_something.tmp.data` and then debugging the `test_suite_something.tmp` executable, but I think my solution is faster because we can build the executable once and then dynamically run only the tests we're interested into.

## PR checklist

- [ ] **changelog** not required because:  just for testing
- [ ] **TF-PSA-Crypto development PR** provided: this one
- [ ] **TF-PSA-Crypto 1.1 PR** not required because: maybe, it depends on what reviewers think of this one first.
- [ ] **mbedtls development PR** not required because: it benefits from this one
- [ ] **mbedtls 4.1 PR** not required because: see above comments
- [ ] **mbedtls 3.6 PR** not required because: I would say no
- **tests**  provided


gilles-peskine-arm:
The CI is unhappy.

tests/suites/host_test.function
```diff
@@ -507,6 +513,25 @@ static void try_chdir_if_supported(const char *argv0)
 }
 #endif /* MBEDTLS_HAVE_CHDIR */
 
+/**
+ * \brief Check if the given test case should be filtered out or not.
+ *
```
bjwtaylor: Trailing whitespace is causing the ci to fail

tests/suites/host_test.function
```diff
@@ -597,6 +624,15 @@ static int execute_tests(int argc, const char **argv)
                    strcmp(next_arg, "-h") == 0) {
             mbedtls_fprintf(stdout, USAGE);
             mbedtls_exit(EXIT_SUCCESS);
+        } else if (strcmp(next_arg, "--filter") == 0 ||
+                   strcmp(next_arg, "-f") == 0) {
+            if (test_case_filters_count >= TEST_CASE_FILTER_MAX) {
+                mbedtls_fprintf(stderr, "Too many test filters specified.\n");
+                mbedtls_exit(MBEDTLS_EXIT_FAILURE);
+            }
+            arg_index++;
```
bjwtaylor: If I run ./test_suite_base64 -f DOES_NOT_MATCH_ANY_CASE --filter it causes a segfault. Is this expected, I think this is being caused because you need to check an operand exists before consuming it.

valeriosetti: Well, ideally yes, but this is really meant to be used only for debug purposes. I mean, I can try to fix it, but IMO this doesn't necessarily need to be unbreakable.

bjwtaylor (approved):
I think yes, as it's only debug code it's probably no worth investing time in the fix.

tests/suites/host_test.function
```diff
@@ -507,6 +513,25 @@ static void try_chdir_if_supported(const char *argv0)
 }
 #endif /* MBEDTLS_HAVE_CHDIR */
 
+/**
+ * \brief Check if the given test case should be filtered out or not.
+ *
+ * \param test_case     The name of the test case to be evaluated.
+ * \param filters       Array of filters to be checked against.
+ * \param filters_count The number of items in \p filters
+ * \return int          1 is the test should be filtered out; 0 otherwise.
```
mpg: Typo: s/is/if/

tests/suites/host_test.function
mpg: I'd prefer not to have spurious segfaults in my debug sessions. If I'm deep in the middle of something complicated, I'm more likely to not notice the segfault is due to me being silly as opposed to the thing I'm debugging.

Anyway, I can see you fixed it in the meantime. But I still wanted to comment on the fact that IMO debug code should either work or give a reasonably clean error. (Since it's only for us devs, it's OK if understanding the error requires looking at the source. But still the error needs to be clearly distinct from whatever we're debugging at the moment, which a segfault isn't.)

tests/suites/host_test.function
```diff
@@ -651,6 +690,20 @@ static int execute_tests(int argc, const char **argv)
...
                 break;
             }
+
+            /* Apply filters on test cases (if any). */
+            if (test_case_filters[0] != NULL) {
+                if (is_test_case_filtered_out(line.content, test_case_filters,
+                                              test_case_filters_count)) {
+                    /* Skip the next 2 lines (i.e. depends_on and the test function) */
+                    length = get_line(file, &line);
+                    if (strstr(line.content, "depends_on") != NULL) {
```
mpg: Minor: codex points out that matching anywhere in the line will result in unpleasant outcomes if we ever have a `.data` file that has something like:
```
Test name
test_func:"blabla depends_on"
```

That seems easy enough to fix by not using `strstr()` here but rather `strncmp()`:
```
if (strncmp(line.content, "depends_on:", 11) == 0) { 
```
(Note that while at it, we can also check for a colon.)

Seems mostly theoretical and unlikely to happen in practice, but since the fix is easy, we might as well fix it.

mpg (changes requested):
Mostly loooking good, just two minor things.

Also, I think it's worth back/side-porting to 1.1 and 4.1 at least.

tests/suites/host_test.function
```diff
@@ -651,6 +690,20 @@ static int execute_tests(int argc, const char **argv)
...
                 break;
             }
+
+            /* Apply filters on test cases (if any). */
+            if (test_case_filters[0] != NULL) {
+                if (is_test_case_filtered_out(line.content, test_case_filters,
+                                              test_case_filters_count)) {
+                    /* Skip the next 2 lines (i.e. depends_on and the test function) */
+                    length = get_line(file, &line);
+                    if (strncmp(line.content, "depends_on:", 11) == 0) {
```
bjwtaylor: I think you still have a whitespace issue here?

valeriosetti: Yes, sorry :(

bjwtaylor (approved):

mpg (approved):
LGTM @ b30ad48, thanks!

mpg:
@valeriosetti If you want to clean up the history a bit (like squash typo fixes & whitespace fixes) before backporting, feel free to do so. (I think this won't invalidate our approvals, just require a new CI run.) And if you'd rather just backport as it is, also feel free to do so :)

